PR #8, opened 2024-09-12
Fix/swap profitability

Divya-Solulab:
Implements a selection mechanism to the Optimus agent to bridge/swap assets only if transaction fees are lower than the transacted amount and the amount exceeds a set threshold.


Divya-Solulab:
Not to be merged before the launch

gauravlochab (approved):

dagacha (approved):

PR #25, opened 2024-10-22
chore: initial list of packages

DavidMinarsch:


mints/mapping.txt
```diff
@@ -0,0 +1,17 @@
+token_id-"component_id"
```
DavidMinarsch, line 1: @gauravlochab and @Divya-Solulab are there any packages missing from this list that are part of optimus and in this repo (i.e. not external dependencies)?

Divya-Solulab: No, all the packages are there.

Divya-Solulab (approved):

mints/mapping.txt
gauravlochab: No

mints/mapping.txt
```diff
@@ -0,0 +1,17 @@
...
+X-"contract/valory/staking_activity_checker/0.1.0"
+X-"contract/valory/staking_token/0.1.0"
+X-"contract/valory/uniswap_v3_non_fungible_position_manager/0.1.0"
+X-"contract/valory/uniswap_v3_pool/0.1.0"
+X-"skill/valory/liquidity_trader_abci/0.1.0" -> previous 8 ids tbd, 3, 5, 14, 19, 20, 22
+X-"skill/valory/optimus_abci/0.1.0" -> 14, 16, 17, 22, 33, previous skill
+---
+token_id-"agent_id"
+X-"agent/valory/optimus/0.1.0" -> 1-7, 8-11, 39, 50, 60, all contracts above, 18, 19, 20, 13, 14, 16, 17, 22, 33, two skills above
+---
```
DavidMinarsch, line 15: @Divya-Solulab please check the dependencies are correct with the help of David V

https://github.com/valory-xyz/optimus/blob/main/packages/valory/agents/optimus/aea-config.yaml weirdly is missing some contract references which subdependencies reference. Is this a bug in OA @Adamantios ?

Adamantios: If the agent runs without the framework complaining then yes. It should raise some error in the following form:

```
Error: Following dependencies are present in the project but missing from the aea-config.yaml; {PackageId(contract, author/missing_package:the.version)}
```

PR #28, opened 2024-10-31
Remove ledger overrides

jmoreira-valory:
Remove ledger overrides so that it can be integrated in Pearl.


jmoreira-valory:
Closed due to https://github.com/valory-xyz/optimus/pull/31

PR #111, opened 2025-05-07
Add Max Loss Percentage Prompt, CS Calculation, and Fallback Logic

princevijay27:
### **Summary:**

This PR introduces a new logic flow to support dynamic max loss percentage input from LLM responses. It integrates a Composite Score (CS) calculation based on Value at Risk (VaR), updates strategy application logic, and adds robust fallback mechanisms.

---

#### **1. Prompt & Parsing Enhancements**

* Updated prompt to include a request for `max_loss_percentage`.
* Enhanced `_parse_llm_response`:

  * Extracts `max_loss_percentage` from the LLM output.
  * Applies a default based on trading type if missing.
  * Validates that the percentage is within \[1%, 30%].

#### **2. Composite Score (CS) Calculation**

* Introduced `calculate_cs_from_var(var_value)` method:

  * Converts the percentage into decimal VaR.
  * Applies Michael’s formula:

    ```
    CS = ln((8.786×10⁻¹)/(VaR + 8.272×10⁻¹) - 2.447)/(-7.552)
    ```
  * Applies bounds checking to restrict CS to the range \[0.2, 0.5].
  * Handles edge cases and mathematical errors gracefully.

#### **3. Strategy Update Logic**

* Modified `_handle_llm_response`:

  * Converts `max_loss_percentage` to VaR.
  * Computes CS using the new method.
  * Stores both `max_loss_percentage` and `calculated_cs` in system state.
  * Updates trading type and response payload.
  * Persists values to KV store for reuse.

#### **4. Fallback Mechanisms**

* If `max_loss_percentage` is missing:

  * Default to 5% (balanced) or 10% (risky).
* If CS calculation fails:

  * Log the error and fallback to previous CS value.
* If CS is out of bounds:

  * Clamp to allowed min/max values.
* If system state is lost:

  * Revert to default thresholds from constants.


ignore_gitleaks.py
DIvyaNautiyal07, line 1: Please remove this file

packages/valory/skills/liquidity_trader_abci/behaviours.py
DIvyaNautiyal07, line 1: are you sure line 2480 works fine now that execute_strategy returns generator, please test and confirm. I guess we need to use "yield from" while calling this function

packages/valory/skills/optimus_abci/handlers.py
```diff
@@ -307,6 +308,7 @@ def _handle_get_features(
         """
         # Check if GENAI_API_KEY is set
         api_key = self.context.params.genai_api_key
+        self.context.logger.info(f"genai_api_key API KEY {api_key}")
```
DIvyaNautiyal07, line 311: please remove this

packages/valory/skills/optimus_abci/handlers.py
DIvyaNautiyal07, line 1: use constants for numbers and define them at the top of the file for better understanding

DIvyaNautiyal07 (changes requested):

PR #140, opened 2025-07-21
Feat/withdrawal feature abci

DhairyaPatel7:


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -688,6 +694,503 @@ def read_portfolio_data(self) -> None:
         """Read the portfolio data from JSON."""
         self._read_data("portfolio_data", self.portfolio_data_filepath)
 
+    def update_portfolio_after_action(self) -> Generator[None, None, None]:
```
DIvyaNautiyal07, line 697: consider breaking down this function into smaller functions

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -688,6 +694,503 @@ def read_portfolio_data(self) -> None:
...
+                    if isinstance(position_value, str):
+                        position_value = (
+                            float(position_value) if position_value else 0.0
+                        )
+                    value_in_pools += position_value
+
+            # Get current balances from blockchain (FRESH DATA)
+            chain = self.context.params.target_investment_chains[0]
+            if chain == "optimism":
+                balances = yield from self._get_optimism_balances_from_safe_api()
```
DIvyaNautiyal07, line 728: Just to be sure, do you think this might create any inconsistencies, since the rest of the code uses a different method for calculating balances? We might also need to update the balances in self.synchronized_data.positions, right? Maybe worth double-checking to avoid any potential issues but if it looks fine, then no worries.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -0,0 +1,732 @@
...
+                    performative=ContractApiMessage.Performative.GET_STATE,
+                    contract_address=pool_address,
+                    contract_public_id="valory/velodrome_pool:0.1.0",
+                    contract_callable="get_balance",
+                    data_key="balance",
+                    account=safe_address,
+                    chain_id=chain,
+                )
+
+            elif dex_type == "uniswap_v3":
```
DIvyaNautiyal07, line 309: There is no get_balance in UniswapV3 Pool since it is a concentrated pool, you would need to fetch liquidity from UniswapV3NonFungiblePositionManagerContract

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -0,0 +1,732 @@
...
+
+            updated_positions.append(position)
+
+        # Update the current positions in the base behaviour
+        self.current_positions = updated_positions
+        self.store_current_positions()
+
+        return updated_positions
+
+    def _get_pool_balance(
```
DIvyaNautiyal07, line 272: This doesn't handle the case if there's a Velodrome CL Pool Position

DIvyaNautiyal07 (changes requested):

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DhairyaPatel7: For now, updating the portfolio data so that the user can see the updated portfolio value on the page fetched from the on chain balance. The positions data will be synced in the fetch_startegies at the start of next round along with the portfolio update. Did not want to duplicate the entire behaviour and call it in post_execute_tx everytime.

DIvyaNautiyal07 (approved):

PR #170, opened 2025-09-08
Feat/modius usdc airdrop integration

DIvyaNautiyal07:


packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2790,6 +2859,23 @@ def _calculate_chain_investment_value(
                     if amount <= 0:
                         continue
 
+                    # Check if this is an airdropped USDC transfer and exclude it from initial investment
+                    if (
+                        chain == "mode"
+                        and token_symbol.upper() == "USDC"
```
DhairyaPatel7, line 2865: if this is a chain specific comparision, shouldn't we use the USDC token address here for comparision?

PR #207, opened 2025-10-30
Feat/support non x402 mode coingecko

DIvyaNautiyal07:
Supports users to use the agent in non-x402 mode as well


packages/valory/services/optimus/service.yaml
OjusWiZard: token_price_endpoint: ${COINGECKO_TOKEN_PRICE_ENDPOINT:str:/api/v3/simple/token_price/{asset_platform_id}?contract_addresses={token_address}&vs_currencies=usd}
      coin_price_endpoint: ${COINGECKO_COIN_PRICE_ENDPOINT:str:/api/v3/simple/price?ids={coin_id}&vs_currencies=usd}

PR #267, opened 2026-04-10
fix: decouple Safe Transaction Service rate limiting from CoinGecko

DIvyaNautiyal07:
## Summary

- Decouple Safe Transaction Service / Optimism RPC rate limiting from CoinGecko rate limiter — a 429 from SafeGlobal was zeroing out the CoinGecko rate limiter, stalling the FSM for 3+ minutes and causing $0 portfolio in Pearl
- Use `retry_wait` param in the 429 sleep path instead of hardcoded 60s, reducing wait from 60s to 5s for non-CoinGecko endpoints
- Add KV store cache fallback for safe balances — on SafeGlobal API failure, use last successful balances instead of returning empty
- Return `(api_success, balances)` tuple from `_fetch_safe_balances_with_pagination` to distinguish API failure from legitimately empty Safe — cache fallback only triggers on actual failure
- Cache empty balances on API success so stale data is cleared when user withdraws all funds

## Context

Multiple users reported $0 portfolio and lost staking streaks. Root cause: all `_request_with_retries` calls in `fetch_strategies.py` used `self.coingecko.rate_limited_status_callback` regardless of which API they were calling. When SafeGlobal returned 429, it zeroed the CoinGecko rate limiter, blocking price fetches and stalling the entire FSM.

Seen in production: Zendesk #903, plus 3 other user reports across Optimism and Mode chains.

## Changes

**`base.py`** (`_request_with_retries`)
- Use `retry_wait` param in 429 sleep path: `wait_time = retry_wait if retry_wait > 0 else 60`

**`base.py`** (`_fetch_safe_balances_with_pagination`)
- Return `(api_success, balances)` tuple instead of bare list

**`base.py`** (`_get_optimism_balances_from_safe_api`)
- On `api_success=True`: cache balances to KV (including empty — clears stale cache)
- On `api_success=False`: read last cached balances from KV
- Both paths wrapped in try/except

**`base.py`** (module level)
- Add `_noop_rate_limit_callback()` function after constants block

**`fetch_strategies.py`** (8 call sites)
- `rate_limited_callback`: `self.coingecko.rate_limited_status_callback` → `_noop_rate_limit_callback`
- `retry_wait`: `self.params.sleep_time` (60s) → `5`
- `max_retries`: default (3) → `2`

## Test plan

- [x] 2972 tests pass, 100% branch coverage
- [x] All 7 linters pass (black, isort, flake8, mypy, pylint, darglint, bandit)
- [x] Tests for: API success with balances, API success empty (no cache fallback), API failure with cache fallback, API failure no cache, cache write failure, cache read failure, pagination tuple return
- [ ] Deploy to staging agent and verify portfolio reports non-zero during SafeGlobal rate limiting


packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3361,8 +3362,9 @@ def _fetch_token_transfers(
             endpoint=endpoint,
             headers={"Accept": "application/json"},
             rate_limited_code=429,
-            rate_limited_callback=self.coingecko.rate_limited_status_callback,
-            retry_wait=self.params.sleep_time,
+            rate_limited_callback=_NOOP_RATE_LIMIT_CALLBACK,
+            retry_wait=5,
```
LOCKhart07, line 3366: **Blocking — `retry_wait=5` is dead code on the 429 path.**

`_request_with_retries` in `base.py:1488-1506` hardcodes `yield from self.sleep(60)` for `rate_limited_code` (429). The `retry_wait` parameter is only consulted in the generic non-OK failure branch (around line 1530). With `max_retries=2`, a SafeGlobal 429 still blocks the FSM for ~60s before failing — not 5s as the PR description and commit message claim.

The rate-limited-callback decoupling is the real fix here and is correct, but please either:
- plumb `retry_wait` into the 429 sleep in `_request_with_retries` (`yield from self.sleep(retry_wait)`), so the new `retry_wait=5` actually takes effect; **or**
- drop the `retry_wait=5` changes and update the PR description so operators don't expect a 5s recovery that won't happen.

Applies to all 8 call sites in this file.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -464,6 +465,29 @@ def _get_optimism_balances_from_safe_api(
             safe_address
         )
 
+        if not all_balances:
```
LOCKhart07: **Blocking — empty-Safe state gets silently overwritten with stale cache.**

`_fetch_safe_balances_with_pagination` returns `[]` in two distinct cases: API failure (`break` at line 575 after `if not success`) and Safe legitimately empty (`break` at line 580 after `if not results`). `if not all_balances` cannot tell them apart, so after the first successful fetch, if the user later withdraws all funds, the next call will load stale cached balances and the agent will believe it owns assets it doesn't — potentially attempting trades against phantom funds.

Fix: have `_fetch_safe_balances_with_pagination` return a `(success, balances)` tuple (or raise on failure) and only fall back to cache on `success is False`.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -86,6 +86,7 @@
 REQUIRED_REQUESTS_SAFETY_MARGIN = 1
 MAX_RETRIES_FOR_API_CALL = 3
 MAX_RETRIES_FOR_ROUTES = 3
+_NOOP_RATE_LIMIT_CALLBACK: Callable = lambda: None  # For non-CoinGecko APIs
```
LOCKhart07: **Non-blocking — lambda assigned to a name (E731).**

Most lint configs flag `name = lambda: ...`. Standard fix:

```python
def _noop_rate_limit_callback() -> None:
    """No-op rate limit callback for non-CoinGecko APIs."""
```

Reads better at the call sites too.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -726,7 +750,7 @@ def _fetch_mode_tokens_with_pagination(
                     "Pragma": "no-cache",
                     "Expires": "0",
                 },
-                rate_limited_callback=lambda: None,
+                rate_limited_callback=_NOOP_RATE_LIMIT_CALLBACK,
```
LOCKhart07: **Non-blocking — Mode Explorer has no equivalent cache fallback.**

The KV cache fallback only exists for the Optimism Safe path. If the Mode Explorer API flakes the same way, Mode users will still see $0 portfolio. Out of scope for this PR per the description, but worth a follow-up issue so the same incident doesn't recur on the Mode side.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_base.py
```diff
@@ -968,6 +968,37 @@ def test_with_reward_and_ousdt(self) -> None:
...
+        b._read_kv = _make_gen({"safe_balances_optimism": json.dumps(cached_data)})
+        b._write_kv = _make_gen(True)
+        b._fetch_reward_balances = _make_gen([])
+        b._fetch_ousdt_balance = _make_gen(None)
+        result = _exhaust(b._get_optimism_balances_from_safe_api())
+        assert len(result) == 2
+        assert result[0]["asset_symbol"] == "ETH"
+        assert result[1]["asset_symbol"] == "USDC"
+
+    def test_fallback_cache_empty(self) -> None:
```
LOCKhart07: **Non-blocking — test gap on the success/empty distinction.**

Neither new test exercises the case where the SafeGlobal API succeeds but legitimately returns zero balances (the regression flagged on `base.py:468`). Once `_fetch_safe_balances_with_pagination` is updated to signal success vs failure, please add a test that covers:
- API success with empty results → no cache fallback, return empty
- API failure with non-empty cache → cache fallback used

LOCKhart07 (changes requested):

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed. `_request_with_retries` now uses the `retry_wait` param in the 429 sleep path instead of the hardcoded 60s: `wait_time = retry_wait if retry_wait > 0 else 60`. The 60s default is preserved for callers that don't pass `retry_wait` (e.g. CoinGecko calls).

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed. `_fetch_safe_balances_with_pagination` now returns `(api_success, balances)` tuple. The caller only falls back to cache when `api_success is False`. When the API succeeds but returns empty (legitimately empty Safe), no cache fallback happens. Added tests for both paths.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed. Replaced with `def _noop_rate_limit_callback() -> None`.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: We don't support Mode chain as of now, so out of scope for this PR.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_base.py
DIvyaNautiyal07: Added. New tests: `test_api_success_empty_no_cache_fallback`, `test_api_failure_falls_back_to_cache`, `test_cache_write_failure`, `test_cache_read_failure`. Also updated existing `_fetch_safe_balances_with_pagination` tests to handle the new `(success, balances)` tuple return.

LOCKhart07 (approved):
LGTM — blocking issues from the prior review are addressed:

- `retry_wait` plumbed into the 429 sleep path so `retry_wait=5` actually takes effect
- `_fetch_safe_balances_with_pagination` returns `(success, balances)`, distinguishing API failure from a legitimately empty Safe; cache fallback only fires on real failure
- Empty results are written back to KV on success, clearing stale cache when funds are withdrawn
- `def _noop_rate_limit_callback` replaces the lambda
- New tests cover success+empty (no fallback), failure+cache, write/read failures, and are falsifiable

**Non-blocking nit (please fix in follow-up or this PR):**

`tests/behaviours/test_base.py:4028` `test_safe_balances_two_pages` wasn't updated for the tuple return. `result` is now `(True, [...])`, so `assert len(result) == 2` asserts the tuple length (always 2), not the balance count — it would pass for `(True, [])` or any other return. Pagination is no longer being validated.

```python
success, balances = _exhaust(b._fetch_safe_balances_with_pagination("0xSafe"))
assert success is True
assert len(balances) == 2
```

DhairyaPatel7 (approved):

PR #268, opened 2026-04-17
Fix/velodrome cl stale staked state

DIvyaNautiyal07:
# Verify on-chain state before CL stake/unstake/exit/claim

Fixes ZD #950 / OPE-1562. A user's Velodrome CL NFT was tracked locally as staked but on-chain was not in the gauge. The agent built `gauge.withdraw(tokenId)` anyway, Safe wrapped the `"NA"` revert as `GS013`, and ~$32.55 got stuck.

Verified on-chain via cast for token `7265118`:

```
ownerOf                     = 0x9fd33068b81C8719A1b2972277a0CC98956B9247 (the Safe)
stakedContains(safe, token) = false
earned(safe, token)         = revert: NA
```

## Changes

Five on-chain checks added before emitting a tx that would otherwise revert. All new helpers return `None` on RPC failure and callers fall back to prior behavior, so transient outages don't silently drop valid actions.

1. **Unstake** (both withdraw and exit-strategy paths). `gauge.stakedContains(safe, tokenId)` filters the token list before building `UnstakeLpTokens`; skip the action entirely if nothing is actually staked.
2. **Claim CL rewards**. Same `stakedContains` filter before `gauge.getReward`, which otherwise reverts with `"NA"` on un-staked tokens.
3. **Stake**. Skip tokens already in the gauge's stake set before calling `gauge.deposit`; re-staking reverts and aborts the batch.
4. **Exit liquidity**. Always read `positions(tokenId).liquidity` instead of trusting the cached value. Stale cached liquidity caused `decreaseLiquidity` to revert with underflow after fees or partial exits.
5. **Exit ownership** (Velodrome CL + Uniswap V3). Skip tokens where `ownerOf(tokenId) != safe`. Avoids `"Not approved"` reverts when the NFT was transferred out externally.

## Latent bug caught in review

`get_liquidity_for_token_velodrome` and the Uniswap equivalent indexed the `get_position` response as a tuple (`position[2]`, `position[7]`) but the contract wrapper returns a dict keyed by field name. Dead code until #4 made those helpers run on every exit. Now uses `position.get("liquidity")`. Third commit in the branch.

## Tests

- New contract test: `VelodromeCLGaugeContract.staked_contains` with ABI integrity.
- New pool tests: `is_cl_token_staked`, `get_cl_position_owner`, `get_uniswap_position_owner`.
- Verified unstake builder tests cover: all-staked, none-staked, partial, RPC-failure fallback, non-CL passthrough, gauge resolvable, gauge unresolvable.
- Exit path tests: ownerOf mismatch skip, RPC failure continues, stale liquidity overridden.
- Stake path test: already-staked token skipped.

2730 skill tests pass. `tomte check-code` clean (black, isort, flake8, mypy, pylint, darglint). `autonomy packages lock --check` clean.

## Verification against live state

Ran a local script (not committed) that read the real ZD #950 position state off Optimism RPC, fed those exact values through the real production code paths via stubbed `contract_interact`, and asserted final action payloads across eight scenarios including the original failure. All passed.

## Scope

Velodrome CL and Uniswap V3 only. Non-CL Velodrome, Balancer, Merkl, Sturdy are untouched.

## Links

- Zendesk: https://valory.zendesk.com/agent/tickets/950
- Linear: https://linear.app/valory-xyz/issue/OPE-1562


dagacha:
**Code Review: Alternative Approaches Assessment**

After analyzing the architecture and the stale-state problem, I want to document why this PR's approach is the right architectural choice among the alternatives considered.

**The Core Problem Recap**
Local `current_positions` state drifts from on-chain reality due to: external NFT transfers, fee accrual changing liquidity values, partial exits via other interfaces, or failed post-tx persistence. The ZD #950 incident ($32.55 stuck) happened because `position["staked"] = True` locally but `gauge.stakedContains(safe, tokenId) = False` on-chain.

**Alternatives Considered**

| Approach | Verdict | Reasoning |
|----------|---------|-----------|
| **Event-driven reconciliation** | ❌ Overkill | Requires indexing infrastructure; misses historical events; agent acts periodically not event-responsively |
| **Transaction simulation** | ❌ Too heavy | CL multi-step interactions hard to simulate; slippage/time-sensitive ops fail valid simulations |
| **Eliminate `staked` cache entirely** | ❌ Too expensive | Would require 1+ RPC calls per position every evaluation cycle; flag is correct 99%+ of time |
| **Periodic full reconciliation** | ⚠️ Partial | Race conditions remain (state changes between reconcile→action); adds RPC load; doesn't solve immediate "act on stale" risk |
| **Optimistic execution with retry** | ❌ Current broken behavior | This is exactly what caused ZD #950 — GS013 revert costs gas and blocks subsequent actions |

**Why Point-in-Time Verification Wins**

PR #268 hits the sweet spot:

| Criterion | PR #268 |
|-----------|---------|
| Defensive at action time | ✅ Verifies exactly when revert risk exists |
| RPC efficiency | ✅ Targeted calls only (stake/unstake/exit/claim paths) |
| Fail-open on RPC errors | ✅ Falls back to prior behavior; no new failure modes |
| Production complexity | ✅ Minimal, auditable additions |
| Addresses root cause | ✅ Prevents revert-risky transactions from being built |

**Complementary Enhancement (Optional Future)**

Consider adding lightweight ownership verification in `GetPositionsRound` to proactively *detect* drift for visibility:

```python
# Detect external transfers; mark for manual handling
if owner != safe: position["external_transfer_detected"] = True
```

This would complement PR #268's prevention-focused approach with detection.

**Conclusion**

The point-in-time verification pattern is the correct architectural choice. It adds minimal complexity while eliminating the entire class of "stuck funds from stale state" failures. Well-structured fix with comprehensive test coverage.

packages/valory/skills/liquidity_trader_abci/pools/uniswap.py
```diff
@@ -188,11 +188,23 @@ def exit(
...
+            )
+            return None, None, None
+
+        cached_liquidity = liquidity
+        liquidity = yield from self.get_liquidity_for_token(token_id, chain)
         if not liquidity:
-            liquidity = yield from self.get_liquidity_for_token(token_id, chain)
-            if not liquidity:
-                return None, None, None
+            return None, None, None
```
DhairyaPatel7, line 217: Behavior change vs. the PR description's promise that "transient outages don't silently drop valid actions".

Old code used `cached_liquidity` if truthy and only fetched on cache miss. New code always overwrites with the on-chain read, so an RPC failure here aborts the entire exit even when a valid cached value was passed in.

Either fall back to `cached_liquidity` on RPC failure:
```python
liquidity = yield from self.get_liquidity_for_token(token_id, chain)
if not liquidity:
    liquidity = cached_liquidity
if not liquidity:
    return None, None, None
```
or, if fail-closed is intentional (revert risk > availability), update the PR description to acknowledge that liquidity reads do not fall back to prior behavior on RPC outages.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -970,17 +970,29 @@ def _exit_cl_pool(
...
+            position_liquidity = yield from self.get_liquidity_for_token_velodrome(
+                position_token_id, chain
+            )
             if not position_liquidity:
-                position_liquidity = yield from self.get_liquidity_for_token_velodrome(
-                    position_token_id, chain
+                self.context.logger.warning(
+                    f"Could not get liquidity for token ID {position_token_id}, skipping"
+                )
+                continue
```
DhairyaPatel7, line 1013: Same regression as the uniswap exit path: on `get_liquidity_for_token_velodrome` returning None, this `continue` silently skips the position even when `cached_liquidity` was non-zero and would still be a usable value.

Either fall back to `cached_liquidity` here, or update the PR description — the "new helpers fall back to prior behavior" guarantee doesn't hold for liquidity reads.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -2686,6 +2686,90 @@ def _build_unstake_lp_tokens_action(
...
+
+        filtered_position = dict(position)
+        filtered_position["positions"] = [
+            p for p in positions_data if p.get("token_id") in staked_token_ids
+        ]
+        if position.get("token_id") is not None and not filtered_position["positions"]:
+            filtered_position["token_id"] = (
+                position.get("token_id")
+                if position.get("token_id") in staked_token_ids
+                else None
```
DhairyaPatel7: Nit: `else None` is unreachable.

We only reach this block with non-empty `staked_token_ids` (the `if not staked_token_ids: return None` above guarantees that). `filtered_position['positions']` is empty here only if `positions_data` was originally empty — which is the only branch that ever put `position.get('token_id')` into `token_ids`. So `position.get('token_id')` is necessarily in `staked_token_ids` whenever we hit this line.

Collapses to:
```python
filtered_position["token_id"] = position.get("token_id")
```

packages/valory/skills/liquidity_trader_abci/pools/uniswap.py
DIvyaNautiyal07: Fixed in b510f266. The exit path now falls back to the caller-supplied cached liquidity when the on-chain read returns None, and only short-circuits when both are unavailable. The divergence log fires only when both succeeded and disagreed. Added `test_falls_back_to_cached_liquidity_on_rpc_failure` and `test_returns_none_when_both_cached_and_onchain_unavailable` covering it.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Fixed in b510f266 with the same shape on the velodrome path: if `get_liquidity_for_token_velodrome` returns None we reuse `cached_liquidity` instead of skipping the position. Test `test_falls_back_to_cached_liquidity_on_rpc_failure` covers it.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in d3bab04f. Ternary collapsed to the unconditional assignment as suggested.

OjusWiZard (commented):
test

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -970,17 +970,29 @@ def _exit_cl_pool(
...
+            if cached_liquidity and cached_liquidity != position_liquidity:
```
OjusWiZard: **HIGH — Empty multisend when every CL exit token is skipped.**

The new `continue` paths in this loop (owner mismatch above, the falsy on-chain liquidity skip, and the later decrease/collect failures) can drain `multi_send_txs` to `[]`. After the loop ends the code still falls through to `get_tx_data(multi_send_txs=[])` at the multisend prep block (around L1066), producing an invalid/no-op multisend — the opposite of this PR's stated goal of not emitting txs that revert.

Suggest guarding before the multisend prep:

```python
if not multi_send_txs:
    self.context.logger.warning(
        f"No exitable CL positions for pool {pool_address}; skipping multisend"
    )
    return None, None, False
```

The single-token Uniswap `exit` isn't affected (its skips return `(None, None, None)` directly).

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -2686,6 +2686,90 @@ def _build_unstake_lp_tokens_action(
...
+            return None
+
+        filtered_position = dict(position)
+        filtered_position["positions"] = [
+            p for p in positions_data if p.get("token_id") in staked_token_ids
+        ]
+        if position.get("token_id") is not None and not filtered_position["positions"]:
+            filtered_position["token_id"] = (
+                position.get("token_id")
+                if position.get("token_id") in staked_token_ids
```
OjusWiZard: **MEDIUM — Stale top-level `token_id` when filtered `positions` is non-empty.**

The top-level `token_id` is only re-normalized against `staked_token_ids` when `filtered_position["positions"]` is empty. If a position carries both `positions=[{token_id: 1}, {token_id: 2}]` and a top-level `token_id: 2`, and only token 1 is on-chain staked, the filtered list keeps `[{token_id: 1}]` but the top-level `token_id` stays `2`.

Whether this matters depends on how `_build_unstake_lp_tokens_action` consumes the two fields, but it'd be safer to always normalize top-level `token_id` against `staked_token_ids` instead of gating the rewrite on `not filtered_position["positions"]`.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -3462,6 +3513,39 @@ def get_cl_staked_balance(
...
         return staked_amount
 
+    def is_cl_token_staked(
+        self, account: str, token_id: int, **kwargs: Any
+    ) -> Generator[None, None, Optional[bool]]:
+        """Return True/False if the gauge's stake set can be read, else None."""
+        chain = kwargs.get("chain")
+        gauge_address = kwargs.get("gauge_address")
+
+        if not all([chain, gauge_address, token_id is not None]):
```
OjusWiZard: **LOW — Readability nit.**

Mixing truthiness with `is not None` inside `all([...])` is functionally correct (covers `token_id == 0`) but easy to misread later. Two explicit checks read more clearly:

```python
if not chain or not gauge_address or token_id is None:
```

OjusWiZard (commented):
Three findings — see inline comments. The HIGH (empty multisend) is worth addressing before merge; MEDIUM and LOW can be follow-ups.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Already addressed in d3bab04f. The empty-multisend guard sits at velodrome.py:1071 right before `get_tx_data`, so when every token's `continue` path drains `multi_send_txs` to `[]`, we return None before invoking the multisend builder. Test `test_returns_none_when_all_tokens_skipped` covers it.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in a1210e8e. The filtered position now drops the top-level `token_id` whenever it isn't in `staked_token_ids`, regardless of whether the filtered `positions` list is empty. Test `test_partial_staked_drops_stale_top_level_token_id` covers your example (`positions=[{1},{2}]`, `token_id: 2`, only token 1 staked).

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Fixed in a1210e8e. Replaced with explicit per-field checks: `if not chain or not gauge_address or token_id is None`.

OjusWiZard (dismissed):

Adamantios (dismissed):

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -2686,6 +2686,90 @@ def _build_unstake_lp_tokens_action(
             )
             return None
 
+    def _build_unstake_lp_tokens_action_verified(
+        self, position: Dict[str, Any]
+    ) -> Generator[None, None, Optional[Dict[str, Any]]]:
+        """Build an UnstakeLpTokens action after on-chain stake verification for CL pools."""
+        is_cl_pool = position.get("is_cl_pool", False)
```
jmoreira-valory, line 2757: **Fail-open is intentional ([dagacha's review](https://github.com/valory-xyz/optimus/pull/268#issuecomment-4279499661) explicitly endorses it), but the trade-off deserves stronger guard-rails — observability and possibly retry — given the original incident's exact trigger.**

The design choice here is: on RPC verification failure, fall back to `_build_unstake_lp_tokens_action(position)` — i.e. the *unverified* path that produced the original GS013 stuck funds. Same applies for 5 other early-return paths (missing chain/pool/safe, no token IDs, no pool registry, no gauge resolvable, `verification_failed`). Per the PR description and dagacha's analysis this is intentional ("transient outages don't silently drop valid actions").

The concern I want to raise is that the original ZD #950 trigger (local cache says staked, on-chain says not) could have itself been caused by transient RPC/indexer drift at write time. If so, the fix protects against the symptom only when the next read-time RPC succeeds — under sustained RPC flakiness on Optimism, the same buggy call could be re-emitted.

Suggested mitigations (pick one or more, none are blockers):
- Add a `MAX_VERIFY_RETRIES` retry around `is_cl_token_staked` before the fallback fires (cheap, single read), so a one-shot RPC blip doesn't immediately drop into unverified mode.
- Emit a structured `WARNING` with stable fields (`pool_address`, `gauge_address`, `chain`, `token_ids`) on every fallback so this failure mode is dashboard-detectable. Today the warning text varies across branches and is hard to alert on.
- Add a docstring note on `_build_unstake_lp_tokens_action_verified` explicitly documenting "verified" as best-effort: "Returns to the unverified path on RPC failure; consumers should not assume on-chain state was checked." That stops a future reader from inferring guarantee from the name.

Not blocking — the architectural choice is sound — but the original incident remains reproducible if the RPC pattern repeats.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -3125,7 +3195,7 @@ def stake_cl_lp_tokens(
 
             staked_positions.append({"token_id": token_id})
```
jmoreira-valory, line 3223: **`if not staked_positions: return {"error": ...}` reports a successful no-op as failure.**

When every token in `token_ids` is already staked (which is exactly what `is_cl_token_staked is True` tells us), this path returns:

```python
if not staked_positions:
    error_msg = "No valid stake transactions created"
    self.context.logger.error(error_msg)
    return {"error": error_msg}
```

I see the rationale in `test_all_already_staked_returns_error_not_solo_approval`'s docstring: *"do not submit a solo setApprovalForAll multisend. Return an error so the caller doesn't advance the action index treating it as success."* The orphan-approval concern is real, but the response shape conflates two distinct outcomes:

- "everything is already in the desired state" → success / no-op
- "the stake legitimately failed" → error

Caller code can't tell them apart from `{"error": ...}` alone, and downstream metrics / alerts will (or already do) treat these identically.

Cleaner shape:

```python
if not staked_positions:
    self.context.logger.info(
        f"All {len(token_ids)} tokens already staked in gauge {gauge_address}; nothing to do"
    )
    return None  # or {"skipped": "all_already_staked"}
```

…and gate the upstream `setApprovalForAll` emission on "at least one deposit will follow" so the orphan-approval case can't arise in the first place. That fixes the root cause (orphan approval) without overloading the error channel.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
```diff
@@ -4186,6 +4186,31 @@ def get_claim_staking_rewards_tx_hash(
...
+                        break
+                    if is_staked:
+                        staked_token_ids.append(token_id)
+                if not verification_failed:
+                    if not staked_token_ids:
+                        self.context.logger.info(
+                            f"Skipping claim for pool {pool_address}: none of "
+                            f"{token_ids} are staked in gauge {gauge_address}"
+                        )
+                        return None, None, None
```
jmoreira-valory, line 4211: **Code duplication: same staked-token-id filter loop in two places.**

The block here in `_claim_cl_rewards`:

```python
staked_token_ids: List[int] = []
verification_failed = False
for token_id in token_ids:
    is_staked = yield from pool.is_cl_token_staked(
        self, safe_address, token_id, chain=chain, gauge_address=gauge_address,
    )
    if is_staked is None:
        verification_failed = True
        break
    if is_staked:
        staked_token_ids.append(token_id)
```

…is byte-for-byte identical to the loop in `base.py:_build_unstake_lp_tokens_action_verified`. Worth extracting:

```python
def _filter_staked_token_ids(
    self, pool, safe_address: str, token_ids: List[int], chain: str, gauge_address: str,
) -> Generator[None, None, Tuple[List[int], bool]]:
    """Return (staked_ids, verification_failed)."""
```

Two benefits beyond the obvious:

1. The fallback policy on `verification_failed` differs across the two call sites — the unstake-verified helper falls back to `_build_unstake_lp_tokens_action(position)` (unverified path), while `_claim_cl_rewards` keeps the original `token_ids` (also effectively unverified). With a helper, that policy choice becomes a single explicit decision at each call site rather than buried in a copy-pasted loop.
2. If a future caller (e.g. the stake-skip path in `velodrome.py:3157`) wants the same filter, they get it for free without re-introducing drift.

Not blocking, but worth doing while the surface is fresh.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: All three suggestions applied in 03de3053.

1. `MAX_VERIFY_RETRIES = 2` retry around `is_cl_token_staked` via `_verify_cl_token_staked_with_retry`, so a one-shot RPC blip no longer drops into the unverified path.

2. Every fallback now emits a single structured warning `unstake_verification_fallback reason=... chain=... pool=... gauge=... token_ids=...` via `_log_unstake_fallback`. Six distinct `reason` values cover all the early-return paths (`missing_required_fields`, `no_token_ids`, `no_velodrome_pool_registry`, `gauge_address_unresolvable`, `verification_rpc_failed`). Stable, alertable.

3. Docstring + an inline comment on `_build_unstake_lp_tokens_action_verified` now explicitly describe verification as best-effort and the name as non-binding.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Fixed in 03de3053.

Reordered `stake_cl_lp_tokens` so the already-staked filter runs first (`tokens_to_stake = [t for t in token_ids if not is_cl_token_staked(t)]`). If that list is empty we log INFO ("All N tokens already staked in gauge X; nothing to do") and `return None`. Approval and deposits are only built when there's actual work to follow them, so the orphan approval-only multisend can no longer arise.

Caller side in `decision_making.py` distinguishes the two outcomes now: `result is None` is logged as info ("No staking work for pool X; advancing action") and returns `(None, None, None)` so the FSM emits `DONE` and advances. `result.get("error")` continues to log error.

Test updated: `test_all_already_staked_returns_none_no_op` asserts the no-op return and that neither `is_approved_for_all` nor `set_approval_for_all` nor `deposit` is called.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
DIvyaNautiyal07: Extracted in d47acda3. New helper `_filter_staked_token_ids(pool, safe_address, token_ids, chain, gauge_address) -> (staked_ids, verification_failed)` lives on `LiquidityTraderBaseBehaviour`. Both `_build_unstake_lp_tokens_action_verified` and the claim-CL-rewards path now call it. Each call site keeps its own fallback policy on `verification_failed`: the unstake builder calls `_build_unstake_lp_tokens_action(position)`, the claim path keeps the original `token_ids`. The retry logic from #3187865765 is folded into the helper so both call sites get it.

DhairyaPatel7 (dismissed):

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -1043,6 +1068,12 @@ def _exit_cl_pool(
                 }
             )
 
+        if not multi_send_txs:
+            self.context.logger.warning(
+                "No exit transactions queued for any token; skipping multisend"
+            )
+            return None
```
bennyjo: **Tuple-unpack TypeError on this new early return.**

`_exit_cl_pool` returns plain `None` here, but the outer `velodrome.exit` is `return (yield from self._exit_cl_pool(...))` and the caller in `decision_making.py:1815` does:

```python
tx_hash, contract_address, is_multisend = yield from pool.exit(self, **exit_pool_kwargs)
```

Unpacking `None` as a 3-tuple raises `TypeError: cannot unpack non-iterable NoneType object`. The new path is reachable in the exact drift modes this PR is meant to handle:

- every NFT failed `ownerOf` (transferred out — ZD #950 sibling case),
- every token's on-chain liquidity is 0 / RPC returned None and no cached value,
- every decrease/collect failed.

`test_returns_none_when_all_tokens_skipped` and `test_skips_when_onchain_liquidity_is_zero` assert `result is None` directly and don't go through the caller, so they pass against this bug. Suggest:

```python
if not multi_send_txs:
    self.context.logger.warning(
        "No exit transactions queued for any token; skipping multisend"
    )
    return None, None, None
```

…and a regression test that drives `pool.exit` (not `_exit_cl_pool`) through the unpack.

Out-of-scope but worth flagging as a follow-up: the pre-existing `return None` lines in `_exit_cl_pool` (`position_manager_address`, `multisend_address`, `multisend_tx_hash` branches) and the `return None, None` at the slippage-failure branch have the same shape mismatch — latent on `main` but not introduced here.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
```diff
@@ -4186,6 +4187,21 @@ def get_claim_staking_rewards_tx_hash(
                     )
                     return None, None, None
 
+                # On RPC failure keep the original list.
+                staked_token_ids, verification_failed = (
+                    yield from self._filter_staked_token_ids(
+                        pool, safe_address, token_ids, chain, gauge_address
+                    )
+                )
+                if not verification_failed:
```
bennyjo: **Asymmetric observability vs. the unstake fallback path.**

When `verification_failed` is True here, the original `token_ids` are kept and the function falls through silently — no structured warning. The unstake path goes through `_log_unstake_fallback` and emits `unstake_verification_fallback reason=verification_rpc_failed chain=... pool=... gauge=... token_ids=...` for the same condition.

Same RPC failure mode, same fallback policy, but only one is alertable from logs. Would you consider mirroring the structured warning on the claim side (e.g. `claim_verification_fallback reason=verification_rpc_failed ...`) so dashboards covering "fallback fired, on-chain state not actually verified" pick up both code paths?

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -3050,7 +3107,34 @@ def stake_cl_lp_tokens(
             )
             return {"error": f"No position manager address found for chain {chain}"}
 
-        # Check if we need to approve the gauge for all NFTs
+        # First pass: filter out tokens already in the gauge's stake set so we
+        # can decide whether the approval and deposit txs are even needed.
+        tokens_to_stake = []
+        for token_id in token_ids:
+            is_staked = yield from self.is_cl_token_staked(
```
bennyjo: **Stake path bypasses the retry helper that unstake/claim use.**

`_filter_staked_token_ids` (used by unstake and claim) goes through `_verify_cl_token_staked_with_retry`, which retries on `None`. The first-pass loop here calls `is_cl_token_staked` directly — single shot. Same RPC, same blip, different policy: a one-shot RPC outage on the stake check sends the token straight into the "try anyway" branch (re-stake → revert) without the retry the other paths get.

Happy to leave as-is if this is deliberate (e.g. you want stake to fail fast on RPC issues rather than retry), but flagging because the asymmetry isn't obvious from the code and the comment block above this loop says "filter out tokens already in the gauge's stake set" — which the fail-open does the opposite of when verification fails.

bennyjo (commented):
Three findings — see inline. The first (tuple-unpack TypeError on the new empty-multisend return) is worth addressing before merge.

jmoreira-valory (dismissed):
All 3 of my comments resolved cleanly:

- **#1 RPC fallback observability** — fixed in `03de3053`. `MAX_VERIFY_RETRIES = 2` retry via `_verify_cl_token_staked_with_retry`, structured `unstake_verification_fallback reason=... chain=... pool=... gauge=... token_ids=...` warning emitted at every fallback path (six distinct `reason` values), docstring on `_build_unstake_lp_tokens_action_verified` documents verification as best-effort.
- **#2 All-already-staked → no-op** — fixed in `03de3053`. `stake_cl_lp_tokens` reordered to filter-first; returns `None` (no-op) when every token is already staked, builds approval+deposit only when there's actual work. Caller in `decision_making.py` distinguishes `None` (advances action) from `{"error": ...}` (real failure). Test `test_all_already_staked_returns_none_no_op` covers it.
- **#3 Code duplication** — fixed in `d47acda3`. `_filter_staked_token_ids` extracted on `LiquidityTraderBaseBehaviour`, retry logic folded in, both `_build_unstake_lp_tokens_action_verified` and `_claim_cl_rewards` use it with their own explicit fallback policy.

Nice work threading the new structured warning + retry through the helper rather than at the call sites.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Fixed in c50c57ee. `_exit_cl_pool` now returns `(None, None, None)` on the empty-multisend path so the caller's 3-tuple unpack of `pool.exit` no longer raises. Added `test_cl_pool_exit_unpacks_when_all_tokens_skipped` in `TestExit` that drives `pool.exit` end-to-end (rather than `_exit_cl_pool` directly) so the unpack itself is exercised. Updated three existing tests that hit this branch (`test_decrease_fails_skips`, `test_returns_none_tuple_when_all_tokens_skipped`, `test_skips_when_onchain_liquidity_is_zero`) to expect the 3-tuple. Latent `return None` branches on the other paths in this function flagged for a separate follow-up.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
DIvyaNautiyal07: Fixed in c50c57ee. Generalised `_log_unstake_fallback` to `_log_verification_fallback(action, reason, ...)`, the five existing unstake call sites just pass `action="unstake"` so the existing `unstake_verification_fallback` log key is preserved, and the claim path now emits `claim_verification_fallback reason=verification_rpc_failed chain=... pool=... gauge=... token_ids=...` on the RPC-failure branch. Same dashboard alert covers both paths. `test_claim_cl_emits_structured_warning_on_verification_failure` locks the new emission.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Fixed in c50c57ee. Routed through a retry helper. Added `MAX_STAKE_VERIFY_RETRIES = 2` and `_is_cl_token_staked_with_retry` on `VelodromePoolBehaviour`, mirroring the trader-side `_verify_cl_token_staked_with_retry` but living on the pool so the existing `_ConcreteVelodrome` stake tests don't need the trader behaviour mixed in. Fail-open semantics preserved (a token still gets queued if every attempt returns `None`, with a warning on exhaustion) but only after the retry, so the policy now matches unstake/claim. Tests: `test_stake_filter_retries_on_rpc_failure` for `[None, True]`, `test_stake_filter_falls_open_after_retry_exhaustion` for `[None, None]`, plus direct unit tests on the helper.

DhairyaPatel7 (approved):

OjusWiZard (approved):

bennyjo (approved):
Re-review: all three findings from my earlier review are addressed cleanly in c50c57e with regression tests for each.

1. `_exit_cl_pool` empty-multisend return is now `(None, None, None)` so `pool.exit`'s 3-tuple unpack at the caller no longer raises. New test drives `pool.exit` (not `_exit_cl_pool`) through the unpack.
2. Claim path emits `claim_verification_fallback reason=verification_rpc_failed ...` on RPC failure, mirroring the unstake side. Helper renamed `_log_unstake_fallback` → `_log_verification_fallback` and parameterized — rename consistent across all callsites.
3. Stake pre-filter now goes through `_is_cl_token_staked_with_retry` (MAX_STAKE_VERIFY_RETRIES=2), emits a structured warning when retries exhaust before queuing the deposit anyway.

No new issues introduced. Pre-existing `return None` paths elsewhere in `_exit_cl_pool` (lines 963, 1088, etc.) still have the same shape mismatch with the caller's unpack — flagged as a follow-up in the prior review, latent on main, not blocking #268.

Approving.

jmoreira-valory (approved):
All 3 of my comments resolved. Verified in code:

- **#1 fail-open guard-rails**: `_verify_cl_token_staked_with_retry` (`MAX_VERIFY_RETRIES = 2`) folded in, structured `unstake_verification_fallback` warning emitted from 6 distinct fallback sites with stable fields (`reason`, `chain`, `pool`, `gauge`, `token_ids`), docstring updated to call out best-effort semantics.
- **#2 all-already-staked**: `stake_cl_lp_tokens` reordered to filter first via `tokens_to_stake`, returns `None` (no-op) when empty with INFO log, approval + deposit txs only built when at least one deposit will follow.
- **#3 dedup**: `_filter_staked_token_ids` extracted on `LiquidityTraderBaseBehaviour`, both `_build_unstake_lp_tokens_action_verified` and `_claim_cl_rewards` now call it, retry logic inherited.

Also good shape across the prior round (cached-liquidity fallback, empty-multisend guard, stale top-level token_id, readability nits all resolved). Ship it.

PR #271, opened 2026-04-22
fix: work around aea-helpers bin-template-path crash in build-agent-runner

DhairyaPatel7:
## Problem

`open-aea-helpers` 0.21.17 / 0.21.18 / 0.21.19 ship a broken `bin-template-path` CLI command:

```
$ uv run aea-helpers bin-template-path
  File ".../aea_helpers/pyinstaller_deps.py", line 132, in bin_template_path
    from aea_helpers import bin_template
  File ".../aea_helpers/bin_template.py", line 41, in <module>
    validation_module._CUR_DIR = Path(sys._MEIPASS) / validation_module._CUR_DIR
AttributeError: module 'sys' has no attribute '_MEIPASS'
```

`bin_template.py` is a PyInstaller entry-point template that runs `Path(sys._MEIPASS)` at module top level — valid only inside a frozen bundle. `bin_template_path()` did `from aea_helpers import bin_template` just to read `__file__`, which crashes everywhere else.

Net effect on this repo's `make build-agent-runner`: `$(shell uv run aea-helpers bin-template-path)` expands to empty and pyinstaller errors with `pyinstaller: error: the following arguments are required: scriptname` — the PyInstaller build has been silently broken since the move to aea-helpers commands.

## Fix in this repo

Replace `$(shell uv run aea-helpers bin-template-path)` in both Makefile targets (`build-agent-runner`, `build-agent-runner-mac`) with a direct `python -c` that resolves the template path via `os.path.dirname(aea_helpers.__file__)` — no module import of the template itself.

## Upstream fix

Proper fix in open-autonomy: [valory-xyz/open-autonomy#2487](https://github.com/valory-xyz/open-autonomy/pull/2487). Once that lands in a released `open-aea-helpers`, this workaround can be reverted (~2-line revert).

## Test plan

Verified locally on a sibling repo (trader) using the same Makefile pattern and the same workaround:

- `make build-agent-runner` → 60 MB Mach-O arm64 `dist/agent_runner_bin`, `--version` succeeds.
- `make check-agent-runner` → `[OK] Found 'Starting AEA' within 80s`.

Identical two-line patch committed to trader in [valory-xyz/trader#911 @ 6e83379](https://github.com/valory-xyz/trader/pull/911/commits/6e833793) and meme-ooorr in [valory-xyz/meme-ooorr#53](https://github.com/valory-xyz/meme-ooorr/pull/53).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ build-agent-runner: uv-install agent
 	--hidden-import aea_ledger_cosmos \
 	--hidden-import aea_ledger_ethereum_flashbots \
 	$(shell uv run aea-helpers build-binary-deps ./agent) \
-	--onefile $(shell uv run aea-helpers bin-template-path) \
+	--onefile $(shell uv run python -c "import aea_helpers, os; print(os.path.join(os.path.dirname(aea_helpers.__file__), 'bin_template.py'))") \
 	--name agent_runner_bin
 	./dist/agent_runner_bin --version
 
@@ -178,7 +178,7 @@ build-agent-runner-mac: uv-install  agent
 	--hidden-import aea_ledger_cosmos \
 	--hidden-import aea_ledger_ethereum_flashbots \
 	$(shell uv run aea-helpers build-binary-deps ./agent) \
-	--onefile $(shell uv run aea-helpers bin-template-path) \
+	--onefile $(shell uv run python -c "import aea_helpers, os; print(os.path.join(os.path.dirname(aea_helpers.__file__), 'bin_template.py'))") \
 	--codesign-identity "${SIGN_ID}" \
 	--name agent_runner_bin
 	./dist/agent_runner_bin --version
--- a/Makefile
+++ b/Makefile
@@ -202,6 +202,12 @@ build-agent-runner-mac: uv-install  agent
 
 .PHONY: check-agent-runner
 check-agent-runner:
+	# aea-config.yaml uses named env-var templates (${STORE_PATH:str:/data})
+	# for both the kv_store connection and the optimus_abci skill params,
+	# so a single STORE_PATH override drives both. Path-based env vars
+	# like SKILL_..._STORE_PATH / CONNECTION_..._STORE_PATH are the
+	# fallback when the template has no explicit var name and are
+	# silently ignored here — with the old target, /data (template
+	# default) bubbled through and mkdir'd on read-only FS on macOS.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -160,7 +160,7 @@ build-agent-runner: uv-install agent
 	--hidden-import aea_ledger_cosmos \
 	--hidden-import aea_ledger_ethereum_flashbots \
 	$(shell uv run aea-helpers build-binary-deps ./agent) \
-	--onefile $(shell uv run aea-helpers bin-template-path) \
+	--onefile $(shell uv run python -c "import aea_helpers, os; print(os.path.join(os.path.dirname(aea_helpers.__file__), 'bin_template.py'))") \
```
DIvyaNautiyal07, line 163: **Nit:** the `$(shell uv run python -c "...")` expression is duplicated verbatim on L181 in `build-agent-runner-mac`. Since the PR description says this workaround will be reverted once the upstream aea-helpers fix ships, extracting to a make variable would reduce the revert to a single edit and rule out drift if one target gets touched and the other does not:

```make
BIN_TEMPLATE_PATH := $(shell uv run python -c "import aea_helpers, os; print(os.path.join(os.path.dirname(aea_helpers.__file__), 'bin_template.py'))")
```

Then both targets use `--onefile $(BIN_TEMPLATE_PATH)`.

Makefile
```diff
@@ -202,6 +202,12 @@ build-agent-runner-mac: uv-install  agent
 
 .PHONY: check-agent-runner
 check-agent-runner:
+	# aea-config.yaml uses named env-var templates (${STORE_PATH:str:/data})
```
DIvyaNautiyal07, line 205: **Nit:** `${STORE_PATH:str:/data}` here is inside a tab-indented recipe line, so make parses it as a variable reference (substitution-ref syntax without `=`) before handing the line to the shell. The variable name resolves to nothing and the printed recipe line becomes `# aea-config.yaml uses named env-var templates ()`, losing the `${STORE_PATH:str:/data}` example. Verified with a minimal repro. Functionally harmless (it is a shell comment either way), but the make log reads as if the template has no default and loses the main piece of context the comment is trying to convey. Escape with `$$` so make emits it literally: `$${STORE_PATH:str:/data}`.

DIvyaNautiyal07 (approved):

LOCKhart07 (approved):

PR #281, opened 2026-05-05
fix: backstop withdrawal swap planner with on-chain ERC20 reads

DIvyaNautiyal07:
## Summary

When the SafeGlobal API returns empty/error data, `portfolio.json` ends up with an empty `portfolio_breakdown`. The withdrawal swap planner only iterated `portfolio_breakdown` to build per-asset swap actions, so any non-USDC tokens sitting idle in the safe (USDT0, USDT, oUSDT, etc.) would be stranded — the FSM would skip the swap step and leave them behind during the transfer to the user's target address.

This PR adds a second pass after the portfolio-derived loop:

- For every asset in `WHITELISTED_ASSETS[chain]` that wasn't already queued, read the on-chain ERC20 balance of the safe.
- If the balance is non-zero, build a swap-to-USDC action.
- Dedupe by token address so we never queue the same token twice.

Position-derived data (Velodrome / Uniswap NFTs etc.) continues to come from `portfolio_data` — the on-chain reads only cover uninvested whitelisted tokens. USDC and OLAS are skipped (USDC is the destination, OLAS is never swapped during withdrawal).

## Files changed

- `packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py` — `_prepare_swap_to_usdc_actions_standard` is now a generator with the on-chain backstop. Caller updated to `yield from`.
- `packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_withdraw_funds.py` — existing tests adapted to drive the generator; five new tests cover idle-token pickup, zero-balance skip, dedupe vs portfolio, USDC/OLAS skip, and missing safe address.
- Package hashes relocked.

## Test plan

- [ ] CI green (lint, mypy, pytest, coverage, package-hash check)
- [ ] Verify on a real withdrawal that swaps are queued for safe-held non-USDC whitelisted tokens


DhairyaPatel7 (commented):
.

Adamantios (dismissed):

DhairyaPatel7 (commented):
Re-review: commit `b821586` ("fix: handle None return from pool.exit") added a None-result check at `decision_making.py:1815`. The fix is incomplete — `_exit_cl_pool` can also `return None, None` (2-tuple) on per-position slippage failure (`pools/velodrome.py:1027`), and a 2-tuple slips past `is None` then crashes the same unpack with `ValueError`. Suggest normalising `_exit_cl_pool` to always return a 3-tuple (change `return None, None` → `return None, None, None`), which fixes this without further caller-side branching.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -380,85 +383,84 @@ def _prepare_exit_pool_actions(
...
+        if safe_address:
+            for whitelisted_addr, symbol in WHITELISTED_ASSETS.get(chain, {}).items():
+                key = whitelisted_addr.lower()
+                if key in (usdc_lc, olas_lc) or key in seen:
+                    continue
+                balance = yield from self._get_token_balance(
+                    chain, safe_address, whitelisted_addr
                 )
-                continue
+                if balance and balance > 0:
```
OjusWiZard: **Inconsistent dust threshold between the two passes.** The portfolio loop at line 422 skips entries with `value_usd <= 1` (USD), but the on-chain backstop here only checks `balance > 0` (raw token units). A token reported by SafeGlobal at $0.50 is filtered out, but the same token discovered via the backstop — or any whitelisted token holding 1 wei — will still get queued, producing a swap action for dust that may revert or just waste gas.

Consider either applying a comparable minimum-balance threshold here, or documenting why dust filtering intentionally doesn't apply to the backstop path.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -380,85 +383,84 @@ def _prepare_exit_pool_actions(
...
-            if value_usd <= 1:  # nosec B105
-                self.context.logger.info(
-                    f"Skipping {token_symbol} - balance too small: {value_usd}"
+            seen.add(key)
+            tokens_to_swap[token_address] = asset.get("asset") or ""
+
+        # 2. Safe-held backstop: read every whitelisted token on-chain. Picks
+        # up idle tokens that SafeGlobal failed to report.
+        if safe_address:
+            for whitelisted_addr, symbol in WHITELISTED_ASSETS.get(chain, {}).items():
```
OjusWiZard, line 422: **Sequential N RPC calls every withdrawal.** This iterates `WHITELISTED_ASSETS[chain]` and issues a `_get_token_balance` per token, even when SafeGlobal returned valid data. For Optimism that's N round trips on the happy path. Consider gating the backstop on a portfolio-looks-empty heuristic, or batching via multicall. Correctness is fine; this is a perf follow-up.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -380,85 +383,84 @@ def _prepare_exit_pool_actions(
...
-        for asset in portfolio_breakdown:
-            token_symbol = asset.get("asset")
-            self.context.logger.info(f"--- Processing asset: {token_symbol} ---")
-
+        safe_address = self.context.params.safe_contract_addresses.get(chain)
+        usdc_address = self._get_usdc_address(chain) or ""
+        usdc_lc = usdc_address.lower()
+        olas_address = (
+            self._get_olas_address(chain)
+            if hasattr(self, "_get_olas_address")
```
OjusWiZard, line 402: Minor: the `hasattr(self, "_get_olas_address")` guard suggests `_get_olas_address` is optional on the base class. If it's part of the contract, the guard is dead; if it's truly optional, a one-line comment pinning that down would help. Either way, removing the `hasattr` once the contract is settled would simplify this block.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -380,85 +383,84 @@ def _prepare_exit_pool_actions(
...
-
-            # Skip if it's already USDC
-            usdc_address = self._get_usdc_address(chain)
-            if (
-                token_address
-                and usdc_address
-                and token_address.lower() == usdc_address.lower()
-            ):
-                self.context.logger.info("Skipping USDC - it's USDC (by address)")
+            if not token_address or value_usd <= 1:  # nosec B105
```
OjusWiZard: Nit: `# nosec B105` here looks stale. B105 is bandit's hardcoded-password rule and was likely needed in a previous string-ish form of this filter; on a numeric `value_usd <= 1` check it's no longer applicable. Safe to drop.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
```diff
@@ -380,85 +383,84 @@ def _prepare_exit_pool_actions(
...
+        olas_address = (
+            self._get_olas_address(chain)
+            if hasattr(self, "_get_olas_address")
+            else None
+        ) or ""
+        olas_lc = olas_address.lower()
+
+        # token_address (preserving casing) -> symbol for logging.
+        tokens_to_swap: Dict[str, str] = {}
+        seen: Set[str] = set()
```
OjusWiZard, line 408: Nit: `tokens_to_swap` (dict) + `seen` (set) carry overlapping state — keys of `tokens_to_swap` lowercased would already give you the dedup. You preserve original casing here, but downstream `_build_swap_to_usdc_action` doesn't care about case. Could collapse to a single lowercase-keyed dict for simplicity.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_withdraw_funds.py
```diff
@@ -644,9 +661,76 @@ def test_no_has_get_olas_address(self) -> None:
...
+    def test_safe_held_backstop_dedupes_against_portfolio(self) -> None:
+        """A token already queued from portfolio_breakdown is not re-queued from the
+        on-chain backstop."""
+        ousdt = "0x1217bfe6c773eec6cc4a38b5dc45b92292b6e189"
+        obj = self._make_obj(balances={ousdt: 5_000_000})
+        obj._build_swap_to_usdc_action = MagicMock(return_value={"action": "swap"})
+
+        portfolio = {
+            "portfolio_breakdown": [
+                {"asset": "oUSDT", "address": ousdt, "value_usd": 50},
```
OjusWiZard, line 704: Nice coverage on the new paths. One small gap worth a one-line test: portfolio returns a mixed-case address (`0xABC...`) for a token whose whitelist entry is lowercase (`0xabc...`). Today's `seen` set handles it via `.lower()`, but locking that in prevents future regressions if anyone touches the dedup logic.

OjusWiZard (dismissed):
Approving — clean fix for the SafeGlobal-outage stranding bug, well-tested. Left a few inline notes (one important, the rest advisory) that are fine as follow-ups.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
DIvyaNautiyal07: Fixed in c209e88f. Backstop now reads _get_token_decimals and requires balance >= 10**decimals (one whole token) before queuing. Whitelisted assets are stablecoins so this lines up with the >$1 threshold in pass 1. If decimals can't be resolved the token is skipped instead of guessed. Added test_safe_held_backstop_skips_dust_below_one_token_unit and test_safe_held_backstop_skips_when_decimals_unavailable to lock it in.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
DIvyaNautiyal07: Deferring this. Agreed it's worth a follow-up. For Optimism it's 6 reads per withdrawal which is fine on the happy path, but multicall would collapse it to a single round trip per chain. Withdrawals are infrequent so I'd rather not gate this on a heuristic that has its own correctness questions (when do we trust the breakdown vs not). Tracking as a perf improvement, not blocking this PR.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
DIvyaNautiyal07: Deferring. The hasattr guard is the existing pattern in this file (the original swap-planner code I refactored had the same shape), and a few other call sites in the skill still use it. I'd rather not make this one place inconsistent here. Happy to do a separate cleanup PR that audits and removes them across the skill once we confirm _get_olas_address is in fact mandatory on the base.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
DIvyaNautiyal07: Dropped in c209e88f. You're right, B105 is bandit's hardcoded-password rule and doesn't apply to a numeric comparison. Stale.

packages/valory/skills/liquidity_trader_abci/behaviours/withdraw_funds.py
DIvyaNautiyal07: Deferring this one. The two-state shape is intentional for log readability: tokens_to_swap preserves original casing so the per-asset log lines ("Creating swap action for 0xabC...") and the description string passed into _build_swap_to_usdc_action match the casing the user provided. seen does the dedupe in lowercase. Collapsing to a single lowercase-keyed dict would lowercase every address in those messages too, which was a small regression I wanted to avoid. Open to revisiting if it turns out _build_swap_to_usdc_action normalizes anyway downstream.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_withdraw_funds.py
DIvyaNautiyal07: Added in c209e88f as test_safe_held_backstop_dedupes_mixed_case_address. Portfolio entry uses an upper-cased version of the same address that lives lowercase in WHITELISTED_ASSETS, and we assert only one swap action gets queued. If anyone touches the .lower() normalization later this test will fail.

DIvyaNautiyal07:
@DhairyaPatel7 you're right, the b821586 fix only caught the bare `None` return. Confirmed `_exit_cl_pool` line 1027 and `_exit_stable_volatile_pool` line 535 both returned 2-tuples on slippage failure that slipped past `is None` and crashed the 3-way unpack with ValueError.

Fixed in 9edd24a7 by normalizing all error paths in both methods to `(None, None, None)`, exactly as you suggested. Tightened the return type hint on `_exit_cl_pool` to match. Updated the existing tests that asserted the old `is None` / 2-tuple shapes.

OjusWiZard (approved):
Re-approving — c209e88f addresses the dust-threshold concern (now reads decimals and requires `balance >= 10**decimals`), drops the stale `# nosec B105`, and adds the case-mismatched dedup test. Resolved those three threads.

Three remaining open threads are intentional follow-ups, not blockers:
- Sequential N RPC calls per withdrawal (perf — multicall/gating)
- `hasattr(self, "_get_olas_address")` guard
- `tokens_to_swap` + `seen` redundancy

Bonus: 9edd24a7 pushes the `None`-return fix from `get_exit_pool_tx_hash` deeper into the Velodrome pool layer, which is a cleaner home for it.

DhairyaPatel7 (approved):

PR #283, opened 2026-05-06
fix: route WITHDRAWAL_INITIATED from all in-cycle rounds (#279)

### LINKED ISSUE #279: FSM audit: WITHDRAWAL_INITIATED transition unreachable from 5 of 8 rounds + minor cleanups
## Summary

FSM audit of `packages/valory/skills/liquidity_trader_abci/` and `packages/valory/skills/optimus_abci/` found one system-level bug and a handful of cleanups. All four `autonomy analyse` checks (`fsm-specs`, `handlers`, `dialogues`, `docstrings`) pass — the issues below are semantic and not caught by the structural validators.

---

## High — `WITHDRAWAL_INITIATED` transition is unreachable from 5 of 8 rounds

The transition function declares `Event.WITHDRAWAL_INITIATED → WithdrawFundsRound` for 8 rounds (`rounds.py:181, 192, 203, 210, 218, 228, 241, 250`), but the round logic actually emits this event in only 3. In the other 5 rounds, the wired transition can never fire — a withdrawal request that arrives while the FSM is in one of those rounds is silently ignored until the cycle eventually loops back to `FetchStrategiesRound`.

| Round | Status | Reason |
|---|---|---|
| `FetchStrategiesRound` | works | `end_block` checks `payload.get("event") == Event.WITHDRAWAL_INITIATED.value` (`states/fetch_strategies.py:59`) |
| `EvaluateStrategyRound` | works | `end_block` checks `actions_data.get("event") == Event.WITHDRAWAL_INITIATED.value` (`states/evaluate_strategy.py:62`) |
| `DecisionMakingRound` | works | `end_block` does `event = Event(payload["event"])` (`states/decision_making.py:57`) |
| `CallCheckpointRound` | broken | dead check (H-1a) |
| `CheckStakingKPIMetRound` | broken | dead check (H-1b) |
| `APRPopulationRound` | broken | dead class attribute (H-1c) |
| `GetPositionsRound` | broken | dead class attribute (H-1c) |
| `PostTxSettlementRound` | broken | dead class attribute (H-1c) |

### H-1a — `CallCheckpointRound`: dead `WITHDRAWAL_INITIATED` check

`states/call_checkpoint.py:64-65`

```python
res = super().end_block()
...
synced_data, event = cast(Tuple[SynchronizedData, Event], res)

# Check if this is a withdrawal initiation event
if event == Event.WITHDRAWAL_INITIATED:
    return synced_data, Event.WITHDRAWAL_INITIATED
```

`super().end_block()` is `CollectSameUntilThresholdRound.end_block` (`abstract_round_abci/base.py:1631-1657`), which only ever returns `done_event`, `none_event`, or `no_majority_event`. It does not consult any custom class attribute, so `event == Event.WITHDRAWAL_INITIATED` is unreachable.

**Fix:** detect withdrawal the way `FetchStrategiesRound` does — read the most-voted payload before calling `super().end_block()` and short-circuit on `event == "withdrawal_initiated"`. The `CallCheckpointPayload` will need a way to carry that signal (a JSON-encoded `content`-style field, or an explicit `event: Optional[str] = None` field). Or remove the dead check and the `Event.WITHDRAWAL_INITIATED` entry from this round's transition map if withdrawal mid-checkpoint isn't actually intended.

### H-1b — `CheckStakingKPIMetRound`: dead `WITHDRAWAL_INITIATED` check

`states/check_staking_kpi_met.py:64-65` — identical dead-check pattern, same fix as H-1a.

### H-1c — `APRPopulationRound`, `GetPositionsRound`, `PostTxSettlementRound`: dead `withdrawal_initiated` class attribute

- `states/apr_population.py:41`
- `states/get_positions.py:41`
- `states/post_tx_settlement.py:61`

Each declares

```python
withdrawal_initiated: Event = Event.WITHDRAWAL_INITIATED
```

as a class attribute. The framework's `CollectSameUntilThresholdRound` only consults `done_event`/`none_event`/`no_majority_event` — `withdrawal_initiated` is not part of `extended_requirements` (`abstract_round_abci/base.py:1590-1596`) and is never read. `APRPopulationRound` and `GetPositionsRound` have no `end_block` override, so they can never emit this event. `PostTxSettlementRound` has a custom `end_block` but it dispatches purely on `tx_submitter`, never on a withdrawal signal.

**Fix:** remove the unused attribute and the corresponding `Event.WITHDRAWAL_INITIATED → WithdrawFundsRound` transition entries (if these rounds shouldn't honor withdrawal), or implement a custom `end_block` that detects the withdrawal payload (analogous to `FetchStrategiesRound`).

---

### M-1 — `SynchronizedData.actions` return type is inconsistent with runtime

`states/base.py:147-153` annotates the property `Optional[List[Dict[str, Any]]]`, but `EvaluateStrategyBehaviour.async_act` (`behaviours/evaluate_strategy.py:99-109`) writes a JSON-serialized **dict** (`{"event": "withdrawal_initiated", "updates": {}}`) into the same DB key, so `json.loads` can return either a list or a dict. `EvaluateStrategyRound.end_block` relies on `isinstance(actions_data, dict)` succeeding in the dict case, so this is not a runtime bug — but the annotation is misleading and mypy-strict will not catch a downstream consumer that assumes a list.

**Fix:** widen the annotation to `Union[List[Dict[str, Any]], Dict[str, Any], None]` (and update consumers), or normalize the schema so `actions` is always a list and the withdrawal signal travels in a separate field.

---

## Low (cleanup)

- **L-1:** `DecisionMakingRound` also defines `withdrawal_initiated: Event = Event.WITHDRAWAL_INITIATED` (`states/decision_making.py:48`), but its custom `end_block` already routes via `Event(payload["event"])`. The class attribute is unused — drop it.
- **L-2:** Unused `ERROR_PAYLOAD = {}` in `states/apr_population.py:45`, `states/get_positions.py:45`, `states/fetch_strategies.py:49`.
- **L-3:** `Coingecko = Coingecko` self-assignment at `optimus_abci/models.py:65`. Line 31 already imports it.
- **L-4:** Redundant `import re` inside functions at `optimus_abci/handlers.py:155` and `:2171`. Already imported at module level (line 24).
- **L-5:** `WithdrawFundsRound.selection_key = (get_name(SynchronizedData.actions),)` (`states/withdraw_funds.py:46`) is dead config — the round overrides `end_block` and never calls `super().end_block()`. The naming mismatch with `WithdrawFundsPayload.withdrawal_actions` is also misleading.
- **L-6:** Trailing comments like `# Event.ROUND_TIMEOUT` (`states/apr_population.py:47`, `states/get_positions.py:47`, `states/evaluate_strategy.py:74`, `states/fetch_strategies.py:54`) document events not referenced in code; tend to drift.

---

## Tooling note

`autonomy analyse handlers/dialogues/docstrings` need these extra deps installed in the audit environment: `peewee`, `google-generativeai`, `openapi_core`, `w3multicall`. Worth documenting in `CONTRIBUTING.md` or pulling into `pyproject.toml` so CI catches handler/dialogue regressions without manual setup.

---

DIvyaNautiyal07:
## Summary

Fixes [#279](https://github.com/valory-xyz/optimus/issues/279). The transition map declared `Event.WITHDRAWAL_INITIATED -> WithdrawFundsRound` for 8 rounds, but only `FetchStrategiesRound`, `EvaluateStrategyRound`, and `DecisionMakingRound` actually emitted it. From `CallCheckpointRound`, `CheckStakingKPIMetRound`, `GetPositionsRound`, and `APRPopulationRound`, a withdrawal request issued mid-cycle was silently held until the FSM looped back.

- Adds an optional trailing `event` field to the four broken-round payloads, peeks at `most_voted_payload_values[-1]` in each round's `end_block` before `super()`, and gates each behaviour's `async_act` on `investing_paused` so it emits a withdrawal payload when the flag is set. `FetchStrategiesBehaviour` now also gates, matching its round's existing detection.
- Hoists `_read_investing_paused` (5 copies, the audit found 3) into `LiquidityTraderBaseBehaviour`.
- Cleanups: drops unused `ERROR_PAYLOAD` in `apr_population.py` and `fetch_strategies.py`, redundant `import re` in `optimus_abci/handlers.py`, and trailing event-list comments.

## Why PostTxSettlementRound is intentionally excluded

The original audit listed `PostTxSettlementRound` among the broken rounds with a dead `Event.WITHDRAWAL_INITIATED -> WithdrawFundsRound` transition. We drop the transition rather than wire detection. A few reasons:

1. **PostTxSettlement is a routing round, not a decision point.** Its only job is to dispatch on `synced_data.tx_submitter` back to the caller round (CallCheckpoint after a checkpoint tx, CheckStakingKPI after a vanity tx, DecisionMaking after an action tx, FetchStrategies after a transfer or withdrawal-completion tx). It has no agent-side decision to make.
2. **A tx has just settled at this point.** Funds are already on-chain. There is nothing to preempt. Routing to WithdrawFunds from here would skip the tx-result accounting that the dispatch performs (e.g., `period_number_at_last_cp = period_count` after a checkpoint tx).
3. **Withdrawal is honored at the very next opportunity.** Whichever round PostTxSettlement dispatches to is now withdrawal-aware (CallCheckpoint, CheckStakingKPI, DecisionMaking, FetchStrategies all have the gate). So a withdrawal flag set during tx settlement is picked up one round later (~30s with the default `ROUND_TIMEOUT`). No loss of responsiveness.
4. **No clean payload schema.** `PostTxSettlementPayload.content: str` is unused by `end_block` (which dispatches purely on `tx_submitter`). Adding withdrawal detection here would require either overloading the existing field or adding a new one solely to carry a signal that the next round will catch anyway.

## Notes on the audit cleanups (L-1, L-3, L-5, M-1)

Each of these was verified against the code before applying. Some were partially or fully reverted:

- **L-1** (drop `withdrawal_initiated` class attr from `DecisionMakingRound`): kept the attribute. `DecisionMakingRound.end_block` dispatches via `Event(payload[\"event\"])`, which `check-abciapp-specs` cannot statically trace. Without the attribute the static check fails.
- **L-3** (drop `Coingecko = Coingecko` self-assign in `optimus_abci/models.py`): kept the self-assign. `optimus_abci/skill.yaml` references `Coingecko` by name via the framework's `class_name` lookup, so the symbol must appear in the module surface; the bare import alone trips flake8 F401.
- **L-5** (drop `selection_key` on `WithdrawFundsRound`): kept it. The framework's `extended_requirements` enforces `selection_key` via `hasattr` at class creation time; removing it breaks instantiation.
- **M-1** (widen `SynchronizedData.actions` annotation): reverted. Widening to `Union[List, Dict, None]` rippled into mypy errors at three consumers in `decision_making.py` that call `.insert()` / iterate as if list-only. Out of scope for this bug fix; left as-is.

## Test plan

- [x] `pytest packages/valory/skills/liquidity_trader_abci/tests/ packages/valory/skills/optimus_abci/tests/` (3083 pass, 13 new tests covering the round-level peek, the behaviour gate on-paused/off-paused paths, and the transition-map invariants)
- [x] `tomte check-code` (black, isort, flake8, mypy, pylint, darglint all green)
- [x] `tox -e check-abciapp-specs`
- [x] `tox -e check-abci-docstrings`
- [x] `tox -e check-handlers`
- [x] `tox -e check-packages`
- [x] `tox -e check-hash` after `autonomy packages lock`
- [x] `tox -e check-doc-hashes`
- [x] `tox -e check-dependencies`
- [x] `tox -e copyright-check`
- [x] `autonomy push-all` (packages published to IPFS so CI's `autonomy packages sync` resolves)


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1613,6 +1613,28 @@ def _read_kv(
...
+            investing_paused_value = result.get("investing_paused")
+            if investing_paused_value is None:
+                self.context.logger.warning(
+                    "investing_paused value is None in KV store"
+                )
+                return False
+
+            return investing_paused_value.lower() == "true"
+        except Exception as e:
+            self.context.logger.error(f"Error reading investing_paused flag: {str(e)}")
```
OjusWiZard: **Critical — fails open on withdrawal request.** A bare `except Exception → return False` here means any KV read fault (connection drop, dialogue/protocol error, `AttributeError` if `result` is not a dict, `ValueError`/`AttributeError` from `.lower()` on a non-string, etc.) is logged and silently treated as "not paused". The user pressed Withdraw; the agent keeps deploying capital with no surfacing.

Please narrow the catch to the specific exceptions you actually expect from `_read_kv` / dict access (e.g. `(KeyError, AttributeError, ValueError)` plus whichever the KV connection raises), and either re-raise the rest or escalate via an error event so a corrupt/unreadable flag is visible rather than ignored. Distinguishing "absent" (legitimate first-run) from "unreadable" (error) would also help.

packages/valory/skills/liquidity_trader_abci/states/apr_population.py
```diff
@@ -38,10 +41,15 @@ class APRPopulationRound(CollectSameUntilThresholdRound):
...
-    withdrawal_initiated: Event = Event.WITHDRAWAL_INITIATED
     collection_key = get_name(SynchronizedData.participant_to_context_round)
     selection_key = get_name(SynchronizedData.context)
 
-    ERROR_PAYLOAD = {}
+    def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
+        """Process the end of the block."""
+        if (
+            self.threshold_reached
+            and self.most_voted_payload_values[-1] == Event.WITHDRAWAL_INITIATED.value
```
OjusWiZard: **Critical — implicit contract on payload field order.** `most_voted_payload_values` is `astuple(self)[3:]`, so `[-1]` is correct only as long as `event` stays the trailing dataclass field. Appending any field downstream silently misroutes withdrawals with no startup signal.

Suggest either:
- a `ClassVar[int]` slot computed once from `fields(APRPopulationPayload)` (`next(i for i, f in enumerate(fields(...)) if f.name == "event")`), or
- a shared `peek_withdrawal_event(round_)` helper that resolves the field by name.

The helper option also factors out the four near-identical peek blocks across `apr_population` / `call_checkpoint` / `check_staking_kpi_met` / `get_positions`.

packages/valory/skills/liquidity_trader_abci/states/call_checkpoint.py
```diff
@@ -54,16 +54,18 @@ class CallCheckpointRound(CollectSameUntilThresholdRound):
 
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
         """Process the end of the block."""
+        if (
+            self.threshold_reached
+            and self.most_voted_payload_values[-1] == Event.WITHDRAWAL_INITIATED.value
```
OjusWiZard: Same `[-1]` positional-index concern as `apr_population.py:51`. `CallCheckpointPayload` already inherits a long field chain (`tx_submitter → tx_hash → safe_contract_address → chain_id → service_staking_state → event`) — adding any field after `event` would silently break this peek. Please use a name-based lookup or shared helper.

packages/valory/skills/liquidity_trader_abci/states/check_staking_kpi_met.py
```diff
@@ -54,16 +54,18 @@ class CheckStakingKPIMetRound(CollectSameUntilThresholdRound):
 
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
         """Process the end of the block."""
+        if (
+            self.threshold_reached
+            and self.most_voted_payload_values[-1] == Event.WITHDRAWAL_INITIATED.value
```
OjusWiZard: Same `[-1]` positional-index concern. `CheckStakingKPIMetPayload` field chain is long enough that the trailing-position assumption is non-obvious from the call site. Resolve `event` by name or via a shared helper.

packages/valory/skills/liquidity_trader_abci/states/get_positions.py
```diff
@@ -38,10 +41,17 @@ class GetPositionsRound(CollectSameUntilThresholdRound):
...
     selection_key = get_name(SynchronizedData.positions)
 
-    ERROR_PAYLOAD = {}
+    ERROR_PAYLOAD: dict = {}
+
+    def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
+        """Process the end of the block."""
+        if (
+            self.threshold_reached
+            and self.most_voted_payload_values[-1] == Event.WITHDRAWAL_INITIATED.value
```
OjusWiZard: Same `[-1]` concern. Also: `ERROR_PAYLOAD: dict = {}` is added here while the equivalent line is *removed* from `apr_population.py` — please align (either both keep it or both drop it) and parameterise the annotation (`Dict[str, Any]` / `dict[str, Any]`) for mypy strict.

packages/valory/skills/liquidity_trader_abci/tests/states/test_apr_population.py
```diff
@@ -21,7 +21,79 @@
 
 # pylint: skip-file
 
+from unittest.mock import MagicMock, PropertyMock, patch
+
+from packages.valory.skills.abstract_round_abci.base import (
+    CollectSameUntilThresholdRound,
+)
```
OjusWiZard, line 28: **Important — test gap.** The four new `end_block` peeks are only exercised with fully-shaped tuples. Worth adding one parametrised case per round for `most_voted_payload_values=()` (would currently raise `IndexError` after `threshold_reached`) and one for an unknown event string (e.g. `"withdrawal_init"`) → expected fall-through to `super()`. Locks the contract against future payload-arity changes.

OjusWiZard (changes requested):
Thanks for the thorough write-up — the FSM-level invariant test in `test_rounds.py::TestWithdrawalTransitionMap` and the `side_effect=AssertionError` gate tests are exactly the right shape for this fix, and the `PostTxSettlementRound` exclusion rationale is well-argued.

Two issues I'd like resolved before merge:

**C1. `_read_investing_paused` fails open on every exception.** A bare `except Exception → return False` means any KV fault (connection drop, dialogue error, non-string value, etc.) is silently logged and treated as "not paused" — i.e. the user pressed Withdraw and the agent keeps deploying capital. Narrow the catch to the specific KV/parse errors and let the rest propagate (or emit an error event), so a corrupt flag is visible.

**C2. `most_voted_payload_values[-1]` is an implicit, unenforced contract.** `BaseTxPayload.values` returns dataclass fields in declaration order; the peek only works while `event` stays the trailing field. `CallCheckpointPayload` / `CheckStakingKPIMetPayload` already inherit a chain of fields — appending one more anywhere downstream silently misroutes withdrawals with zero startup signal. Resolve `event` by name (a `ClassVar[int]` slot computed once from `fields(PayloadClass)`, or a small `peek_withdrawal_event(round_)` helper that reads by name). The helper would also DRY up the four near-identical guards.

Plus a couple of advisories below. Inline comments follow.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Narrowed in 1e1b1bb2. Dropped the bare `except Exception`. The three known cases are now handled explicitly:

- `_read_kv` returns None (KV unreachable): `logger.error` + `return False`
- value is None (legitimate first-run, flag never written): `return False` with no log noise
- value is not a str (corrupt KV entry): `logger.error` + `return False`

Anything else propagates. The FE-visible withdrawal response is preserved on transient failures, but "absent" and "unreadable" are now distinguishable in the agent log per your suggestion.

packages/valory/skills/liquidity_trader_abci/states/apr_population.py
DIvyaNautiyal07: Resolved by name in 1e1b1bb2. Added a small helper `peek_withdrawal_event(round_)` in [`states/base.py`](https://github.com/valory-xyz/optimus/blob/fix/issue-279-withdrawal-routing/packages/valory/skills/liquidity_trader_abci/states/base.py) that reads the field index from `dataclasses.fields(payload_class)`. The four withdrawal-aware rounds call the helper, so adding fields anywhere on the payload no longer silently misroutes withdrawals. Also factored out the four near-identical guard blocks as you suggested.

packages/valory/skills/liquidity_trader_abci/states/call_checkpoint.py
DIvyaNautiyal07: Same fix in 1e1b1bb2. Now uses the shared `peek_withdrawal_event` helper from `states/base.py` which resolves `event` by name via `fields(CallCheckpointPayload)`. The trailing-position assumption is gone.

packages/valory/skills/liquidity_trader_abci/states/check_staking_kpi_met.py
DIvyaNautiyal07: Same fix as the others, in 1e1b1bb2. Field is resolved by name via the shared `peek_withdrawal_event` helper.

packages/valory/skills/liquidity_trader_abci/states/get_positions.py
DIvyaNautiyal07: Two-part response:

**`[-1]` peek:** fixed in 1e1b1bb2 via the shared `peek_withdrawal_event` helper, same as the other three rounds.

**`ERROR_PAYLOAD` alignment:** kept here, not added to `apr_population.py`. The asymmetry is by design: `GetPositionsRound.ERROR_PAYLOAD` is consumed by `behaviours/get_positions.py:69` (uses it as the fallback positions value when `get_positions()` returns None) and by two test sites. APRPopulationRound has no such consumer, so adding the constant there would be dead code. Annotation tightened to `Dict[str, Any]` for mypy strict per your nit.

packages/valory/skills/liquidity_trader_abci/tests/states/test_apr_population.py
DIvyaNautiyal07: Both cases added in 1e1b1bb2.

**Empty-tuple case** (`payload_values=()`): one test per round under `tests/states/` — `test_end_block_empty_payload_values_falls_through`. The new helper returns None when the resolved index is out of range of the tuple, so the round falls through to `super()` cleanly without an IndexError.

**Unknown event string** (e.g. `"withdrawal_init"`): one test per round, `test_end_block_unknown_event_string_falls_through`. Confirms typo / unrecognised events do not match `Event.WITHDRAWAL_INITIATED.value` and the round dispatches via its normal `super()` path.

Also added direct unit tests for `peek_withdrawal_event`'s defensive branches (missing `payload_class`, payload class without `event` field) in `tests/states/test_base.py::TestPeekWithdrawalEvent` to keep coverage at 100%.

DIvyaNautiyal07:
Thanks for the careful review @OjusWiZard. Both critical issues plus the advisory and test gap are addressed in 1e1b1bb2.

**C1** (`_read_investing_paused` fails open): the bare `except Exception` is gone. The three known cases are handled explicitly with ERROR-level logs and `return False`, which preserves the FE-visible withdrawal response on transient KV failures while making corrupt or unreadable flags visible in monitoring. Anything truly unexpected from `_read_kv` now propagates rather than being silently swallowed.

**C2** (implicit `[-1]` contract): added `peek_withdrawal_event(round_)` in `states/base.py` that resolves the `event` field by name from `fields(payload_class)`. The four withdrawal-aware rounds call the helper, so adding fields anywhere on the payload no longer silently misroutes withdrawals. This also DRY's up the four near-identical guard blocks as you suggested.

**Advisory** (ERROR_PAYLOAD): tightened to `Dict[str, Any]`. Kept it on `GetPositionsRound` (consumed by behaviour and tests) but not added to `APRPopulationRound` (no consumer there). The asymmetry is intentional and called out in the inline reply.

**Test gap**: added empty-tuple and unknown-event-string cases for each of the four fixed rounds, plus direct tests for `peek_withdrawal_event`'s defensive branches.

3094 tests pass, 100% coverage maintained, all lint and abci checks green.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1613,6 +1613,29 @@ def _read_kv(
...
+        result = yield from self._read_kv(("investing_paused",))
+        if result is None:
+            self.context.logger.error(
+                "KV store unreachable while reading investing_paused"
+            )
+            return False
+
+        raw = result.get("investing_paused")
+        if raw is None:
+            return False
```
OjusWiZard, line 1630: _Advisory (non-blocking)_ — when `result.get("investing_paused")` returns `None` (key absent or never written), the helper silently returns `False` with no log line. The `result is None` path now logs at `error`, but this branch is the one that's truly silent. Pre-existing gap, not a regression from this commit. A `logger.debug` (or `info`) here would help operators distinguish "key never set" from "key explicitly false" in production logs. Optional follow-up.

packages/valory/skills/liquidity_trader_abci/tests/states/test_base.py
```diff
@@ -348,3 +357,37 @@ def test_participant_to_withdraw_funds(self) -> None:
...
+
+@dataclass(frozen=True)
+class _PayloadWithoutEvent(BaseTxPayload):
+    """A subclass without an `event` field, for negative-path coverage."""
+
+    value: str = ""
+
+
+class TestPeekWithdrawalEvent:
+    """Direct tests for the peek_withdrawal_event helper."""
```
OjusWiZard, line 370: _Advisory (non-blocking)_ — `TestPeekWithdrawalEvent` directly covers `payload_class=None` and missing-`event`-field, but the `not threshold_reached → None` first guard is only reached transitively via per-round threshold-not-reached tests. A one-liner here would make the helper's contract self-documenting in its own test file. Cosmetic; coverage is already achieved.

OjusWiZard (dismissed):
Both prior blockers resolved cleanly — thanks for the careful follow-up.

- **C1**: bare `except Exception → return False` is gone; the `isinstance(raw, str)` check makes the contract explicit and `test_unexpected_exception_propagates` locks in the new behaviour.
- **C2**: `peek_withdrawal_event` resolves `event` by name via `dataclasses.fields(payload_class)`. Verified the index math: `BaseTxPayload` declares `sender/round_count/id_` at positions 0–2 and `values = astuple(self)[3:]`, so `field_names.index("event") - 3` is correct for all four payload classes (incl. `CallCheckpointPayload` through the `MultisigTxPayload` chain).
- Helper coverage is complete across the six branches: `payload_class is None`, missing `event` field, empty/short values tuple, threshold-not-reached, withdrawal-hit, and unknown-event fall-through (split between `tests/states/test_base.py` and the four per-round test files).
- `ERROR_PAYLOAD: Dict[str, Any]` annotation aligned. 

Approving. Two non-blocking advisories inline.

DhairyaPatel7:
Both prior critical points (C1, C2) are addressed. Two follow-on items I'd suggest landing in this PR rather than a follow-up — neither anchors to a changed line, hence the top-level comment.

---

**1. Sibling function `_read_withdrawal_status` (`behaviours/decision_making.py:136-143`) has the exact pattern C1 just narrowed for `_read_investing_paused`.**

```python
def _read_withdrawal_status(self) -> Generator[None, None, str]:
    try:
        result = yield from self._read_kv(("withdrawal_status",))
        return result.get("withdrawal_status", "unknown")  # AttributeError if result is None
    except Exception as e:
        self.context.logger.error(...)
        return "unknown"
```

- `result.get(...)` will `AttributeError` when `_read_kv` returns `None`, then the bare `except Exception` swallows it and returns `"unknown"`.
- The bare catch swallows connection drops, dialogue/protocol errors, anything `_read_kv` could raise.

Net effect: a transient KV failure during a real withdrawal makes the DecisionMaking gate (line 95) fall through, and the agent re-deploys capital instead of routing to WithdrawFunds. Same severity as C1, sister function in the same file, called from the gate this PR is hardening. Mirror the pattern from `_read_investing_paused` (`behaviours/base.py:1614-1638`): handle the known cases (`result is None`, `value is None` / unexpected type) explicitly with `logger.error` + safe default, and let truly unexpected exceptions propagate.

---

**2. Undocumented asymmetry between DecisionMaking's gate and the 5 new gates.**

Every other withdrawal-aware behaviour — CallCheckpoint / CheckStakingKPIMet / GetPositions / APRPopulation / FetchStrategies, plus the pre-existing EvaluateStrategy — gates on a single condition:

```python
if investing_paused: → emit WITHDRAWAL_INITIATED
```

DecisionMaking (line 92-95) adds a second clause:

```python
if investing_paused AND withdrawal_status == "INITIATED": → emit WITHDRAWAL_INITIATED
```

This is *probably* intentional — the lifecycle is `INITIATED → WITHDRAWING → COMPLETED`, `investing_paused` stays `"true"` throughout, and DecisionMaking is the action executor that needs to drain queued withdrawal actions during the WITHDRAWING phase rather than preempt itself back to WithdrawFunds. The other rounds preempt unconditionally because they have no queued work to drain.

But nothing in the code or PR description says this, and the obvious next maintenance edit is "these all look the same except DecisionMaking, normalise them" — which would silently break the WITHDRAWING-phase action drain. Please add a one-line comment at `decision_making.py:92` explaining why the stricter check is deliberate.

packages/valory/skills/liquidity_trader_abci/states/base.py
```diff
@@ -20,16 +20,42 @@
...
+    payload_class = round_obj.payload_class
+    if payload_class is None:
+        return None
+
+    field_names = [f.name for f in fields(payload_class)]
+    if "event" not in field_names:
+        return None
+
+    # BaseTxPayload.values is astuple(self)[3:], dropping sender / round_count / id_.
+    event_index = field_names.index("event") - 3
```
LOCKhart07: Follow-up on C2 — the trailing position is now name-safe, but the leading offset is still a magic number that duplicates the framework's `BaseTxPayload.values` `excluded = 3`. If `BaseTxPayload` ever gains a 4th base field, all four round peeks silently misroute again — exactly the brittleness this helper was meant to eliminate.

Trivial fix:

```python
from packages.valory.skills.abstract_round_abci.base import BaseTxPayload
_BASE_FIELD_COUNT = len(fields(BaseTxPayload))
...
event_index = field_names.index("event") - _BASE_FIELD_COUNT
```

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Added in 5b985545. The absent-key path now logs at `debug` level (`investing_paused key absent in KV store; treating as not paused`), so operators can distinguish "key never written" from "key explicitly false" without adding noise to the warn/error stream during normal first-run startup.

packages/valory/skills/liquidity_trader_abci/tests/states/test_base.py
DIvyaNautiyal07: Added `test_returns_none_when_threshold_not_reached` in `TestPeekWithdrawalEvent` in 5b985545. The first guard is now exercised in the helper's own test file rather than only transitively via per-round threshold-False tests.

packages/valory/skills/liquidity_trader_abci/states/base.py
DIvyaNautiyal07: Good catch. Replaced the magic `3` with `_BASE_PAYLOAD_FIELDS = len(fields(BaseTxPayload))` in 5b985545. The helper now derives the offset from the framework's own dataclass rather than mirroring `BaseTxPayload.values`'s `excluded = 3` constant, so adding or removing a base field on `BaseTxPayload` keeps the four round peeks correct without any change here.

DIvyaNautiyal07:
Thanks @DhairyaPatel7. Both points landed in 5b985545.

**1. `_read_withdrawal_status` mirrored.** You were right — same fail-open pattern as C1, and the `AttributeError` from `result.get(...)` on a None result was indeed silently swallowed. Mirrored the C1 shape:

```python
def _read_withdrawal_status(self) -> Generator[None, None, str]:
    result = yield from self._read_kv(("withdrawal_status",))
    if result is None:
        self.context.logger.error(
            "KV store unreachable while reading withdrawal_status"
        )
        return "unknown"
    raw = result.get("withdrawal_status")
    if raw is None:
        return "unknown"
    if not isinstance(raw, str):
        self.context.logger.error(...)
        return "unknown"
    return raw
```

KV unreachable now logs at error and returns "unknown"; absent key returns "unknown" silently (legitimate first-run); non-string value logs and returns "unknown"; truly unexpected exceptions propagate. Added matching tests: `test_returns_unknown_when_kv_unreachable`, `test_returns_unknown_when_value_non_string`, and `test_unexpected_exception_propagates`.

**2. DecisionMaking gate asymmetry comment.** Added at `decision_making.py:91-94`:

```python
investing_paused = yield from self._read_investing_paused()
if investing_paused:
    # DecisionMaking only preempts on INITIATED, not on every paused
    # state. Sister rounds preempt unconditionally; DecisionMaking
    # must keep running during WITHDRAWING to drain the queued
    # exit/swap/transfer actions back through tx settlement.
    withdrawal_status = yield from self._read_withdrawal_status()
    if withdrawal_status == "INITIATED":
```

Captures the lifecycle reasoning you described so a future "normalise these gates" edit doesn't silently break the WITHDRAWING-phase action drain.

LOCKhart07 (approved):

DhairyaPatel7 (approved):

Issue #280, opened 2026-05-03
Resilience audit: P0 Web3 timeouts, verify=False, _estimate_gas BP11; P1 mirror_db + handler dispatch


DavidMinarsch:
## Summary

External-request resilience audit of the project-owned packages: skills `liquidity_trader_abci` + `optimus_abci`, connection `mirror_db`, and customs (`asset_lending`, `balancer_pools_search`, `merkl_pools_search`, `uniswap_pools_search`, `velodrome_pools_search`, `max_apr_selection`). Framework library skills/connections are out of scope.

Three operational impact categories:
- **CRASH** — no agent-process-killing crash paths verified.
- **STUCK** — Web3 RPC partition is the highest-impact stuck scenario; affects FSM behaviours and the executor backing HTTP handlers.
- **SIDE-EFFECT** — `verify=False` MITM exposure, `_estimate_gas` return-type bug, and possible duplicate on-chain swap submissions can affect financial outcomes.

---

### P0-1 — `Web3.HTTPProvider` instantiated without a timeout

`Web3.HTTPProvider(rpc_url)` has no client-level timeout. Calls like `get_transaction_count`, `gas_price`, `estimate_gas`, contract `.call()` inherit the underlying socket default and can hang indefinitely on a partition.

Sites:
- `packages/valory/customs/balancer_pools_search/balancer_pools_search.py:302`
- `packages/valory/customs/velodrome_pools_search/velodrome_pools_search.py:407`
- `packages/valory/skills/optimus_abci/handlers.py:455` (`_get_web3_instance`)

Aggravating factor: in `optimus_abci/handlers.py` the Web3 calls run inside a `ThreadPoolExecutor(max_workers=1)` (line 275). One hang stalls the entire executor and starves any future submissions.

**Fix:** pass `request_kwargs={"timeout": N}` to every `Web3.HTTPProvider(...)`. Consider a small helper `make_web3(rpc_url)` that always sets a sane default.

### P0-2 — `verify=False` on direct `requests.get` calls

All four direct `requests.get(...)` calls in `behaviours/fetch_strategies.py` disable TLS certificate verification:

- `packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py:3502`
- `:3650`
- `:4725`
- `:4838`

[...]

PR #284, opened 2026-05-06
fix: harden external-request resilience (issue #280, P0+P1)

DIvyaNautiyal07:
## Summary
- Six audit fixes from #280: Web3 30s timeouts, secure-default TLS verification, `_estimate_gas` return-type fix, mirror_db malformed-payload handling, mirror_db retry on 429/5xx/network errors, HTTP handler dispatch wrapped to return 500 instead of hanging.
- New `tls_verify: bool = true` param on liquidity_trader_abci, default secure; opt out via `TLS_VERIFY=false`.
- New `MirrorDBHTTPError(status_code, msg)` so retries dispatch on the status code, not on string-matching the error message.

## Scope
P0+P1 of the audit only. P2, P3, P4 (CC1/CC2/CC3) will land in a follow-up PR.

## Files
- `packages/valory/skills/optimus_abci/handlers.py` — `_estimate_gas` returns `None`; Web3 provider gets `request_kwargs={"timeout": 30}`; new `HttpCode.INTERNAL_SERVER_ERROR = 500`; new `_handle_internal_error` helper; `handle()` dispatch wrapped in try/except.
- `packages/valory/customs/balancer_pools_search/balancer_pools_search.py` and `packages/valory/customs/velodrome_pools_search/velodrome_pools_search.py` — `WEB3_HTTP_TIMEOUT_SECONDS = 30` constant; HTTPProvider gets the timeout. (Customs run via `exec()` so the constant is duplicated rather than imported.)
- `packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py` — four `verify=False, # nosec B501` call sites now read `self.params.tls_verify`.
- `packages/valory/skills/liquidity_trader_abci/models.py` — new `tls_verify: bool` param via `kwargs.get("tls_verify", True)`.
- `packages/valory/skills/liquidity_trader_abci/skill.yaml`, `packages/valory/services/optimus/service.yaml`, `packages/valory/agents/optimus/aea-config.yaml` — `tls_verify` declared, defaulted to True.
- `packages/valory/connections/mirror_db/connection.py` — `json.loads` of the SRR payload guarded; `_handle_done_task` swallows `task.result()` exceptions and queues a placeholder envelope; new `MirrorDBHTTPError`; retry decorator dispatches on a typed exception (429, 5xx) plus `aiohttp.ClientConnectorError`, `aiohttp.ServerTimeoutError`, `asyncio.TimeoutError`; `print()` retry/exhaustion logs replaced with `self.logger`.

## Tests added or updated
- `packages/valory/connections/mirror_db/tests/test_connection.py` — replaced rate-limit-string tests with status-code-driven retry tests (429, 5xx, 4xx-other-than-429); new `_is_retryable` parametrized tests; new tests for malformed-payload error reply and `_handle_done_task` exception swallowing.
- `packages/valory/skills/optimus_abci/tests/test_handlers.py` — updated `test_estimate_gas_no_web3` to assert `is None`; new `test_get_web3_instance_passes_timeout_to_provider`; new `test_handle_internal_error_sends_500`; new `test_handle_dispatch_catches_handler_exception`.
- `packages/valory/skills/liquidity_trader_abci/tests/test_models.py` — `test_tls_verify_defaults_to_true` and `test_tls_verify_can_be_disabled`.
- `packages/valory/customs/{balancer,velodrome}_pools_search/tests/...` — provider-construction tests asserting `request_kwargs={"timeout": 30}`.

## Test plan
- [ ] `tox -e py3.10-darwin` (or matching platform) green
- [ ] CI checks green: `copyright-check`, `check-hash`, `check-packages`, `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`, `check-dependencies`, `check-doc-hashes`, `tomte check-code`, `tomte check-security`, `uv lock --check`
- [ ] Manual smoke: confirm an agent on a deployment with valid certs comes up cleanly with `TLS_VERIFY=true` (default). Confirm `TLS_VERIFY=false` reproduces the prior bypass for environments that need it.

## Mutations my tests would not catch
1. If a refactor renamed the `request_kwargs` key (`timeout` → `request_timeout`), the timeout would be silently ignored by web3.py and my tests would still pass.
2. The Params unit test verifies `tls_verify` defaults to True, but does not verify `fetch_strategies.py` reads `self.params.tls_verify` at the four call sites; reintroducing `verify=False` at a call site would not be caught.
3. Off-by-one in the retry loop (`attempt < max_retries - 1` → `attempt < max_retries`) — my tests verify "succeeds on Nth attempt" and "always-fail raises", not the strict failure-path retry count.


packages/valory/skills/optimus_abci/handlers.py
```diff
@@ -1181,7 +1188,23 @@ def handle(self, message: Message) -> None:
...
+        except Exception as e:  # pylint: disable=broad-except
+            self.context.logger.exception(
+                f"Unhandled exception while dispatching HTTP handler for "
+                f"{http_msg.method} {http_msg.url}: {e}"
+            )
+            try:
+                self._handle_internal_error(
+                    http_msg,
+                    http_dialogue,
+                    error_msg=f"Internal server error: {e}",
```
DhairyaPatel7: **Echoes raw exception text into the HTTP response body — info-disclosure risk.**

`error_msg=f"Internal server error: {e}"` lands verbatim in `_handle_internal_error`'s `body=error_msg.encode(...)`. Whatever `str(e)` carries — file paths, query fragments, internal addresses, secret-bearing values that happened to appear in the original error — gets returned to the HTTP client. The handler already calls `logger.exception(...)` two lines above, so the full detail is preserved server-side.

For an internal-only API the threat is small; for anything externally reachable it's a leak. Suggest passing a generic body and letting the log carry the detail:

```python
self._handle_internal_error(
    http_msg, http_dialogue,
    error_msg="Internal server error",
)
```

packages/valory/connections/mirror_db/connection.py
```diff
@@ -43,26 +43,62 @@
...
+    aiohttp.ClientConnectorError,
+    aiohttp.ServerTimeoutError,
+    asyncio.TimeoutError,
+)
+
+
+def _is_retryable(exc: BaseException) -> bool:
+    """Return True if the exception is a transient failure worth retrying."""
+    if isinstance(exc, MirrorDBHTTPError):
+        return exc.status_code == 429 or 500 <= exc.status_code < 600
```
DhairyaPatel7: **Consider also retrying 408 (Request Timeout).**

429 and 5xx are retried; 408 is a server-side semantic timeout that's typically as retry-worthy as 429 / 503. RFC 7231: "the server would like to shut down this unused connection" — a fresh request should succeed. Suggest:

```python
return exc.status_code in (408, 429) or 500 <= exc.status_code < 600
```

Non-blocking.

LOCKhart07 (commented):
Five non-blocking observations — no blockers.

**1. `_handle_internal_error` body leaks `str(e)` to clients** — `packages/valory/skills/optimus_abci/handlers.py:1202`. The wrapper passes `error_msg=f"Internal server error: {e}"`, so the original exception text (paths, internal state, exception chain) is returned in the HTTP body. Matches the existing `_handle_bad_request` pattern, but here `e` is an arbitrary uncaught exception. Consider returning a generic body (e.g. `b"Internal server error"`) and keeping `str(e)` only in the log line above — the full detail is already going through `self.context.logger.exception(...)`.

**2. `_handle_done_task` queues `None` envelope on task exception** — `packages/valory/connections/mirror_db/connection.py:299`. When `task.result()` raises, `response_message` becomes `None` and the consumer of `response_envelopes` then sees a `None` envelope from `receive()`. Same observable behaviour as the prior `response_message is None` branch, but it loses the chance to deliver a typed failure. Since `request` (popped a few lines above) is still available, you could call `prepare_error_message(request.message, dialogue, str(e))` and queue a real error envelope so the requesting behaviour gets a proper error response rather than a silent dropout.

**3. Three duplicate `WEB3_HTTP_TIMEOUT_SECONDS = 30` constants** — `handlers.py:137`, `customs/balancer_pools_search/balancer_pools_search.py:89`, `customs/velodrome_pools_search/velodrome_pools_search.py:318`. The PR description acknowledges this is intentional because customs are loaded via `exec()`. Add a one-line comment at each site (`# Mirror of constant in handlers.py / customs — keep in sync`) so a future edit doesn't drift one copy and silently leave the other two on the old timeout.

**4. `models.py:450` uses `kwargs.get("tls_verify", True)` instead of `self._ensure(...)`** — surrounding params (`request_timeout`, `lifi_quote_to_amount_url`, etc.) use `self._ensure(...)` to enforce presence and type. `kwargs.get("tls_verify", True)` means a typo or missing YAML entry silently falls through to the default. Functionally fine because the secure default is True, but `self._ensure("tls_verify", kwargs, bool)` would catch a misnamed key and match the local convention.

**5. `test_handle_dispatch_catches_handler_exception` mocks `_handle_internal_error`** — `tests/test_handlers.py:1188`. The dispatch path is verified to *call* the helper, but not that the 500 response actually lands in the outbox. `test_handle_internal_error_sends_500` covers the helper directly, so coverage is adequate, but a single end-to-end test (raising handler + unmocked helper + assert `outbox.put_message` called with `status_code=500`) would tighten the integration assertion and catch refactors that decouple the two paths.

PR description is thorough and the "Mutations my tests would not catch" section is a nice touch.

packages/valory/skills/optimus_abci/handlers.py
DIvyaNautiyal07: Done in fe7fa023. Body is now the static "Internal server error". The exception detail is preserved in the self.context.logger.exception line above so server-side debugging is unchanged. Updated test_handle_dispatch_catches_handler_exception to assert the body and added test_handle_dispatch_writes_500_to_outbox_end_to_end which exercises the full dispatch -> _handle_internal_error -> outbox.put_message path without mocking the helper.

packages/valory/connections/mirror_db/connection.py
DIvyaNautiyal07: Done in fe7fa023. _is_retryable now matches 408 alongside 429 and 5xx. New test_408_is_retryable in TestIsRetryable parametrized class.

DIvyaNautiyal07:
Thanks for the careful reviews. All non-blocking items addressed in fe7fa023. Quick rundown.

LOCKhart:

1. Static 500 body (overlaps with Dhairya #1) — done. Body is now "Internal server error", exception detail stays in the log line.
2. _handle_done_task None envelope — now wraps the entire body in a defensive try/except so a duplicate-callback pop does not starve the queue, and on task failure builds a typed SRR error reply via prepare_error_message and queues that envelope. The log line now carries request.to and request.sender for traceability. Added test_handle_done_task_queues_typed_error_on_task_failure, test_handle_done_task_recovers_from_pop_failure, and test_handle_done_task_falls_back_to_none_when_build_fails.
3. Three duplicate WEB3_HTTP_TIMEOUT_SECONDS — left as-is. The customs run via exec() so they cannot import from each other or from the skill. Documenting in three places didn't seem worth it given the value is only set in one place by an audit and the type of edit that would touch this is rare.
4. tls_verify uses _ensure — done. Switched to self._ensure("tls_verify", kwargs, bool). Same change applied to the three new URL params on PR #285 (mode_conduit_explorer_url, safe_api_v1_url, mode_native_explorer_url) for consistency.
5. Dispatch end-to-end test — added test_handle_dispatch_writes_500_to_outbox_end_to_end which runs without mocking _handle_internal_error and asserts the 500 envelope reaches outbox.put_message with body == b"Internal server error".

OJ:

1. Same as LOCKhart #2 — addressed.
2. Retry decorator silent when no logger — falls back to logging.getLogger(__name__) if the wrapped object has no .logger. last_exception type tightened from Optional[BaseException] to Optional[Exception].
3. JSON parse failure unlogged — now self.logger.error before returning the SRR error reply.
4. Same as Dhairya #1 — addressed.
5. Same as LOCKhart #3 — see above.
6. Backoff progression untested — added test_backoff_progression_doubles_each_attempt that patches asyncio.sleep and asserts the captured delays are [1, 2, 4]. Also added test_logger_attribute_emits_retry_warnings which decorates a method on an object with a real MagicMock logger and asserts both the per-attempt warning and the terminal Max retries error landed.

3658 tests pass with 100% coverage on the affected packages.

LOCKhart07 (dismissed):

DhairyaPatel7 (dismissed):

DhairyaPatel7 (approved):

LOCKhart07 (approved):

PR #285, opened 2026-05-06
fix: P2/P3/P4 resilience hardening (issue #280, stacked on #284)

DIvyaNautiyal07:
## Summary
Stacked on #284. Picks up the audit's lower-tier resilience items: executor dedup, mirror_db hardening, pagination caps, hardcoded-URL extraction, CoinGecko credit persistence, Web3 read-call retries, and a per-endpoint circuit breaker. P4-CC3 (customs error format standardization) is intentionally deferred — it is operability, not a fix.

The diff against `fix/issue-280-p0-p1-resilience` shows only this PR's changes. Set the base to main once #284 merges.

## Scope
| Audit item | What it does |
| --- | --- |
| P2-1 executor dedup | Three fire-and-forget executor functions (`_ensure_sufficient_funds_for_x402_payments`, `_delayed_write_kv_extended`, `_write_withdrawal_data`) gate on a process-wide `threading.Lock` with non-blocking `acquire()`. A duplicate concurrent invocation logs and returns. Closes the financial-side-effect path where two simultaneous topup calls could submit two on-chain swaps. |
| P2-2 mirror_db `.json()` fallback | `_raise_for_response` falls back to `response.text()` when the error body is not JSON. |
| P3 `_VALID_ENDPOINTS` enforcement | The regex declared at `connection.py:108-112` was previously unused. `_get_response` now rejects any endpoint string that does not match. New `_is_valid_endpoint` classmethod. |
| P3 pagination caps | New `MAX_PAGINATION_PAGES = 100` constant. The three Mode pagination loops in `fetch_strategies.py` bail with a warning if a malicious or buggy API returns next-page cursors indefinitely. The round timeout is the primary guard; this is a backstop. |
| P3 hardcoded URLs to params | New `mode_conduit_explorer_url`, `mode_native_explorer_url`, `safe_api_v1_url` params (defaults preserve current behaviour). Two stray `https://mainnet.optimism.io` literals now read `optimism_ledger_rpc`. |
| P3 CoingeckoRateLimiter persistence | The monthly `credits_used` counter persists to a small JSON file under `store_path`. A process restart on day 25 of the month no longer silently re-grants the full 10,000-credit allowance. Per-minute counters stay in memory. |
| P4-CC1 Web3 read-call retries | New `_call_with_web3_retries` helper with exponential backoff. `_check_usdc_balance`, `_get_nonce_and_gas_web3`, and `_estimate_gas` now retry transient timeouts and connection errors but propagate contract reverts. New `_is_transient_web3_error` predicate classifies. The send-raw-transaction path is intentionally NOT wrapped to avoid re-broadcasting a tx whose response was lost. |
| P4-CC2 EndpointCircuitBreaker | New class in `models.py` with three states (CLOSED → OPEN → HALF_OPEN). After 5 consecutive failures the breaker opens for 60s; one HALF_OPEN probe decides whether to close. Lives in `SharedState.endpoint_breakers`, keyed by chain. Web3 read calls are gated through `_call_web3_with_breaker` so a sustained RPC outage stops eating retry budget every period. |

## Files
- `packages/valory/skills/optimus_abci/handlers.py` — three module locks, `_call_with_web3_retries`, `_is_transient_web3_error`, `_call_web3_with_breaker`, `CircuitBreakerOpenError` import.
- `packages/valory/connections/mirror_db/connection.py` — `re` import, `_VALID_ENDPOINTS` enforcement in `_get_response`, `_is_valid_endpoint` classmethod, `.text()` fallback in `_raise_for_response`.
- `packages/valory/skills/liquidity_trader_abci/models.py` — `EndpointCircuitBreaker`, `CircuitBreakerOpenError`, `CircuitBreakerState`, `SharedState.get_circuit_breaker`, CoingeckoRateLimiter persistence load/save, three new params.
- `packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py` — `MAX_PAGINATION_PAGES` constant, three pagination cap branches, twelve hardcoded URL replacements.
- `packages/valory/skills/liquidity_trader_abci/skill.yaml`, `services/optimus/service.yaml`, `agents/optimus/aea-config.yaml` — three new param defaults, plus `store_path` exposed to the Coingecko model.

## Tests added
- `test_models.py`: 8 tests for CoingeckoRateLimiter persistence (load, save, stale month, corrupt file, no path, OSError swallow), 7 tests for EndpointCircuitBreaker state transitions, 2 for SharedState.get_circuit_breaker.
- `test_handlers.py`: 1 test for `_is_transient_web3_error` classifications, 3 for `_call_with_web3_retries`, 3 for `_call_web3_with_breaker`, 3 for executor lock dedup (one per protected function).
- `test_connection.py`: 1 for `.text()` fallback on non-JSON bodies, 4 parametrized for valid endpoints, 5 parametrized for rejected endpoints, 1 for the `_is_valid_endpoint` helper.
- `test_fetch_strategies.py`: 3 pagination-cap tests (one per loop) using `MAX_PAGINATION_PAGES = 0` to fire the cap immediately.

3611 tests pass. 100% coverage maintained.

## CI checks (all green locally)
- `tox -e copyright-check`, `check-hash`, `check-packages`, `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`, `check-dependencies`, `check-doc-hashes`
- `tomte check-code` (black, isort, flake8, mypy, pylint, darglint)

## Test plan
- [ ] `tox -e py3.10-darwin` (or matching platform) green on CI
- [ ] Coverage ≥ 100% on changed files
- [ ] Manual smoke: with the breaker open, `_check_usdc_balance` returns None within milliseconds instead of waiting on the RPC.
- [ ] Manual smoke: kill the agent mid-month after 5,000 CoinGecko calls, restart, confirm `coingecko_credits_used.json` is read and the limiter starts at 5,000 used (not 0).

## What is intentionally NOT in this PR
- **P4-CC3 (standardize customs error format).** Operability improvement only. No active incident, no user-visible bug, doesn't change behaviour. Per project bar (fixes and features only) this belongs in a separate PR if at all.
- **Customs Web3 retries.** `_call_with_web3_retries` is wired into `optimus_abci/handlers.py` only. The customs (`balancer_pools_search.py`, `velodrome_pools_search.py`) make many `.call()` invocations for ERC20 lookups; those are LRU-cached and lower priority. Could be a follow-up.

## Mutations these tests would not catch
1. **Lock release omission.** If a `finally: _LOCK.release()` were dropped, the first call leaves the lock held forever and tests in the same process pass anyway because pytest spawns fresh state per session.
2. **Breaker key swap.** The breaker is keyed by chain name, not RPC URL. If two URLs ever map to the same chain (e.g. fallback RPC), the breaker state is shared. Tests don't verify URL-level isolation.
3. **Persistence atomicity.** `_persist_credits_used` writes to the file directly. A crash mid-write could leave a corrupt JSON. Tests verify "corrupt file is treated as zero" but don't verify the write path is atomic.


packages/valory/skills/optimus_abci/handlers.py
```diff
@@ -135,6 +137,63 @@
...
+    such as contract reverts and caller-side errors propagate immediately.
+    """
+    if isinstance(
+        exc, (requests.exceptions.Timeout, requests.exceptions.ConnectionError)
+    ):
+        return True
+    msg = str(exc).lower()
+    if "execution reverted" in msg or "return amount is not enough" in msg:
+        return False
+    if "timeout" in msg or "connection" in msg or "503" in msg or "502" in msg:
```
DhairyaPatel7: **Substring matching on `str(exc)` is brittle.**

Two concrete failure modes:

1. `"503" in msg` matches any substring `"503"`. A contract revert reason like `"value 5039 is invalid"`, an internal address fragment, or a `bytes` payload printed via `str()` would be silently misclassified as transient and retried 3×. Same hazard for `"502"`.
2. `"timeout" in msg` and `"connection" in msg` match arbitrary words appearing in the error text, including in revert reasons and human-readable diagnostics from third-party libraries.

Suggest narrowing — both for the HTTP codes and the keyword list:

```python
import re
_TRANSIENT_HTTP = re.compile(r'\b(408|429|502|503|504)\b')
...
if _TRANSIENT_HTTP.search(msg):
    return True
# keep the explicit isinstance check for requests.Timeout/ConnectionError
```

Also missing from the current set: 429 (rate-limit) and 504 (gateway timeout) — both legitimately transient, both currently classified as deterministic and not retried.

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -45,6 +47,88 @@
...
+                return True
+            if self._state == CircuitBreakerState.OPEN:
+                if (
+                    self._opened_at is not None
+                    and (time() - self._opened_at) >= self._recovery_timeout
+                ):
+                    self._state = CircuitBreakerState.HALF_OPEN
+                    return True
+                return False
+            # HALF_OPEN: only the probe that triggered the transition is in flight
```
DhairyaPatel7: **HALF_OPEN can wedge if the probe call never resolves.**

Flow: `allow()` flips OPEN→HALF_OPEN and returns True for the first caller. Subsequent callers see HALF_OPEN and get False until `on_success` / `on_failure` is invoked by the probe's caller. If the probe call hangs (no `on_*` ever fires), the breaker is stuck blocking everyone forever. The breaker has no probe deadline of its own.

Currently mitigated because `_call_with_web3_retries` caps at 3 attempts and PR #284's `request_kwargs={"timeout": 30}` bounds each attempt — worst case ~95s. So the contract is: HALF_OPEN won't wedge *as long as the lower-layer Web3 timeout stays in place*. That's an undocumented inter-PR dependency.

Suggest one of:
- Track `_half_open_at` and auto-revert HALF_OPEN→OPEN if a probe hasn't reported within `recovery_timeout * 2`. Self-healing without trusting callers.
- At minimum, add a code comment at this transition pointing out that the timeout enforcement lives at the Web3 provider layer, so a future change that drops `request_kwargs={"timeout": ...}` would also break the breaker.

LOCKhart07 (commented):
Review notes — none blocking, all quality nits.

**1. `datetime.utcnow()` deprecation — `models.py` `_current_month`**
`datetime.utcnow()` has been deprecated since Python 3.12 and emits a DeprecationWarning. The repo has a `fix-py314-deprecations` skill, so 3.14 is in scope — new code should use `datetime.now(timezone.utc).strftime(...)`.

**2. Persistence non-atomic + disk-write per burn — `models.py` `_persist_credits_used` / `_burn_credit`**
Two concerns:
- **Non-atomic**: `open(path, 'w')` truncates immediately. A crash between truncate and `json.dump` completing leaves a corrupt file. The loader catches `ValueError` and returns 0, so corruption silently re-grants the full monthly allowance — exactly the failure this PR is trying to prevent. Use a temp file + `os.replace(...)`.
- **Disk-write per burn**: `_burn_credit` calls this on every CoinGecko request, so every call now blocks on a synchronous write. Consider debouncing (write every N burns / on threshold transitions) or batching at shutdown.

**3. New URL params skip `_ensure(...)` — `models.py` Params.__init__**
`mode_conduit_explorer_url`, `safe_api_v1_url`, `mode_native_explorer_url` use `kwargs.get(...)` with a hardcoded fallback. The sibling `lifi_quote_to_amount_url` two lines above (and the rest of `Params.__init__`) uses `self._ensure(..., kwargs, str)`, which validates type and presence. Worth using `_ensure` for consistency.

**4. Pagination cap test is tautological — `test_fetch_strategies.py` `test_pagination_cap_triggers_break` (3 copies)**
Patching `MAX_PAGINATION_PAGES = 0` and asserting `requests.get` was never called only proves the immediate-break branch fires when the cap is zero — it never exercises the actual cap-after-N-pages behaviour. A test with `MAX_PAGINATION_PAGES = 2` plus a fake API that returns a `next_page_params` cursor three times would prove the cap fires after exactly the configured number of pages.

**5. Breaker-open silently swallowed in `_check_usdc_balance` — `handlers.py`**
When the breaker is open, `_call_web3_with_breaker` raises `CircuitBreakerOpenError`, the broad `except Exception` catches it and returns `None`, and the caller `_ensure_sufficient_funds_for_x402_payments` interprets `None` as "could not check, assume sufficient and skip". So a sustained Optimism RPC outage flips the breaker open and silently produces a false "funded" signal.

The behaviour is also asymmetric: `_get_nonce_and_gas_web3` and `_estimate_gas` return `None` on the same path and the caller marks the agent as **insufficient**. Worth either explicitly handling `CircuitBreakerOpenError` to distinguish breaker-open from genuine read failures, or at least adding a comment that the assume-sufficient default here is deliberate.

PR description is solid — no update needed.

packages/valory/skills/optimus_abci/handlers.py
DIvyaNautiyal07: Done in 57553622. _is_transient_web3_error now uses re.compile(r"\b(408|429|502|503|504)\b") so digits embedded in revert reasons or addresses cannot be misclassified, and the brittle "timeout"/"connection" keyword fallback is dropped. 408 and 504 are added; typed requests.Timeout and requests.ConnectionError continue to short-circuit to True before any string parsing.

packages/valory/skills/liquidity_trader_abci/models.py
DIvyaNautiyal07: Done in 57553622. EndpointCircuitBreaker now tracks _half_open_at on the OPEN -> HALF_OPEN transition; if a probe has not reported success or failure within another recovery_timeout, allow() forces the breaker back to OPEN with a fresh _opened_at so callers are no longer blocked by a wedged probe. Self-healing without trusting the caller. test_half_open_wedged_probe_auto_reverts_to_open exercises the path.

DIvyaNautiyal07:
Thanks for the review. All five items addressed in 57553622.

1. datetime.utcnow() — switched _current_month to datetime.now(timezone.utc).strftime("%Y-%m"). Tests updated.

2. Persistence atomicity + I/O per burn — write goes through a temp file and os.replace, so a crash between truncate and dump cannot leave a half-written file. _burn_credit now persists only every _PERSIST_EVERY_N_BURNS = 10 calls, so the synchronous disk write rate is capped at one per ten CoinGecko requests. Worst-case crash loses up to 9 credits of accuracy, which is below the noise floor of a soft monthly cap. Tests: test_burn_credits_persist_at_threshold, test_persist_uses_atomic_temp_then_replace, test_persist_swallows_oserror_and_cleans_tmp_file, test_persist_swallows_double_oserror, test_persist_oserror_with_no_tmp_file_present.

3. URL params via _ensure — done. mode_conduit_explorer_url, safe_api_v1_url, mode_native_explorer_url all use self._ensure(..., kwargs, str) now (sibling consistency with lifi_quote_to_amount_url etc).

4. Pagination cap test was tautological — rewrote both sites with MAX_PAGINATION_PAGES = 2 and a fake API that always returns a next-page cursor, then asserting requests.get.call_count == 2. Now exercises the cap-after-N branch instead of the cap=0 short-circuit.

5. Breaker-open silently swallowed in _check_usdc_balance — narrowed the catch-all so CircuitBreakerOpenError propagates. The caller in _ensure_sufficient_funds_for_x402_payments now catches it explicitly and sets sufficient_funds_for_x402_payments = False, removing the asymmetry where a sustained Optimism RPC outage would have produced a false "funded" signal. Tests: test_check_usdc_balance_propagates_circuit_breaker_open and test_ensure_sufficient_funds_breaker_open_marks_insufficient.

Also addressing Dhairya's two inline points:

- _is_transient_web3_error now uses re.compile(r"\b(408|429|502|503|504)\b") so digits inside revert reasons or addresses cannot be misclassified, and the brittle "timeout"/"connection" keyword fallback is dropped. 408 and 504 are added.
- EndpointCircuitBreaker tracks _half_open_at and auto-reverts HALF_OPEN to OPEN if a probe never reports back within recovery_timeout, so a wedged probe cannot block all callers.

3664 tests pass with 100% coverage on the affected packages.

LOCKhart07 (approved):

DhairyaPatel7 (approved):

DIvyaNautiyal07:
Re-review addressed in commit `a64888dc`. Five items implemented, three made moot by reverting the persistence layer, one deferred.

**1. Stale-probe corrupts breaker state.** Done. `on_success` and `on_failure` now ignore stale callbacks: if the breaker is already OPEN, the late callback returns early; HALF_OPEN→CLOSED only on success, HALF_OPEN→OPEN only on failure. `_consecutive_failures` resets only on the CLOSED path. Covered by `test_stale_probe_success_does_not_overwrite_reverted_open` and `test_stale_probe_failure_does_not_extend_cooldown`.

**2. Silent loss of up to 9 burns on crash.** Moot. The whole on-disk credit counter has been reverted. Production runs `use_x402: true` on every CoinGecko path (aea-config.yaml, service.yaml, optimus_abci/skill.yaml), so the x402 proxy enforces the monthly cap upstream. A per-process counter wasn't authoritative, and the throttle-vs-shutdown tradeoff stops mattering once the source of truth is upstream.

**3. Silent OSError on persist.** Moot — the persist path is gone.

**4. Breaker-open conflates insufficient with unknown.** Logging part done — promoted to `logger.error` with stable ID `x402-funds-check-breaker-open` for Sentry. Tri-state `funds_status: SUFFICIENT | INSUFFICIENT | UNKNOWN` deferred: the breaker recovery window is 60s and the next FSM round re-checks balance, so the worst case is one delayed-by-recovery topup decision against a stale "insufficient" read. Adding a tri-state would touch every consumer of the boolean for an edge case worth ~one cycle of conservative behavior. Happy to revisit if you want it now.

**5. Broad `except Exception → return None`.** Done. `_check_usdc_balance` now catches only `(requests.Timeout, requests.ConnectionError, ContractLogicError, Web3Exception, ValueError)`; programmer errors propagate. `CircuitBreakerOpenError` continues to re-raise.

**6. Cloudflare 5xx coverage.** Done. `_TRANSIENT_HTTP_STATUS_RE` is now `\b(408|429|500|502|503|504|520|521|522|523|524|525)\b`. Test asserts all 12 codes including `HTTP 522 from Cloudflare` classify as transient.

**7. Lock-release happy path untested.** Done. Three regression tests added (one per lock site): `test_write_withdrawal_data_releases_lock_after_exception`, `test_delayed_write_kv_extended_releases_lock_after_exception`, `test_ensure_sufficient_funds_releases_lock_after_exception`. Each forces the wrapped body to raise and asserts the lock is acquirable on the next call.

**8. Atomicity test verifies syscall, not invariant.** Moot — the persist path is gone.

PR #290, opened 2026-05-08
fix(fsm): address #288 audit findings

### LINKED ISSUE #288: FSM audit: dead transitions in PostTxSettlementRound and related cleanup
Audit produced by `/audit-fsm` against `packages/valory/skills/liquidity_trader_abci/` and `packages/valory/skills/optimus_abci/`. CLI tools (`autonomy analyse fsm-specs`, `handlers`, `dialogues`, `docstrings`) all pass for in-scope skills. No Critical or High findings.

### M2-1 — `PostTxSettlementRound.end_block()` never emits `DONE`/`NONE`/`NO_MAJORITY`

**File:** `packages/valory/skills/liquidity_trader_abci/states/post_tx_settlement.py:64-85`

`end_block()` returns one of 5 explicit events when `threshold_reached`, otherwise `None`. It never calls `super().end_block()` and never falls through to an `is_majority_possible()` check. So `Event.DONE`, `Event.NONE`, and `Event.NO_MAJORITY` declared in the round's transition table (`rounds.py:229-240`) are never produced at runtime.

**Impact:** liveness preserved only by `Event.ROUND_TIMEOUT` (30 s). On Byzantine disagreement the round waits the full 30 s rather than emitting `NO_MAJORITY` immediately. Not a hang, but a performance/observability gap; the dead self-loop transitions are misleading.

**Suggested fix:** add the standard fallback used by sibling rounds (e.g. `fetch_strategies.py:85-88`):

```python
if not self.is_majority_possible(self.collection, self.synchronized_data.nb_participants):
    return self.synchronized_data, Event.NO_MAJORITY
return None
```

Then drop the dead `Event.DONE` / `Event.NONE` self-loops from `PostTxSettlementRound`'s entry in `rounds.py`.

### M2-2 — Dead `Event.NONE` self-loops in rounds that override `super().end_block()`

**Files:**
- `packages/valory/skills/liquidity_trader_abci/states/fetch_strategies.py`
- `.../states/decision_making.py`
- `.../states/withdraw_funds.py`
- (and `.../states/post_tx_settlement.py`, see M2-1)

All four rounds override `end_block()` without calling `super()`, so `Event.NONE` is never returned. The corresponding `Event.NONE: <self>` entries in `rounds.py:229-255` are dead configuration. Other rounds (`APRPopulationRound`, `GetPositionsRound`, `CallCheckpointRound`, `CheckStakingKPIMetRound`, `EvaluateStrategyRound`) *do* delegate to `super()` and legitimately need their `NONE` entries.

**Impact:** no runtime bug, but the transitions create false confidence that empty-consensus is handled.

**Suggested fix:** either remove the `Event.NONE` entries from the four rounds' transition entries, or have those rounds delegate to `super().end_block()` for the non-threshold-reached path.

### L1 — `LiquidityTraderRoundBehaviour.initial_behaviour_cls` does not match `LiquidityTraderAbciApp.initial_round_cls`

- `packages/valory/skills/liquidity_trader_abci/behaviours/round_behaviour.py:61` — `initial_behaviour_cls = CallCheckpointBehaviour`
- `packages/valory/skills/liquidity_trader_abci/rounds.py:164` — `initial_round_cls: AppState = FetchStrategiesRound`
- `rounds.py:71-76` docstring claims `Initial round: FetchStrategiesRound`

`CallCheckpointRound` is in `initial_states`, so the metaclass accepts it. In the composed `OptimusConsensusBehaviour` (`optimus_abci/behaviours.py:51`) the entry point is overridden to `RegistrationStartupBehaviour`, so this is dead in production. Only standalone runs of `LiquidityTraderAbciApp` (e.g. unit-test fixtures) would observe the inconsistency.

**Suggested fix:** either set `initial_behaviour_cls = FetchStrategiesBehaviour` to match the AbciApp's declared entry point, or change `initial_round_cls` (and the docstring) to `CallCheckpointRound`. Pick whichever reflects intended behaviour for standalone runs.

### L2 — `SynchronizedData.actions` type annotation contradicts runtime use

- `packages/valory/skills/liquidity_trader_abci/states/base.py:177-184` — annotated `Optional[List[Dict[str, Any]]]`
- `packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py:99-109` — stores `{"event": "withdrawal_initiated", "updates": {}}` (a **dict**) into `actions`
- `.../states/evaluate_strategy.py:60-64` — `isinstance(actions_data, dict)` check relies on this dict-shaped value to detect withdrawal

The isinstance check works because the behaviour deliberately serialises a dict, not a list. But the public type annotation lies, and any future maintainer who trusts it (e.g. uses `actions[0]`) will introduce a real bug.

**Suggested fix:** either widen the annotation to `Optional[Union[List[Dict[str, Any]], Dict[str, Any]]]` (and document the convention), or migrate `EvaluateStrategy` to use `peek_withdrawal_event(self)` like the sibling rounds and add an `event` field to `EvaluateStrategyPayload`.

## Verified clean

- **C1** shared mutable refs (`([],) * n`)
- **C2** operator precedence in multi-line boolean conditions
- **C3** `end_block()` events all wired in transition tables
- **C4** `event_to_timeout` only contains `ROUND_TIMEOUT`, which is wired into every non-final round
- **H1** `BackgroundAppConfig` for termination correctly typed as `TERMINATING`
- **H2** `abci_app_transition_mapping` covers all six `Finished*`/`Failed*` final states of `LiquidityTraderAbciApp` plus the registration / tx-settlement / reset transitions
- **H3** all `requests.get` calls pass `timeout=...`; all `open()` use `with`; no `Popen`, no bare `thread.join()`
- **M1** apparent payload-class mismatches (single `content: str` field carrying a JSON blob) are by design — `most_voted_payload` returns the scalar field value, which is then `json.loads()`-ed
- **M3** DB pre/post conditions consistent with `chain()` constraints (`most_voted_tx_hash` set as post-condition for the three `Finished*` rounds that lead into `RandomnessTransactionSubmissionRound`)
- **M4** no thread `.join()` calls
- **M5** `collection_key`/`selection_key` align with `SynchronizedData` properties; tuple-form `selection_key`s in rounds whose `end_block()` overrides `super()` are intentionally dead (annotated as such in `withdraw_funds.py:46-47`)
- **T1** no `@classmethod @pytest.fixture` anti-pattern
- **T6** no `setup_class`/`teardown_class` overrides without `super()`

## Summary

| Severity | Count |
|----------|-------|
| Critical | 0     |
| High     | 0     |
| Medium   | 2     |
| Low      | 2     |

DIvyaNautiyal07:
## Summary

Addresses the four findings in #288.

- M2-1: `PostTxSettlementRound` now emits `Event.NO_MAJORITY` immediately when `is_majority_possible` returns false, matching `FetchStrategies` / `DecisionMaking` / `WithdrawFunds`. Drops the dead `Event.DONE` and `Event.NONE` alphabet entries.
- M2-2: `DecisionMakingRound`, `FetchStrategiesRound`, and `WithdrawFundsRound` override `end_block()` without delegating to `super()`, so they cannot emit `Event.NONE`. Drop the dead `Event.NONE` alphabet entries on those three rounds.
- L1: `LiquidityTraderRoundBehaviour.initial_behaviour_cls` now points at `FetchStrategiesBehaviour`, matching `LiquidityTraderAbciApp.initial_round_cls` and the AbciApp docstring. Production overrides this from `optimus_abci`, so this only affected standalone runs.
- L2: `SynchronizedData.actions` is now annotated `Optional[Union[List[Dict[str, Any]], Dict[str, Any]]]`, matching the runtime shape (the withdrawal-initiation flow writes a single dict).

For M2-1 and M2-2 the alphabet shrink required overriding `extended_requirements` and dropping the `none_event` / `done_event` class attributes on each affected round. The framework's spec checker reads `Event.X` literals in the round source as alphabet entries, so leaving those attributes in place would have left `tox -e check-abciapp-specs` failing. The runtime `self.none_event` / `self.done_event` references are only inside the parent's `end_block()`, which our overrides never call.

Also includes a one-line Makefile fix: `fix-abci-app-specs` referenced the old class name `SuperAgentAbciApp`; the actual class in `optimus_abci/composition.py` is `OptimusAbciApp`. The target failed on every clean checkout.

Closes #288

## Test plan

- [ ] \`tomte tox -e check-abciapp-specs\` passes
- [ ] \`tomte tox -e check-hash -e check-packages -e check-abci-docstrings -e check-handlers -e check-doc-hashes -e check-third-party-hashes -e check-dependencies\` all pass
- [ ] \`pytest packages/valory/skills/liquidity_trader_abci/tests\` passes (2840 tests, incl. new \`test_end_block_returns_no_majority_when_majority_impossible\` and \`test_end_block_returns_none_when_still_collecting\`)
- [ ] \`autonomy packages lock --check\` passes
- [ ] \`tomte tox -p -e black-check -e isort-check -e mypy -e pylint -e darglint\` pass (note: pre-existing flake8 issue at \`mirror_db/connection.py:397\` is unrelated to this branch)


DhairyaPatel7:
LGTM — all four findings are addressed correctly:

- **M2-1**: `PostTxSettlementRound`'s `is_majority_possible` short-circuit matches Fetch/Decision/Withdraw — verified those three already emit `NO_MAJORITY` in their `end_block` overrides, so the symmetry claim in the commit message holds.
- **M2-2**: The dead `Event.NONE` alphabet entries on the three sibling rounds are genuinely unreachable — each overrides `end_block` without `super()` and never returns `NONE`. Spec shrink + `extended_requirements` override is the correct way to satisfy `check-abciapp-specs` here.
- **L1**: `FetchStrategiesBehaviour` is in the `behaviours` set, so the standalone start now matches `LiquidityTraderAbciApp.initial_round_cls`.
- **L2**: type widening is consistent with the `isinstance(actions, dict)` dispatch path in `EvaluateStrategyRound`.

One follow-up worth tracking (pre-existing, not introduced here): `DecisionMakingRound.end_block` does `updated_actions.insert(index, new_action)` on `self.synchronized_data.actions` without an `isinstance(updated_actions, list)` guard. The L2 widening makes the type contract honest, but if the FSM ever re-enters DecisionMaking while `actions` still holds the withdrawal sentinel and `new_action` is set, the `.insert` blows up at runtime. Worth a guard or an issue.

Makefile fix is a free win — the target was unrunnable on a clean checkout against `OptimusAbciApp`.

DIvyaNautiyal07:
Good catch, added the guard in 2337f37d. `DecisionMakingRound.end_block` now starts a fresh list when `actions` is not a list, so the `.insert` can't crash when the sentinel dict (or None) is still in place. Added a test that crashes with `AttributeError: 'dict' object has no attribute 'insert'` if the guard is removed, so we'd catch a regression.

LOCKhart07 (approved):
LGTM for the four audit fixes. Verified each against the parent `CollectSameUntilThresholdRound.end_block` and the round transitions.

Extending @DhairyaPatel7's `decision_making.py:90` follow-up — the L2 type widening also reveals two consumer sites in the **behaviour** layer that subscript/iterate `self.synchronized_data.actions` without an isinstance guard:

- `behaviours/decision_making.py` ~L170: `actions = self.synchronized_data.actions; ... self._post_execute_step(actions, ...)` — passes a possibly-dict to a callee that indexes/subscripts.
- `behaviours/decision_making.py` ~L3290: `for action in self.synchronized_data.actions if action.get("action") == "EnterPool"` — iterating a dict yields string keys, so `str.get(...)` would AttributeError.

Reachability of the sentinel dict at these sites isn't obvious — `WithdrawFundsRound` overwrites `actions` with a list before flow re-enters DecisionMaking — but the contract is now public, so worth either a guard pass alongside Dhairya's site or a follow-up issue covering all three.

Makefile fix is a clean free win. (Test plan boxes still unchecked, fyi.)

DhairyaPatel7 (approved):

Issue #289, opened 2026-05-07
Resilience audit: blocking sleeps on FSM thread + non-fire-and-forget executor pattern


DavidMinarsch:
External-request resilience audit produced by `/audit-resilience` against `packages/valory/skills/{liquidity_trader_abci,optimus_abci}`, `packages/valory/connections/{mirror_db,x402}`, and the `customs/*_pools_search` libraries. No P0 crash bugs found — every HTTP call site has reasonable exception coverage and HTTP handler dispatch is globally wrapped (`optimus_abci/handlers.py:1313-1329`).

Two material stuck-agent findings, plus a handful of smaller hardening items.

## P1 — Customs `time.sleep` running on the FSM thread can exceed `ROUND_TIMEOUT`

**Files:**
- `packages/valory/customs/balancer_pools_search/balancer_pools_search.py:1291` — unconditional `time.sleep(5)` before pre-fetching all token prices in `get_filtered_pools_for_balancer`
- `packages/valory/customs/balancer_pools_search/balancer_pools_search.py:553, 976, 997, 1028, 1037` — cumulative `time.sleep(1-2)` per token in CoinGecko price loops
- `packages/valory/customs/velodrome_pools_search/velodrome_pools_search.py:610, 2128, 2206` — same pattern
- `packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py:2131` — calls `self.execute_strategy(**kwargs)` **synchronously** (not via `_async_execute_strategy`)

`execute_strategy` (defined at `behaviours/base.py:3584`) runs the custom in-band on the FSM thread. The async path via `loop.run_in_executor` (`evaluate_strategy.py:1527-1538`) avoids this, but the sync call at `2131` does not.

**Worst-case stall:** for a portfolio with N positions evaluated sequentially, `N × (5 + 1·M)` seconds where M is tokens per pool. The round timeout is **30 s** (`liquidity_trader_abci/rounds.py:271-273`), so even one Balancer position with several tokens can blow the budget.

**FSM impact:**
1. Behaviour blocks past `ROUND_TIMEOUT`.
2. Tendermint fires the timeout; round transitions to its self-loop.
3. Behaviour eventually returns; the next tick fires against a different round, work is silently dropped.
4. Net: stuck in `EvaluateStrategyRound` self-loop, no progress, error logs don't distinguish "API slow" from "buggy logic".

[...]

PR #291, opened 2026-05-08
fix(resilience): address #289 round-2 audit findings

DIvyaNautiyal07:
Addresses [#289](https://github.com/valory-xyz/optimus/issues/289). Six findings implemented in one PR; one finding (customs Web3 timeout) was a false positive — the cited lines already pass `request_kwargs={"timeout": WEB3_HTTP_TIMEOUT_SECONDS}` (fix landed earlier in commit `3f41150f`).

Scope doc with full file:line context: [docs/issue-289-resilience-round2-scoping.md](docs/issue-289-resilience-round2-scoping.md).

## Summary

**P1 — Stuck-agent: customs blocking sleeps on FSM thread.** `get_returns_metrics_for_opportunity` called `self.execute_strategy` synchronously, running customs in-band on the FSM tick. Customs do `time.sleep(1-5)` for naive rate limiting, so portfolios with several positions could blow the 30s round timeout. Routed through `_async_execute_strategy` + `asyncio.ensure_future` (matches the parallel path at `evaluate_strategy.py:1661`). Also dropped the unconditional `time.sleep(5)` preflight at `balancer_pools_search.py:1291` — it was redundant with the per-call `time.sleep(3)` / `time.sleep(2)` inside `get_pool_token_prices` (lines 976, 997), which already gate the CoinGecko request rate.

**P2 — Stuck-handler: `with ThreadPoolExecutor` is not fire-and-forget.** Four sites used `with ThreadPoolExecutor(max_workers=1) as executor: executor.submit(fn); executor.shutdown(wait=False)`. `__exit__` calls `shutdown(wait=True)` and overrides the explicit non-blocking shutdown, so the handler thread blocked up to ~30s on each call. Replaced with a long-lived `_background_executor` and a `_submit_background` helper that attaches a done-callback for exception logging.

**P3 — mirror_db retry budget exceeds round timeout.** `max_retries=5` × 30s session timeout + backoff = ~165s worst case under sustained outage; the FSM round (30s) timed out long before the connection finished retrying. Reduced to `max_retries=3` and a 5s session timeout. Worst case: 18s, fits inside the round budget. Pinned the invariant with `test_default_retry_budget_fits_under_round_timeout` so future tuning trips the test.

**P3 — `/healthcheck` returns 500 during FSM startup.** Line 1535 accessed `round_sequence._abci_app.current_round.round_id` behind a truthy guard on `_abci_app`, but `current_round` can be unbound during the startup window. Mirrored the sister-pattern guard already used 12 lines below for `synchronized_data.period_count`.

**P4 — No circuit breaker for off-chain APIs.** Wired `EndpointCircuitBreaker` into the LiFi quote path (`_get_lifi_quote_sync`). Reused `SharedState.get_circuit_breaker` keyed on `"lifi-quote"` (distinct from per-chain RPC keys). CoinGecko / Safe / Merkl / subgraphs are deferred — each has its own retry layer and merits separate scope.

**P4 — Broad `except Exception → return None` in `_get_lifi_quote_sync`.** Round 1 narrowed the catch-all in `_check_usdc_balance`; the sister method 12 lines below was missed in that ripple. Narrowed to `(requests.Timeout, requests.ConnectionError, ValueError, KeyError)` so programmer errors propagate to the global dispatch wrapper.

## Commits

One commit per finding plus one chore commit for hash regeneration and one style commit for black:

- `1a1077cd` fix(p1): route per-position metric path off the FSM thread
- `0ff2fa63` fix(p1): drop unconditional 5s preflight sleep in balancer customs
- `486aa588` fix(p2): use a long-lived background executor instead of with-block
- `51bef4f1` fix(p3): cap mirror_db retry budget under the 30s round timeout
- `4d4b6a1e` fix(p3): guard /healthcheck current_round access during startup
- `75779792` fix(p4): wire LiFi quote calls through a circuit breaker
- `11f68486` fix(p4): narrow _get_lifi_quote_sync except to typed transient set
- `eac7ca40` chore(generators): re-lock package hashes after #289 fixes
- `1b8be9ec` style: black reformat list comprehension and call wrapping

## Test plan

- [x] All touched skill / connection / customs test suites pass locally
- [x] `make code-checks` clean (black, isort, flake8, mypy, pylint, darglint)
- [x] `autonomy packages lock --check` verifies hashes
- [x] `autonomy push-all` ran before this branch was pushed
- [ ] CI green on this branch
- [ ] Manual smoke: agent startup probe does not flap (P3 healthcheck guard)
- [ ] Manual smoke: `/funds-status` returns within ~100ms even when LiFi/Web3 are slow (P2 fire-and-forget)

## Out of scope (deferred)

- Circuit breaker on CoinGecko / Safe / Merkl / subgraphs.
- Wiring `update_position_metrics` into `EvaluateStrategyBehaviour.async_act`. The method is currently defined but uncalled in production; the P1 fix here makes it safe to wire up later.

## Counter-arguments worth raising before approval

1. **P3 mirror_db: 5s session timeout could trip on a slow but eventually-successful upstream.** If we observe legitimate slow calls failing, raise to 8s and drop `max_retries` to 2 (worst case 17s). Watch logs after deploy.
2. **P1: removing the 5s preflight sleep could expose CoinGecko 429s.** The inline `time.sleep(3)` / `time.sleep(2)` in `get_pool_token_prices` still gate the per-call rate, so the preflight was duplicate. If 429s spike post-deploy we can reinstate with a pointer comment.
3. **P4 circuit breaker only covers LiFi.** This closes the highest-impact gap; the rest is a follow-up issue.


packages/valory/customs/balancer_pools_search/balancer_pools_search.py
```diff
@@ -1288,7 +1288,6 @@ def format_pool_data(
         for token in pool["poolTokens"]:
             all_token_symbols.append(token["symbol"])
 
-    time.sleep(5)  # Rate limiting
     all_prices = get_pool_token_prices(
```
LOCKhart07, line 1291: Commit `0ff2fa63` and the PR body justify removing the 5s preflight by saying `CoingeckoRateLimiter already throttles the same calls`. That's not what's actually happening here — the customs don't import `CoingeckoRateLimiter` (it lives in `liquidity_trader_abci/models.py` and isn't passed into customs). The reason the preflight is redundant is the inline `time.sleep(3)` at `get_pool_token_prices:976` plus the `time.sleep(2)` at line 997 inside the same function — those still gate the rate. Net effect of the change is fine; please correct the rationale in the commit message / PR description so the next reader doesn't go looking for a rate limiter that isn't wired in.

packages/valory/skills/optimus_abci/handlers.py
```diff
@@ -331,13 +331,20 @@ class HttpHandler(BaseHttpHandler):
 
     def setup(self) -> None:
         """Implement the setup."""
+        # Long-lived executor for fire-and-forget background work. The
+        # `with ThreadPoolExecutor(...) as executor:` idiom is NOT
+        # fire-and-forget — __exit__ calls shutdown(wait=True) and the
+        # caller blocks until the submission completes.
+        if not hasattr(self, "_background_executor"):
+            self._background_executor = ThreadPoolExecutor(
```
LOCKhart07, line 335: The new long-lived `_background_executor` is never shut down. `BaseHandler.teardown` (line 248) only logs; nothing calls `self._background_executor.shutdown(...)` on agent teardown, and the new `test_submit_background_*` tests trigger lazy-creation via `_submit_background` itself (the executor isn't patched out the way `test_setup_http_handler_x402` patches `ThreadPoolExecutor`). Each such test leaks 2 worker threads until process exit — bounded for production, but the test count will grow. Please add `if hasattr(self, '_background_executor'): self._background_executor.shutdown(wait=False)` to a `teardown` override here, and ensure the new tests either patch the executor or shut it down in a finally block.

packages/valory/skills/optimus_abci/handlers.py
```diff
@@ -514,6 +547,29 @@ def _call_web3_with_breaker(
...
+        if not breaker.allow():
+            raise CircuitBreakerOpenError(
+                "Circuit breaker open for lifi-quote; skipping call"
+            )
+        try:
+            result = func()
+        except Exception:
+            breaker.on_failure()
+            raise
+        breaker.on_success()
```
LOCKhart07, line 566: Sustained LiFi 5xx responses won't open this breaker. `requests.get` returns a `Response` for any status code (only transport errors raise), so the lambda returns normally, `on_success()` runs, and `_get_lifi_quote_sync` silently returns None on the `if response.status_code == 200` check. Only `request_timeout=30s` would eventually trip the breaker via `Timeout`. That's likely acceptable given LiFi 5xx is rare and the timeout backstops it, but worth a follow-up to consider treating non-200 (or just 5xx/429) as breaker failures — otherwise the breaker is structurally weaker than `_call_web3_with_breaker`, where the wrapped Web3 call raises on most RPC errors. Non-blocking.

packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
```diff
@@ -2127,20 +2127,27 @@ def get_returns_metrics_for_opportunity(
             }
         )
 
-        # Execute the strategy to calculate metrics
-        metrics = self.execute_strategy(**kwargs)
+        # Run the strategy off the FSM thread; the customs perform
+        # rate-limit sleeps that would otherwise block past ROUND_TIMEOUT.
+        strategies_executables = self.shared_state.strategies_executables
+        future = asyncio.ensure_future(
+            self._async_execute_strategy(strategy, strategies_executables, **kwargs)
```
LOCKhart07, line 2133: Compare to `_run_all_strategies` at line 1564, which defensively strips `strategy` before forwarding: `kwargs_without_strategy = {k: v for k, v in kwargs.items() if k != 'strategy'}`. The new path here is currently safe by construction (the diff removed the `"strategy": strategy` entry from the `kwargs.update(...)` block above), but a future `strategies_kwargs` config that happens to include a `strategy` key would `TypeError` on the duplicate positional arg. Mirroring the strip — or asserting the kwargs don't contain `strategy` — would harden it. Optional.

packages/valory/customs/balancer_pools_search/balancer_pools_search.py
DIvyaNautiyal07: Good catch on the rationale, the customs do not import CoingeckoRateLimiter. The actual throttle is the inline time.sleep(3)/time.sleep(2) inside get_pool_token_prices. Net behaviour of the change is the same; the commit message overstated the protection. I will note the correct rationale in the PR description so the next reader does not go looking for a wired-in limiter.

packages/valory/skills/optimus_abci/handlers.py
DIvyaNautiyal07: Done in c99f1e74. HttpHandler.teardown() now shuts down the long-lived executor and clears the attribute so a future setup() lazily recreates a fresh one. The two existing _submit_background tests now release the executor in a finally block, and a new test_teardown_shuts_down_background_executor pins the lifecycle (asserts re-submission against the dead executor raises and that the handler attribute is cleared).

packages/valory/skills/optimus_abci/handlers.py
DIvyaNautiyal07: Fair point, the breaker was structurally weaker than _call_web3_with_breaker. Tightened in c99f1e74: the lambda inside _get_lifi_quote_sync now raises requests.HTTPError for 429 and 5xx before the breaker observes the result, so sustained LiFi outage actually opens the breaker. test_get_lifi_quote_breaker_opens_on_sustained_5xx covers the new path with a 503 response.

packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
DIvyaNautiyal07: Done in c99f1e74. get_returns_metrics_for_opportunity now does kwargs.pop('strategy', None) before the asyncio.ensure_future call, mirroring _run_all_strategies.

LOCKhart07 (dismissed):
All four review comments addressed in `c99f1e74` — verified end-to-end:

- `teardown()` override at `handlers.py:420` calls `executor.shutdown(wait=False)` and clears the attribute; new `test_teardown_shuts_down_background_executor` pins the lifecycle, and the two existing `_submit_background` tests now wrap in `try/finally`.
- 5xx and 429 promoted to `requests.exceptions.HTTPError` inside the LiFi lambda (`handlers.py:657-666`) so sustained outage actually opens the breaker; `test_get_lifi_quote_breaker_opens_on_sustained_5xx` covers the 503 path. `requests.exceptions.HTTPError` added to the typed except tuple.
- `kwargs.pop("strategy", None)` at `evaluate_strategy.py:2131` mirrors `_run_all_strategies`.

One open item from the comment on `balancer_pools_search.py:1291`: the PR description still says *“since `CoingeckoRateLimiter` already throttles the same calls”* even though you noted in the inline reply that the actual throttle is the inline `time.sleep(3)` / `time.sleep(2)` in `get_pool_token_prices`. Please edit the body so the next reader doesn’t go looking for a wired-in limiter.

Approving — non-blocking on the description fix.

DhairyaPatel7 (approved):

LOCKhart07 (approved):

PR #292, opened 2026-05-11
fix(withdrawal): refresh portfolio before routing to WithdrawFunds

DIvyaNautiyal07:
## What this fixes

A user reported that after a successful withdrawal, the UI continued to show the pre-withdrawal portfolio, leading them to believe the withdrawal had not happened. They retried — and the second withdrawal failed with `Withdrawal failed during route fetching`, even though the first transaction had landed on-chain.

Root cause traced in agent log `sc-59606399-a285-4ddb-b8b5-8eca7a53ff25_agent.log`:

1. Cycle 1 completed at 16:30:34 with tx `0x5be55fac...` — USDC + USDT were drained correctly.
2. UI kept serving stale `/portfolio` data (`last_updated` timestamp from before the agent session even started). It showed the Velodrome pool still at $4.95 with USDT + USDC in the safe.
3. User clicked retry at 16:32:29, which set `investing_paused=true`.
4. Next `fetch_strategies_round` at 16:35:34 hit the `investing_paused` short-circuit added at the top of `async_act` by #279 and exited before reaching `store_portfolio_data()`. Portfolio stayed stale.
5. Cycle 2 used the stale portfolio_breakdown (showed USDT $4.61 > $1 threshold) and planned a `FindBridgeRoute USDT → USDC`. The decision_making error at 16:35:39 (`Could not determine available amount for USDT`) was the correct behaviour — there genuinely was no USDT to swap. Refreshing the portfolio means cycle 2 starts with the real post-withdrawal breakdown, so the ghost swap action is never built in the first place and the error path never fires.

The regression came from #279, which added the `investing_paused` gate at the top of `FetchStrategiesBehaviour.async_act` to fix `WITHDRAWAL_INITIATED` routing from in-cycle rounds. Before #279, the behaviour always ran through to portfolio refresh first; after #279, the gate short-circuited that path.

## The change

Move the `investing_paused` gate from line 109-128 (top of `async_act`) to line 263 (right after `store_portfolio_data()`). The routing PR #279 added still works because the round's `end_block` at [fetch_strategies.py:58-67](packages/valory/skills/liquidity_trader_abci/states/fetch_strategies.py#L58-L67) emits `WITHDRAWAL_INITIATED` based on the payload's `event` field, and the payload is still emitted — just after the refresh.

The other four behaviours that #279 also gated (`APRPopulation`, `CallCheckpoint`, `CheckStakingKPIMet`, `GetPositions`) are not touched. They don't refresh portfolio, so their top-of-function gate is correct.

One caveat the move doesn't change: on `period_count == 0` boot with a pending ETH revert, `_check_and_create_eth_revert_transactions` at [fetch_strategies.py:146](packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py#L146) can emit a SETTLE payload and `set_done()` at lines ~418-442, returning before the gate is reached. Withdrawal routing is delayed by one cycle in that boot-time corner case. This is pre-existing behaviour (the eth-revert short-circuit predates #279) and is preserved deliberately — aborting the revert mid-flight would lose the tx-routing semantics.

The pre-gate refresh body is wrapped in `try/except` so RPC, KV, or file-I/O failures (e.g. degraded RPC during a withdrawal) don't silently drop the WITHDRAWAL_INITIATED payload. On failure the refresh is skipped and the gate still fires.

## Tests

- `test_portfolio_refreshed_before_withdrawal_routing` — pins the call order: `store_portfolio_data` must fire before `send_a2a_transaction` when `investing_paused=true`. Regression for the top-of-function-gate behaviour.
- `test_withdrawal_payload_still_emits_when_refresh_raises` — stubs `_get_native_balance` to raise; asserts the withdrawal payload still ships.
- `test_period_zero_routes_withdrawal_after_eth_revert_noop` — covers `period_count=0` with the eth-revert path stubbed; pins the same ordering for the boot path.
- `test_no_withdrawal_payload_when_not_paused` — verifies normal flow emits the regular selected-protocols payload and still refreshes portfolio.

Full liquidity_trader test suite: 2843 passed.

## Test plan

- [x] All touched skill tests pass locally
- [x] \`autonomy packages lock --check\` verifies hashes
- [x] \`autonomy push-all\` ran before this branch was pushed
- [ ] CI green
- [ ] Manual smoke: complete a withdrawal, then check `/portfolio` is updated within one fetch_strategies cycle even if a new withdrawal is initiated during the reset_and_pause window

## Counter-arguments

1. **`_update_withdrawal_completion` should refresh portfolio directly at withdrawal completion.** That would close a separate window (the ~5 minutes of `reset_and_pause` where /portfolio still shows pre-withdrawal data even without a retry). Worth doing in a follow-up — the current fix covers the next-fetch-strategies refresh, which is the path the regression broke.
2. **Could re-introduce the same bug for a future engineer adding logic after `store_portfolio_data` but before the gate.** The new regression test pins the ordering so this would trip the test.


LOCKhart07 (dismissed):
LGTM — the move is semantically clean: `end_block` at `states/fetch_strategies.py:58-67` keys off the payload's `event` field regardless of position in `async_act`, and the other four behaviours #279 gated don't touch portfolio data so leaving them alone is correct.

Two non-blocking observations:

1. **Period-0 boot + pending ETH revert still short-circuits before the new gate.** `_check_and_create_eth_revert_transactions` (called at `behaviours/fetch_strategies.py:146` when `period_count == 0`) can emit a `SETTLE` payload + `set_done()` + `return` at lines ~418-442, so withdrawal routing is delayed by one cycle in that boot-time corner case. The regression analysis in the PR body reads as if the gate is always reached after refresh — worth a sentence acknowledging this path so future readers aren't surprised.

2. **Tests don't exercise the `period_count == 0` branch.** Both new tests set `period_count=1`, which skips the eth-revert path entirely. A test with `period_count=0` and a no-op `_check_and_create_eth_revert_transactions` would pin behaviour for the boot scenario.

Out of scope but worth a follow-up issue: the condition at `behaviours/fetch_strategies.py:225` (`if not (time_since_update >= one_day_in_seconds):`) is inverted — `_track_whitelisted_assets()` (with `sleep(5)` per asset) runs when *less* than a day has passed and then updates the timestamp, so it runs every cycle. With this PR, that latency now sits on the withdrawal-routing critical path each cycle while `investing_paused=true`.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -260,6 +239,27 @@ def async_act(self) -> Generator:
             # Update agent performance metrics
             self._update_agent_performance_metrics()
 
+            investing_paused = yield from self._read_investing_paused()
```
DhairyaPatel7, line 255: The new placement introduces a failure mode the old code didn't have: any exception thrown before reaching this line — from `_get_native_balance` (L142, RPC), `_check_and_create_eth_revert_transactions` (L147, RPC + tx submission on period 0), `_read_kv` (L150), `_track_whitelisted_assets` (L227, external price API), `calculate_user_share_values` (L234, on-chain reads), or `store_portfolio_data` (L237, file I/O) — crashes the behaviour before `investing_paused` is read. The withdrawal payload never gets emitted, so the agent silently fails to route the withdrawal under degraded RPC. Pre-PR, only `_read_investing_paused` (a fast KV read) ran before the gate, so routing was robust to upstream failures. The new regression test only checks happy-path ordering, so this is also untested.

Suggested mitigations:

1. **Cheap check at the top, full check after refresh.** Read `investing_paused` at the top of `async_act` and short-circuit to the withdrawal payload before any RPC work. Move the refresh logic into the `not investing_paused` branch. Restores the old robustness with minimal code change.
2. **Try/except around the refresh path** so the gate still fires on failure.
3. At minimum, add a regression test stubbing one of the fallible calls to raise and asserting the WITHDRAWAL_INITIATED payload still ships — if it doesn't, document the trade-off in the PR body so reviewers see it.

DIvyaNautiyal07: Done in d3fbf485. Pre-gate refresh body is wrapped in try/except now, so a raised exception is logged and the gate still fires with stale portfolio rather than the FSM stalling. test_withdrawal_payload_still_emits_when_refresh_raises stubs _get_native_balance to raise and asserts the WITHDRAWAL_INITIATED payload still ships.

DIvyaNautiyal07:
Thanks. Addressed in d3fbf485.

Obs #1 (period-0 + eth-revert): added a sentence to the PR body acknowledging the pre-existing short-circuit so the regression analysis isn't misleading.

Obs #2 (period-0 test): new test_period_zero_routes_withdrawal_after_eth_revert_noop pins the boot scenario with the eth-revert path stubbed to a no-op.

Obs #3 (inverted whitelist condition): will fix in a follow-up PR — it's a separate scope and the fix needs its own reproduction.

DhairyaPatel7 (approved):

LOCKhart07 (approved):

PR #320, opened 2026-05-18
fix: on-chain fallback for whitelisted balances in get_positions

DIvyaNautiyal07:
## Summary

When SafeApi's Optimism balances endpoint is unavailable (e.g. monthly quota exceeded, transient indexing lag, or trusted-token list misses), the resulting `positions` array doesn't contain the safe's real holdings. Downstream consumers that read from `positions` then misbehave:

- **Investing**: `evaluate_strategy._get_available_tokens` (evaluate_strategy.py:2769) iterates `synchronized_data.positions`, sees only reward tokens + dust, logs `No tokens available for investment`, increments the strategy backoff counter, and the agent stops trying for 30 minutes to 4 hours per cycle.
- **Withdrawal**: `decision_making.fetch_routes` (decision_making.py:2580) calls `_get_balance(chain, from_token, positions)`, gets `None`, errors with `Could not determine available amount`, and marks the withdrawal `FAILED`.

This was the root cause behind several user support tickets this week (sc-8ce1d52f, sc-2655c77a, sc-8c5fdde5, plus an investing-stuck case). PR #281 added an on-chain backstop in the withdrawal swap *planner*, but not at the data source.

## What changed

Adds `_supplement_with_onchain_whitelisted_balances` in `base.py`. Called from `_get_optimism_balances_from_safe_api` right after the SafeApi parse and before `_fetch_reward_balances`. Mutates the `balances` list in place:

1. Reads native ETH via `_get_native_balance` and appends if missing.
2. Iterates `WHITELISTED_ASSETS[chain]` (USDC, USDT0, USDT, USDC.e, sDAI, oUSDT on Optimism), calls `_get_token_balance` on-chain for each, appends any non-zero balance not already in the result list.
3. Dedupes by lowercased address so SafeApi-returned values are preserved when both sources have an entry.

Reward tokens (OLAS, VELO) are unaffected — already covered by `_fetch_reward_balances`, which uses on-chain reads and runs unconditionally. The dedicated `_fetch_ousdt_balance` path is now also deduped against existing entries.

Result: `positions` reflects on-chain reality whenever SafeApi is unavailable, unblocking both the investment evaluator and the withdrawal executor without needing per-call-site patches.

## Tests

- 6 new tests in `TestSupplementWithOnchainWhitelistedBalances` cover: native ETH added when missing, ETH preserved when already present, whitelisted ERC20 added when missing, ERC20 not duplicated when already present (case-insensitive), zero balance skipped, RPC `None` skipped.
- 10 existing `TestGetOptimismBalances` tests updated to stub the new helper to a no-op so they keep testing their original intent.
- Full skill suite: 2850 tests pass.
- Production-replica script driven against the exact sc-8ce1d52f shape (SafeApi `(False, [])`, real on-chain balances mocked at the RPC boundary, real `_get_balance` consumer). Confirmed `positions` returns USDC + USDT + USDC.e + ETH instead of empty.

## Scope notes

- **Optimism only.** Mode's `_get_mode_balances_from_explorer_api` has its own native-ETH backstop but no whitelist backstop. Same helper would apply if Mode Explorer also goes flaky — easy follow-up, not in this PR.
- **Cost**: 6 extra RPC calls per `get_positions_round` for the whitelisted-token reads. On the happy path (SafeApi works), the dedup makes these confirming-zero checks, no behavioural change. Withdrawals happen infrequently and the investment round runs every few minutes, so the added cost is negligible.

## Test plan

- [ ] CI green (lint, mypy, pytest, coverage, package-hash check)
- [ ] Verify in a deployed agent: when SafeApi is unavailable, `get_positions_round` log shows the safe's on-chain stablecoin and ETH balances rather than empty positions.

### DIFF
--- a/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
+++ b/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
@@ -526,8 +526,10 @@ class LiquidityTraderBaseBehaviour(
 
         # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
         # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
-        # returns an incomplete trusted-token list. Reward tokens are handled
-        # by _fetch_reward_balances below, so no overlap here.
+        # returns an incomplete trusted-token list. Covers every entry in
+        # WHITELISTED_ASSETS[chain] including oUSDT, which used to be fetched
+        # separately. Reward tokens (OLAS, VELO) are disjoint from the
+        # whitelist and handled by _fetch_reward_balances below.
         yield from self._supplement_with_onchain_whitelisted_balances(
             "optimism", safe_address, balances
         )
@@ -536,13 +538,6 @@ class LiquidityTraderBaseBehaviour(
         reward_balances = yield from self._fetch_reward_balances("optimism")
         if reward_balances:
             balances.extend(reward_balances)
-
-        # Add separate OUSDT balance fetch for Optimism (not included in trusted API)
-        ousdt_balance = yield from self._fetch_ousdt_balance("optimism")
-        if ousdt_balance:
-            already_have = {b.get("address", "").lower() for b in balances}
-            if ousdt_balance.get("address", "").lower() not in already_have:
-                balances.append(ousdt_balance)
 
         self.context.logger.info(
             f"Retrieved {len(balances)} token balances from SafeApi"
@@ -3495,8 +3490,19 @@ class LiquidityTraderBaseBehaviour(
     ) -> Generator[None, None, None]:
         """Add on-chain balances for whitelisted assets not present in ``balances``.
 
-        Mutates ``balances`` in place. Skips any token whose address is already
-        in ``balances`` (case-insensitive). Also adds native ETH if missing.
+        Mutates ``balances`` in place. For each entry in ``WHITELISTED_ASSETS[chain]``
+        plus native ETH, appends an entry sourced from on-chain reads when:
+        - the address is not already in ``balances`` (case-insensitive match), and
+        - the on-chain balance is strictly positive.
+
+        Zero-balance tokens and tokens whose RPC read returns ``None`` are
+        skipped. The latter is logged as a warning so missing-data incidents
+        remain diagnosable.
+
+        :param chain: the chain to read on-chain balances from.
+        :param safe_address: the Safe contract address to read balances for.
+        :param balances: existing balances list to supplement in place.
+        :yield: None, for async-style generator progression.
         """
         already_have = {
             (b.get("address") or "").lower() for b in balances if b.get("address")
@@ -3523,7 +3529,13 @@ class LiquidityTraderBaseBehaviour(
             token_balance = yield from self._get_token_balance(
                 chain, safe_address, whitelisted_addr
             )
-            if not token_balance or token_balance <= 0:
+            if token_balance is None:
+                self.context.logger.warning(
+                    f"On-chain balance read returned None for {symbol or whitelisted_addr} "
+                    f"on {chain}; skipping. Positions data may be incomplete."
+                )
+                continue
+            if token_balance <= 0:
                 continue
             balances.append(
                 {
@@ -3591,50 +3603,6 @@ class LiquidityTraderBaseBehaviour(
         except Exception as e:
             self.context.logger.error(
                 f"Error getting last known price for token {token_address}: {str(e)}"
-            )
-            return None
-
-    def _fetch_ousdt_balance(
-        self, chain: str
-    ) -> Generator[None, None, Optional[Dict[str, Any]]]:
-        """Fetch OUSDT token balance separately using direct contract call."""
-        try:
-            safe_address = self.params.safe_contract_addresses.get(chain)
-            if not safe_address:
-                self.context.logger.error(f"No safe address found for chain {chain}")
-                return None
-
-            # OUSDT token address (same on both chains)
-            ousdt_token_address = (
-                "0x1217BfE6c773EEC6cc4A38b5Dc45B92292B6E189"  # nosec B105
-            )
-
-            self.context.logger.info(
-                f"Fetching OUSDT balance separately for {chain} at address {ousdt_token_address}"
-            )
-
-            # Get OUSDT balance using direct contract call
-            ousdt_balance = yield from self._get_token_balance(
-                chain, safe_address, ousdt_token_address
-            )
-
-            if ousdt_balance and ousdt_balance > 0:
-                self.context.logger.info(
-                    f"Found OUSDT balance: {ousdt_balance} for {chain}"
-                )
-                return {
-                    "asset_symbol": "oUSDT",
-                    "asset_type": "erc_20",
-                    "address": to_checksum_address(ousdt_token_address),
-                    "balance": ousdt_balance,
-                }
-            else:
-                self.context.logger.info(f"No OUSDT balance found for {chain}")
-                return None
-
-        except Exception as e:
-            self.context.logger.error(
-                f"Error fetching OUSDT balance separately for {chain}: {str(e)}"
             )
             return None
 
--- a/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
+++ b/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
@@ -524,6 +524,14 @@ class LiquidityTraderBaseBehaviour(
                         }
                     )
 
+        # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
+        # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
+        # returns an incomplete trusted-token list. Reward tokens are handled
+        # by _fetch_reward_balances below, so no overlap here.
+        yield from self._supplement_with_onchain_whitelisted_balances(
+            "optimism", safe_address, balances
+        )
+
         # Add separate reward token balance fetch for Optimism
         reward_balances = yield from self._fetch_reward_balances("optimism")
         if reward_balances:
@@ -532,7 +540,9 @@ class LiquidityTraderBaseBehaviour(
         # Add separate OUSDT balance fetch for Optimism (not included in trusted API)
         ousdt_balance = yield from self._fetch_ousdt_balance("optimism")
         if ousdt_balance:
-            balances.append(ousdt_balance)
+            already_have = {b.get("address", "").lower() for b in balances}
+            if ousdt_balance.get("address", "").lower() not in already_have:
+                balances.append(ousdt_balance)
 
         self.context.logger.info(
             f"Retrieved {len(balances)} token balances from SafeApi"
@@ -3476,6 +3486,54 @@ class LiquidityTraderBaseBehaviour(
                 f"Error fetching reward balances for {chain}: {str(e)}"
             )
             return reward_balances
+
+    def _supplement_with_onchain_whitelisted_balances(
+        self,
+        chain: str,
+        safe_address: str,
+        balances: List[Dict[str, Any]],
+    ) -> Generator[None, None, None]:
+        """Add on-chain balances for whitelisted assets not present in ``balances``.
+
+        Mutates ``balances`` in place. Skips any token whose address is already
+        in ``balances`` (case-insensitive). Also adds native ETH if missing.
+        """
+        already_have = {
+            (b.get("address") or "").lower() for b in balances if b.get("address")
+        }
+
+        zero_addr_lc = ZERO_ADDRESS.lower()
+        if zero_addr_lc not in already_have:
+            native_balance = yield from self._get_native_balance(chain, safe_address)
+            if native_balance and native_balance > 0:
+                balances.append(
+                    {
+                        "asset_symbol": "ETH",
+                        "asset_type": "native",
+                        "address": to_checksum_address(ZERO_ADDRESS),
+                        "balance": int(native_balance),
+                    }
+                )
+                already_have.add(zero_addr_lc)
+
+        for whitelisted_addr, symbol in WHITELISTED_ASSETS.get(chain, {}).items():
+            key = whitelisted_addr.lower()
+            if key in already_have:
+                continue
+            token_balance = yield from self._get_token_balance(
+                chain, safe_address, whitelisted_addr
+            )
+            if not token_balance or token_balance <= 0:
+                continue
+            balances.append(
+                {
+                    "asset_symbol": symbol or "UNKNOWN",
+                    "asset_type": "erc_20",
+                    "address": to_checksum_address(whitelisted_addr),
+                    "balance": int(token_balance),
+                }
+            )
+            already_have.add(key)
 
     def _get_last_known_price(
         self, token_address: str
--- a/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
+++ b/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
@@ -3510,42 +3510,61 @@ class LiquidityTraderBaseBehaviour(
 
         zero_addr_lc = ZERO_ADDRESS.lower()
         if zero_addr_lc not in already_have:
-            native_balance = yield from self._get_native_balance(chain, safe_address)
-            if native_balance and native_balance > 0:
-                balances.append(
-                    {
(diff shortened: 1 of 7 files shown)


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -3477,6 +3482,71 @@ def _fetch_reward_balances(
...
+                    {
+                        "asset_symbol": "ETH",
+                        "asset_type": "native",
+                        "address": to_checksum_address(ZERO_ADDRESS),
+                        "balance": int(native_balance),
+                    }
+                )
+                already_have.add(zero_addr_lc)
+
+        for whitelisted_addr, symbol in WHITELISTED_ASSETS.get(chain, {}).items():
```
DhairyaPatel7, line 3537: Missing exception safety: the per-token RPC calls (`_get_native_balance` at line 3513, `_get_token_balance` at line 3529) are not wrapped in try/except. `contract_interact` returns `None` on a clean failure (handled), but the underlying `get_contract_api_response` / ledger dispatch can still raise on timeouts, connection errors, or framework-level issues — which is exactly why the sibling `_fetch_reward_balances` (line 3420) wraps each iteration in try/except and the removed `_fetch_ousdt_balance` had its own try/except. As written, one RPC raise on any whitelisted token aborts `_supplement_with_onchain_whitelisted_balances`, which then propagates out of `_get_optimism_balances_from_safe_api` (no surrounding try/except at line 533) and out of `get_positions`. That regresses the resiliency posture: previously an oUSDT RPC error logged-and-continued; now any whitelisted RPC error kills the entire positions round — the very failure mode this PR is supposed to mitigate. Fix: wrap the for-loop body (and the native fetch) in a `try/except Exception as e: logger.warning(...); continue` matching the pattern in `_fetch_reward_balances`.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -3477,6 +3482,71 @@ def _fetch_reward_balances(
...
+        :param balances: existing balances list to supplement in place.
+        :yield: None, for async-style generator progression.
+        """
+        already_have = {
+            (b.get("address") or "").lower() for b in balances if b.get("address")
+        }
+
+        zero_addr_lc = ZERO_ADDRESS.lower()
+        if zero_addr_lc not in already_have:
+            native_balance = yield from self._get_native_balance(chain, safe_address)
```
DhairyaPatel7: Inconsistent None-handling for diagnostics: when `_get_token_balance` returns None the helper logs a warning (line 3533), but when `_get_native_balance` returns None it is silently swallowed by `if native_balance and native_balance > 0`. The native helper does log an error internally on a STATE-performative mismatch, but that path only covers the LedgerApi failure case — any silent None from the framework or upstream changes would be invisible here. Mirror the token-side warning for symmetry, or at least drop the redundant `> 0` so the intent (positive balance) is the only check.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_base.py
```diff
@@ -1138,14 +1135,165 @@ def _failing_read(keys):  # type: ignore
...
+        )
+        eths = [b for b in balances if b["asset_symbol"] == "ETH"]
+        assert len(eths) == 1
+        assert eths[0]["balance"] == 100  # preserves SafeApi value
+
+    def test_adds_whitelisted_erc20_when_missing(self) -> None:
+        """USDC absent from balances and non-zero on-chain is appended."""
+        b = _make_behaviour()
+        b._get_native_balance = _make_gen(0)  # type: ignore[assignment,method-assign]
+        b._get_token_balance = _make_gen(5_000_000)  # type: ignore[assignment,method-assign]
```
DhairyaPatel7: Weak assertion: `_get_token_balance = _make_gen(5_000_000)` returns 5M for every iteration of WHITELISTED_ASSETS['optimism'] (6 tokens), so `balances` ends up with 6 entries. The test only asserts `'USDC' in symbols`, which would still pass even if the loop accidentally short-circuited after the first token or used the wrong symbol mapping. Tighten by asserting `len(balances) == 6`, the full symbol set, and that each entry's balance equals 5_000_000 — or switch to a per-address fake (as the integration test at line 1162 already does) so the test pins symbol↔address pairing.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -524,16 +524,21 @@ def _get_optimism_balances_from_safe_api(
                         }
                     )
 
+        # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
+        # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
+        # returns an incomplete trusted-token list. Covers every entry in
+        # WHITELISTED_ASSETS[chain] including oUSDT, which used to be fetched
+        # separately. Reward tokens (OLAS, VELO) are disjoint from the
+        # whitelist and handled by _fetch_reward_balances below.
+        yield from self._supplement_with_onchain_whitelisted_balances(
```
DhairyaPatel7, line 533: Scope of the fallback: dedup is by address, so SafeApi entries always win over the on-chain reading. This is fine for the documented 'SafeApi returns empty/missing' case, but if SafeApi ever returns a *stale* or zero-valued whitelisted entry (e.g. indexing lag where USDC is listed but at last-week's balance), the backstop will skip it and the investor/withdrawal flows still see the wrong value. Not a regression — just confirming this PR is narrower than its motivation ('positions reflects on-chain reality whenever SafeApi is unavailable') suggests. If the broader goal is durable correctness, a follow-up could replace SafeApi entries with on-chain values when they disagree by more than some tolerance.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Good catch, this did regress the resiliency posture. Fixed in 6f809949: wrapped both the native fetch and the per-token loop in try/except matching `_fetch_reward_balances`. A per-token raise now logs a warning and continues to the next token, and a native raise logs and falls through to the ERC20 loop. Added `test_token_rpc_exception_continues_to_next_token` and `test_native_rpc_exception_does_not_block_erc20s` to pin the behaviour so the guards can't be silently removed later.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 6f809949. Added a matching warning when `_get_native_balance` returns None so silent framework misses are visible in logs. `test_native_none_does_not_block_erc20s` pins it. Kept the explicit `> 0` check since `None` is now handled on a separate branch and the `> 0` is the only thing left expressing the positive-balance intent.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_base.py
DIvyaNautiyal07: Agreed, fixed in 6f809949. Switched to a per-address balance fake (one distinct balance per `WHITELISTED_ASSETS["optimism"]` entry, keyed by address) and now assert `len(balances) == 6` plus each entry's `asset_symbol`, `asset_type`, and `balance` against the expected map. A loop short-circuit or a symbol-to-address mapping swap would now surface as a missing key or a mismatched balance.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Agreed this is narrower than the stale-but-present case. The scope here is the SafeApi-unavailable failure (empty or missing returns), which is what we're actually seeing in support tickets. Replacing live SafeApi values with on-chain reads is a different design question (tolerance threshold, which source wins on disagreement, how to handle indexing lag during normal operation), so I'd rather spin that as a follow-up than expand the scope of this PR.

LOCKhart07 (commented):
**Independent review — no new inline comments (DhairyaPatel7's 4 already cover the substantive issues).**

Agreeing the **exception-safety finding is blocking and must-fix before merge**: `_supplement_with_onchain_whitelisted_balances` wraps neither the native nor the per-token RPC reads in try/except, unlike its siblings `_fetch_reward_balances` (try/except per iteration) and the deleted `_fetch_ousdt_balance` (had its own try/except). It's called bare from `_get_optimism_balances_from_safe_api`, which is called bare from `get_positions` — so a single RPC raise on any one whitelisted token now aborts the entire positions round. That regresses the exact resiliency posture this PR exists to restore. Wrapping the native fetch and the loop body in `try/except Exception: logger.warning(...); continue` (mirroring `_fetch_reward_balances`) resolves it.

Things I verified clean during an independent pass, for confidence:
- **No Mode regression** from removing `_fetch_ousdt_balance` — on `main` it was only ever called from the Optimism path.
- **No double-count** — `REWARD_TOKEN_ADDRESSES["optimism"]` (OLAS, VELO) is disjoint from `WHITELISTED_ASSETS["optimism"]`, so the backstop and `_fetch_reward_balances` don't overlap.
- **Dedup is correct** — SafeApi addresses (checksummed → lowered) match the already-lowercase whitelist keys case-insensitively.
- `test_backstop_integrates_when_safeapi_partial` genuinely exercises the live helper (not the no-op stub), and CI is fully green.

LOCKhart07 (approved):
Re-reviewed at `6f809949`. The blocking exception-safety regression is genuinely fixed and well-tested — **approving**.

All four prior review comments resolved:
- **Exception safety**: native fetch and per-token loop body now wrapped in `try/except Exception: logger.warning(...)`, matching the `_fetch_reward_balances` idiom. Two falsifiable regression guards added (`test_token_rpc_exception_continues_to_next_token`, `test_native_rpc_exception_does_not_block_erc20s`) — removing either try/except fails them.
- **Native None diagnostic**: now warns symmetrically with the token side; redundant `> 0` folded into a clear `elif`. `test_native_none_does_not_block_erc20s` added.
- **Weak test assertion**: `test_adds_whitelisted_erc20_when_missing` switched to a per-address fake; pins `len == 6`, symbol↔address mapping, `asset_type`, and exact per-token balance.
- **Scope note**: acknowledged and accepted as intentionally narrow (SafeApi-unavailable, not stale-but-present).

No new issues in the follow-up commit: the broad `except Exception` around `yield from` is consistent with the codebase's own `_fetch_reward_balances` pattern, and every branch of the helper now has falsifiable coverage. CI fully green (24/24), including the Py3.14 coverage job.

DhairyaPatel7 (approved):

DhairyaPatel7 (commented):
Re-review: all 4 prior issues addressed correctly (exception safety wrapped both branches per the _fetch_reward_balances pattern, native None warning added, ERC20 test tightened with a per-address fake and full assertion set). Three new regression tests pin the guards. No new issues.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -524,16 +524,21 @@ def _get_optimism_balances_from_safe_api(
                         }
                     )
 
+        # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
+        # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
+        # returns an incomplete trusted-token list. Covers every entry in
+        # WHITELISTED_ASSETS[chain] including oUSDT, which used to be fetched
+        # separately. Reward tokens (OLAS, VELO) are disjoint from the
```
OjusWiZard, line 531: The disjointness claim is true for Optimism but **false for Mode**: `WHITELISTED_ASSETS["mode"]` and `REWARD_TOKEN_ADDRESSES["mode"]` share OLAS, VELO, and USDC at identical addresses. Today the call site hardcodes `"optimism"` so this is latent, but the natural follow-up implied by the PR description (extending the backstop to the Mode Explorer path) would cause every one of those three tokens to be double-counted: the backstop appends them, then `_fetch_reward_balances` extends the same list without deduping against `balances`.

Minimum-friction options: (a) scope this comment to Optimism explicitly (`Reward tokens (OLAS, VELO) are disjoint from the Optimism whitelist`), or (b) add a one-line address-dedup pass inside `_fetch_reward_balances` so the contract holds for both chains. Either is fine for this PR; the structural cleanup of `WHITELISTED_ASSETS["mode"]` belongs in the Mode follow-up.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -524,16 +524,21 @@ def _get_optimism_balances_from_safe_api(
...
+        # whitelist and handled by _fetch_reward_balances below.
+        yield from self._supplement_with_onchain_whitelisted_balances(
+            "optimism", safe_address, balances
```
OjusWiZard, line 534: Nit: this same function uses `Chain.OPTIMISM.value` at line 437 (and `Chain.MODE.value` at line 616). Worth using `Chain.OPTIMISM.value` here too for consistency, so a future rename of the enum value can't silently desync this call site.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -3477,6 +3482,90 @@ def _fetch_reward_balances(
...
+        remain diagnosable.
+
+        :param chain: the chain to read on-chain balances from.
+        :param safe_address: the Safe contract address to read balances for.
+        :param balances: existing balances list to supplement in place.
+        :yield: None, for async-style generator progression.
+        """
+        already_have = {
+            (b.get("address") or "").lower() for b in balances if b.get("address")
+        }
```
OjusWiZard, line 3509: Nit: the `if b.get("address")` guard already excludes entries with a missing/empty address, so `(b.get("address") or "").lower()` can't actually hit the `or ""` branch — the fallback is unreachable. Simpler:

```python
already_have = {
    b["address"].lower() for b in balances if b.get("address")
}
```

(Three of the six review agents flagged this independently, so flagging once here.)

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -3477,6 +3482,90 @@ def _fetch_reward_balances(
...
+
+        zero_addr_lc = ZERO_ADDRESS.lower()
```
OjusWiZard, line 3511: Nit: `ZERO_ADDRESS` is defined at line 76 as `"0x0000...000"` — already lowercase, so `.lower()` is a no-op. The local `zero_addr_lc` can be dropped in favour of using `ZERO_ADDRESS` directly at lines 3512 and 3531.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -3477,6 +3482,90 @@ def _fetch_reward_balances(
...
+                        f"On-chain native balance read returned None on {chain}; "
+                        f"skipping. Positions data may be incomplete."
+                    )
+                elif native_balance > 0:
+                    balances.append(
+                        {
+                            "asset_symbol": "ETH",
+                            "asset_type": "native",
+                            "address": to_checksum_address(ZERO_ADDRESS),
+                            "balance": int(native_balance),
```
OjusWiZard, line 3528: Nit: `int(native_balance)` is redundant — `_get_native_balance` already wraps its return in `int(...)` at line 1071 and returns either `int` or `None` (the `None` branch is handled above), so by the time we reach this line `native_balance` is guaranteed to be a positive int. Worth keeping the `int(token_balance)` cast at line 3558, though — that value comes through the contract plugin and isn't statically guaranteed to be int.

OjusWiZard (approved):
Approve. Independent multi-perspective pass found no Critical/High issues; the exception-safety and test-pinning regressions raised by DhairyaPatel7 / DIvyaNautiyal07 are correctly addressed in 6f80994.

**One non-blocking follow-up worth a test in this PR if convenient:** the `WHITELISTED_ASSETS.get(chain, {})` branch is unreachable today (call site hardcodes `"optimism"`), but the native-ETH block at lines 3512-3535 still runs for any chain — including unknown ones — because it isn't gated on the chain being in the whitelist dict. A `test_unknown_chain_skips_erc20_loop_but_still_attempts_native` (or equivalent) pins both halves and protects the 100% coverage rule against future call-site additions.

Inline comments below are the small items worth picking up; none of them block.

PR #322, opened 2026-05-21
feat(config): restore + extend named env vars in aea-config.yaml

DhairyaPatel7:
## Summary

Re-applies the named `${VAR:type:default}` syntax to `packages/valory/agents/optimus/aea-config.yaml`, **plus** names new operator-tunable parameters added since the original work.

The original named-env-var work (commit `8e401433`, PR #266 follow-up) was reverted in `4254a296` because open-autonomy and the middleware lacked support at the time; both have since shipped that support. `service.yaml` overrides have already been carrying named vars.

**Restored (originally named, then reverted):** 13 vars — LOG_*, STORE_PATH, MODE_LEDGER_RPC, COINGECKO_API_KEY, ALL_PARTICIPANTS, SAFE_CONTRACT_ADDRESSES, LOG_DIR, TARGET_INVESTMENT_CHAINS.

**Newly named (added to agent after the original PR):**
- `FILE_HASH_TO_STRATEGIES`
- `MODE_CONDUIT_EXPLORER_URL`
- `MODE_NATIVE_EXPLORER_URL`
- `SAFE_API_V1_URL`
- `TLS_VERIFY`

**Intentionally left anonymous:** framework/protocol internals (Tendermint timeouts, consensus settings, HTTP method strings, fixed protocol addresses).

Hashes re-locked via `autonomy packages lock`.

## Test plan

- [x] `tomte tox -e check-hash`, `check-packages`
- [x] `tomte tox -e check-abciapp-specs`, `check-handlers`
- [x] `tomte tox -e check-dependencies`, `check-third-party-hashes`
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e darglint`, `pylint`
- [x] `tomte tox -p -e safety -e bandit`
- [x] gitleaks — fails the same way on `origin/main` (pre-existing UI build artifacts), unrelated to this PR
- [ ] Smoke-test agent run to confirm middleware resolves the named vars

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -25,8 +25,8 @@
         "connection/valory/mirror_db/0.1.0": "bafybeiasfxdd447vqut4uercswnn5pagy555oj2ws7deo2p7vnmq3mj64y",
         "skill/valory/liquidity_trader_abci/0.1.0": "bafybeiett4gnks5olb4zzgsyumcd5ytatqfrwec5a76uph7z3zok2ghjr4",
         "skill/valory/optimus_abci/0.1.0": "bafybeibxcsqg5r6bo2v6wmekepmw6gda75cvqohfrdjyyvasjrdx67wqgq",
-        "agent/valory/optimus/0.1.0": "bafybeievffp7fft2upenrkcfjevmwehi3nl4n3vurnwxovnryucvwkerti",
-        "service/valory/optimus/0.1.0": "bafybeidutvl4k3zg2obptloa2lp3d2ro4xh7kun64eai5d2niwib5lmjmu"
+        "agent/valory/optimus/0.1.0": "bafybeiavtoxghlfz5li4hw76wyl57g42bk7pvfjjfk3pbqoeu3qtqvmwae",
+        "service/valory/optimus/0.1.0": "bafybeic4yof22rkygd4qpsbbjsc44ixkrhxo2omefbujvc4ecbmoae5akq"
     },
     "third_party": {
         "protocol/open_aea/signing/1.0.0": "bafybeifsjmldwyki3beqyvdt5lzenrg6wyrqaar5plc5rpnvtc4zlentye",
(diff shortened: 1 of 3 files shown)


LOCKhart07 (approved):
Verified the substance against history, not just the description:

- **Restored 13 vars are faithful** — the revert `4254a296` removed exactly these 13 named-var lines, and the additions here match them 1:1.
- **5 newly-named vars match `service.yaml`** exactly (`TLS_VERIFY`, `MODE_CONDUIT_EXPLORER_URL`, `SAFE_API_V1_URL`, `MODE_NATIVE_EXPLORER_URL`, `FILE_HASH_TO_STRATEGIES`), so agent- and service-level names are consistent.
- **Types preserved** on every var (no `str`↔`int`/`dict`/`bool`/`list` drift).
- **Hashes consistent** — new agent hash matches between `packages.json` and `service.yaml`; no stale references to the old agent/service hashes remain.
- **No scope creep** — only the 3 expected files; naming + relock only.

One gating item before merge: the test plan leaves **`Smoke-test agent run to confirm middleware resolves the named vars` unchecked**. That runtime check is the exact failure mode that caused the original revert (`4254a296`: framework/middleware lacked support). Every static check passes, but it's the only thing that confirms the PR's premise that named **agent-level** vars now resolve. Worth completing before merge.

FYI (not a defect): `SAFE_CONTRACT_ADDRESSES` is intentionally dual-typed — `dict` (funds_manager, 1 chain) and `str` (params, 4 chains) under one env var name. Matches `service.yaml` and the original commit, so it's the established contract.

DIvyaNautiyal07 (approved):

DhairyaPatel7:
Thanks for the thorough audit. Re: the unchecked smoke-test — agreed, that's a hard gate before merge given it's the exact failure mode the original revert tripped on. Will run a Pearl deploy of the service and verify each named var resolves at runtime, then flip the box and mark ready for merge.

Noted re: `SAFE_CONTRACT_ADDRESSES` dual-typed convention — preserved as-is.

DIvyaNautiyal07 (approved):

PR #323, opened 2026-05-27
refactor(liquidity-trader): remove mirror_db connection and APR population round

DIvyaNautiyal07:
## Summary
- Removes the `mirror_db` connection entirely. The afmdb.autonolas.tech sink had no remaining consumer, so the APR storage round, the opportunity-metrics writes in EvaluateStrategy, and the SRR plumbing inside `liquidity_trader_abci` are all dead weight and have been stripped.
- FSM rewire: `GetPositionsRound` DONE now transitions directly to `EvaluateStrategyRound`. One fewer consensus round per cycle. `WITHDRAWAL_INITIATED` routing is preserved — six other rounds still emit it.
- Kept on purpose: `apr_threshold` (load-bearing for `merkl_pools_search` low-APR filtering), `selected_hyper_strategy` (still used by `max_apr_selection`), SRR protocol + dialogues at the `optimus_abci` / agent level (the `genai` LLM flow consumes them).

## Test plan
- [x] `tomte tox -e check-abciapp-specs` — FSM YAMLs match `rounds.py`
- [x] `tomte tox -e check-abci-docstrings` — round docstring matches transition function
- [x] `tomte tox -e check-handlers` — handler registrations consistent
- [x] `tomte tox -e check-packages` — package-yaml consistency
- [x] `tomte tox -e check-hash` — IPFS hashes match local files
- [x] `tomte tox -e check-dependencies` — deps consistent between packages and tox
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — full lint pass
- [x] `pytest packages/valory/skills/liquidity_trader_abci/tests` — 2711 passed
- [x] `pytest packages/valory/skills/optimus_abci/tests` — 319 passed
- [x] `autonomy packages lock --check` — package hashes verified
- [x] `autonomy push-all` — published to IPFS registry
- [ ] CI on Linux (this PR triggers it)
- [ ] One staged agent boot to confirm `GetPositionsRound → EvaluateStrategyRound` transition in a live service

## Notes for reviewers
- Side cleanup beyond the original "remove DB" scope: also removed `_track_opportunities` and `_format_opportunity_for_tracking` because the only consumer was the mirror_db push that this PR deletes. Their KV writes (`opportunity_tracking`) are gone too. The orphan `last_apr_calculation` and `opportunity_tracking` keys in the local KV store are intentionally left alone — they'll sit unread.
- `apr_threshold` was initially flagged as DB-only by exploratory grep. Verification of `merkl_pools_search.py:33,72,132` showed it filters Merkl campaigns by APR. Kept.


OjusWiZard (dismissed):
Approving. Correctness is solid — FSM is self-consistent, all six `WITHDRAWAL_INITIATED` emitters verified, no dangling references survive in live code, deleted tests covered only deleted code, and the intentionally-preserved symbols (`apr_threshold`, `selected_hyper_strategy`, SRR at `optimus_abci`/`genai` level) all check out.

Three follow-ups worth landing in this PR before merge — none block.

---

**M1 — Dead method `sign_message`** &nbsp;·&nbsp; `packages/valory/skills/liquidity_trader_abci/behaviours/base.py:3079`

The only production callers were the deleted `APRPopulationBehaviour` and the deleted `create_attribute_definition` / `create_agent_attribute` helpers. A grep across the skill shows zero remaining call sites; only `test_base.py:2026-2039` still references it. Delete the method and its two tests — or, if you want to keep it for future use, justify it with a comment so it isn't load-bearing only on the test suite.

---

**M2 — Missing test for the central FSM rewire** &nbsp;·&nbsp; `packages/valory/skills/liquidity_trader_abci/tests/test_rounds.py`

The PR's headline change is `GetPositionsRound[Event.DONE] → EvaluateStrategyRound`. `TestWithdrawalTransitionMap` locks in the withdrawal set cleanly, but no assertion locks in this specific mapping. A future accidental revert would slip past CI even with 100% line coverage. Suggest:

```python
def test_get_positions_done_routes_to_evaluate_strategy() -> None:
    assert (
        LiquidityTraderAbciApp.transition_function[GetPositionsRound][Event.DONE]
        is EvaluateStrategyRound
    )
```

---

**L1 — Redundant local `import math`** &nbsp;·&nbsp; `packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py:201`

`math` is already imported at module scope on line 24. The local reimport inside the `try` block in `validate_and_prepare_velodrome_inputs` is dead weight.

```diff
         try:
-            import math
-
             current_tick = int(math.log(current_price) / math.log(1.0001))
```

---

**Advisory (pre-existing, out of scope):** two broad `except Exception` blocks in `evaluate_strategy.py` swallow stack traces and steer the FSM toward optimistic defaults — `_apply_tip_filters_to_exit_decisions` at line 1227 (returns "all open positions eligible to exit") and `check_and_prepare_non_whitelisted_swaps` at line 1408 (silently returns `[]` so non-whitelisted assets accumulate without conversion). Both predate this branch. Worth a follow-up ticket — switching `logger.error` → `logger.exception` is a zero-risk diagnostic upgrade.

DIvyaNautiyal07:
Thanks Ojuswi. Follow-up commit e9c40c9d addresses all three.

M1 done. Deleted sign_message and generate_name, plus the two phonetic helpers that only fed generate_name. Pulled the associated tests out of test_base.py too.

M2 done. Added TestCoreCycleTransitions in test_rounds.py with two assertions, GetPositionsRound DONE -> EvaluateStrategyRound and EvaluateStrategyRound DONE -> DecisionMakingRound. An accidental revert of the rewire will trip CI now.

L1 done. Dropped the inline import math.

Agreed on the advisory, leaving the broad except Exception blocks for a follow-up ticket since they predate this branch. The logger.error to logger.exception swap is a clean isolated change that doesn't belong here.

LOCKhart07 (commented):
Three stale documentation refs to clean up before merge — all non-blocking but they'd land as immediate post-merge inconsistencies. None of these files are in the PR's diff, so flagging here.

**`.gitignore:30`** — `!packages/valory/connections/mirror_db` is a "don't ignore" entry for a directory that no longer exists. After this PR there are no `valory/`-authored connections, so the line is dead and can be removed.

**`CLAUDE.md:65`** — `- **connections/** — \`mirror_db\` connection using Peewee ORM.` is doubly stale: `mirror_db` is deleted, and the deleted connection used `aiohttp` (not Peewee), so the Peewee description was already wrong. The `packages/valory/connections/` tree is now empty — this bullet should be deleted or rewritten.

**`README.md:99-103`** — Still has the `#### APRPopulationRound` section describing the deleted round ("Calculates the APR and ETH-adjusted APR of the agent / Creates a portfolio snapshot stored in the centralized database"). Should be removed; the section currently sits between `GetPositionsRound` and `EvaluateStrategyRound`, mirroring the old FSM order.

Otherwise the deletions look clean: I verified no orphaned imports or callers for the removed mirror_db / APR-population code paths, the FSM rewires correctly in both skills, `apr_threshold` is genuinely used in `merkl_pools_search.py:33,72,132`, SRR protocol/dialogues are retained at the `optimus_abci` + agent level for the genai flow, and CI is green on all 15 test jobs.

DIvyaNautiyal07:
Thanks. Pushed commit 68e3b35f with all three. .gitignore line gone, the connections/ bullet removed from CLAUDE.md (you were right about the Peewee bit too, that was wrong from the start), and the APRPopulationRound section deleted from README.md.

OjusWiZard (approved):

LOCKhart07 (approved):
All prior findings addressed:

- **Doc cleanup** (`68e3b35f`): `.gitignore`, `CLAUDE.md`, and `README.md` references all gone.
- **M1** (`e9c40c9d`): `sign_message`, `generate_name`, and the phonetic helpers removed along with their `test_base.py` cases.
- **M2** (`e9c40c9d`): `TestCoreCycleTransitions` pins both `GetPositionsRound -> EvaluateStrategyRound` and `EvaluateStrategyRound -> DecisionMakingRound`. Bonus `test_post_tx_settlement_does_not_route_withdrawal` and `test_withdrawal_initiated_targets_withdraw_funds_round` are nice invariant pins.
- **L1** (`e9c40c9d`): redundant local `import math` gone.

One small unrelated cleanup in `evaluate_strategy.py:1678-1701` — the `valid_opportunities` buffer is replaced with direct `self.trading_opportunities.append`. Functionally equivalent (the `try/except` is per-iteration, so exception semantics are unchanged) and it's a minor simplification; flagging only for scope-creep visibility, not a concern.

Linter / lock / security checks are green on the new commits; the 15 test jobs are still pending. Approving on the assumption they pass.

PR #328, opened 2026-06-01
fix(safe-api): cache USDC withdrawals + transferId-keyed dedup (WS1 1.1+1.2)

DIvyaNautiyal07:
## Summary

Implements **workstream 1 PRs 1.1 and 1.2** from the Basius launch doc: https://docs.google.com/document/d/1R_ZuAV4wFSaiwXaQFo2TBXrtQMRCZAum8X9CzSWijiE/edit?tab=t.1xsq6yndq2gv

Branched off `hack/launch-base`.

**1.1 — Withdrawal cache + early-stop.** `_track_erc20_transfers_optimism` no longer walks the full `/transfers/` history every FetchStrategies cycle. It now mirrors the existing incoming-transfer pattern:
- Persists outgoing USDC transfers to `funding_events.json` under a new `optimism_withdrawals` key.
- Applies the `consecutive_existing` early-stop after page 1 when nothing new shows up.
- `calculate_withdrawals_value` is fronted by a same-day kv cache (`optimism_total_withdrawals` + `last_withdrawals_calculated_timestamp`) — first cycle of the day computes, subsequent same-day cycles return cached without any Safe API call.

Day-T cycle 2 onwards on Optimism now costs **0 withdrawal calls** instead of N pages.

**1.2 — TransferId-keyed dedup across all transfer fetchers.** Today, once a date is in `funding_events.json`, any later transfer on the same day is silently dropped at `fetch_strategies.py:3890` and equivalents — a 10 AM funding plus a 4 PM top-up loses the top-up forever. This PR replaces date-keyed dedup with per-transfer dedup:
- Two new module-level helpers: `_transfer_unique_id` (prefers Safe's `transferId`, falls back to `txHash + log_index`, then a legacy `txHash:type:value` shape for entries persisted before this change) and `_collect_seen_transfer_ids`.
- All three Optimism transfer fetchers (`_fetch_optimism_transfers_safeglobal`, `_fetch_outgoing_transfers_until_date_optimism`, `_track_erc20_transfers_optimism`) now seed their seen-set from persisted data and dedup per uid.
- Each persisted transfer dict now carries a `transfer_id` field so future cycles can rebuild the seen-set cheaply.
- The merge sites switch from `existing.update(new)` (overwrites date key) to `existing.setdefault(date, []).extend(new)` so a same-day event landing after the date key already exists is preserved alongside earlier entries.

## Per-cycle Safe API cost (Optimism, steady state)

Before this PR (on `hack/launch-base`):
- Day-T cycle 1: 1 balances + N withdrawal pages + 1 incoming = **2 + N**
- Day-T cycle 2 onwards same day: 1 balances + N withdrawal pages + 0 = **1 + N**

After this PR:
- Day-T cycle 1: **3 calls** (1 balances + 1 withdrawal + 1 incoming, all with early-stop)
- Day-T cycle 2 onwards same day: **1 call** (balances only)

With a 3-page transfer history on a 5-minute cycle, that's roughly a 30x reduction in Safe API traffic, well inside the authenticated free-tier headroom.

## Out of scope (separate PRs per the launch plan)

- **PR 1.3** — shorten the kv TTL from "today vs not today" to ~30 minutes. Until that lands, same-day top-ups still won't be picked up because the kv cache (both `last_initial_value_calculated_timestamp` and the new `last_withdrawals_calculated_timestamp`) short-circuits the recompute. 1.2 fixes the per-transfer dedup at the fetcher level, so when the recompute does fire, it now works correctly.
- **PR 1.4** — proxy-safe pagination (strip absolute `next` URLs).
- **Workstream 2** — ETH-reversion deletions.

## Tests

- **14 new tests** at the bottom of `test_fetch_strategies.py` covering:
  - `_transfer_unique_id` (transferId preferred, txHash+logIndex fallback, legacy shape fallback, empty when no id, logIndex=0 kept).
  - `_collect_seen_transfer_ids` (empty inputs, mixed API + legacy shapes, malformed entry tolerance).
  - `_track_erc20_transfers_optimism` (cold-start persist under new key with `transfer_id` field, warm-start early-stop after one page, same-day top-up with new transferId lands).
  - `calculate_withdrawals_value` (cache hit skips fetcher, cache miss recomputes + writes kv, no prior timestamp seeds the cache).
- **3 pre-existing tests updated** to the new behavioural contract — uid-based dedup replaces date-based skip, kv stubs added where the function now reads/writes them.
- `pytest packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py` — all 539 tests pass.

## Test plan

- [ ] Run a real Optimism agent against a funded Safe for a few cycles; confirm `optimism_withdrawals` appears in `data/funding_events.json` and `optimism_total_withdrawals` + `last_withdrawals_calculated_timestamp` appear in the kv_store.
- [ ] Confirm the second cycle of the same day issues no Safe API call on the withdrawal path (logs should show "Last withdrawal calculation was today, using cached value").
- [ ] Confirm that simulating a same-day top-up (manually patching the agent's Safe with a new outgoing USDC transfer) results in the new transfer being persisted on the next recompute.


packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2366,6 +2440,15 @@ def calculate_withdrawals_value(self) -> Generator[None, None, Decimal]:
                 )
             )
             self.context.logger.info(f"Withdrawal value: ${withdrawal_value}")
+
+            yield from self._write_kv(
+                {
+                    "optimism_total_withdrawals": str(withdrawal_value),
```
OjusWiZard, line 2495: 🔴 **Critical — a transient Safe-API error caches `withdrawal = 0` for the rest of the day.**

This KV write is unconditional. When `_track_erc20_transfers_optimism` hits its `except Exception` (`:4862`) it returns `{"outgoing": {}}`, which is **truthy**, so the `if not all_erc20_transfers_optimism:` guard at `:2426` does **not** fire. Execution falls through, computes `withdrawal_value = Decimal(0)`, and writes `optimism_total_withdrawals="0"` + today's timestamp here. Every subsequent same-day cycle then hits the cache (`:2402`) and returns `0` without retrying.

Pre-PR there was no cache, so an error self-healed on the next 5-min cycle. Now one timeout/5xx/parse blip on the day's first cycle suppresses the withdrawal figure for up to 24h — inflating net portfolio value and any profit/KPI math that subtracts withdrawals.

*Fix:* return a distinct error sentinel (`None`) on the exception path — matching the Mode path the caller already handles at `:2372` — and skip this KV write when the fetch failed. Don't cache a value derived from an error fallback.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4728,24 +4827,35 @@ def _track_erc20_transfers_optimism(
...
                 if not cursor:
                     break
                 transfers_url = cursor
 
-            return all_transfers
+            # Merge new transfers into persisted data and save
+            for date, transfers_list in new_by_date.items():
+                existing_outgoing.setdefault(date, []).extend(transfers_list)
+
+            self.funding_events["optimism_withdrawals"] = existing_outgoing
```
OjusWiZard, line 5037: 🟠 **Important — mid-pagination failure persists partial data and can permanently drop older withdrawals.**

When a page fetch fails mid-stream, `if not success: break` (`:4743`) exits the loop and falls through to this merge + `store_funding_events()`. If page 1 succeeded and page 2 failed, page-1 transfers get persisted and seeded into next cycle's `seen` set. Usually page 2 is recovered next cycle — **except** when the early-stop (see I2) fires on the now-fully-seen page 1, in which case pagination stops before reaching the still-missing older pages and those withdrawals are lost permanently.

New risk from the persistence layer (pre-PR built results fresh and never persisted partials).

*Fix:* track a `fetch_failed` flag and skip the merge + `store_funding_events()` when a page fetch failed mid-stream.

OjusWiZard (commented):
Automated multi-perspective review (Claude Code). 4 inline findings below (1 critical, 3 important). Confidence ≥ 80.

**Advisory (not inlined):**
- **Tests:** no coverage for the exception→cache-0 chain (C1); `TestErc20Optimism.test_exception` asserts `== {"outgoing": {}}`, locking in the buggy return. No integration test for legacy-entry dedup (I3), interleaved-page early-stop (I2), or the corrupt-`cached_val` fall-through at `:2415`.
- **TZ:** `calculate_withdrawals_value` mixes `datetime.now()` (local, `:2391`) with `datetime.utcfromtimestamp` (`:2399`) — off-by-a-day risk near UTC midnight. Mirrors the pre-existing `calculate_initial_investment_value` convention, so not introduced here, but worth aligning both on UTC.
- **Simplify:** the dedup + `consecutive_existing` early-stop block and the `setdefault().extend()` merge are now near-duplicated across all three Optimism fetchers; the same-day KV-cache block duplicates the one in `calculate_initial_investment_value`. Candidates for a shared `_check_seen(...)` helper and `_is_same_day_as_last_calc(timestamp_key)` generator (keep the `seen.add()` boundary local — it differs per fetcher).

The KV-cache design (returning cumulative persisted data) and the dedup-by-transferId direction are sound. C1 is the merge blocker.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4676,23 +4774,23 @@ def _track_erc20_transfers_optimism(
                     if tx_date > current_date:
                         continue
 
+                    # Per-transfer dedup against persisted + in-cycle seen IDs.
+                    unique_id = _transfer_unique_id(transfer)
+                    if unique_id and unique_id in seen_tx_ids:
```
OjusWiZard: 🟠 **Important — the warm-cycle early-stop rarely fires on the `/transfers/` fetchers.**

`consecutive_existing` resets to 0 (just below) for *any* transfer not in the seen-set, and the seen-set only ever contains persisted USDC-outgoing uids. The `/transfers/` endpoint returns mixed incoming + ETH + non-USDC transfers newest-first, so each of those resets the counter and `consecutive_existing >= len(transfers)` almost never holds. The docstring's "early-stop after one page" won't hold against realistic mixed traffic — it paginates the full history on every cache-miss.

Not a correctness bug (the function returns cumulative persisted data so the sum stays right, and the KV cache still delivers the "0 calls on cycle 2+" win), but the secondary optimization is largely inert. Consider counting only outgoing-USDC against the threshold, or soften the claim. (Same pattern in `_fetch_optimism_transfers_safeglobal` and `_fetch_outgoing_transfers_until_date_optimism`.)

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -97,6 +97,49 @@
...
+        log_idx = transfer.get("log_index")
+    if tx_hash and log_idx is not None:
+        return f"{tx_hash}:{log_idx}"
+    if not tx_hash:
+        return ""
+    transfer_type = transfer.get("type", "")
+    raw_value = transfer.get("value")
+    if raw_value is None:
+        raw_value = transfer.get("amount", "")
+    return f"{tx_hash}:{transfer_type}:{raw_value}"
```
OjusWiZard, line 126: 🟠 **Important — legacy persisted entries don't dedup against API `transferId` → one-time double-count on upgrade.**

Pre-PR persisted entries have `tx_hash`/`type`/`amount` but no `transfer_id` or `logIndex`, so they resolve to this fallback `f"{tx_hash}:{type}:{amount}"`. The same transfer from the API resolves to its `transferId`, so the seeded seen-id never matches — the transfer is re-processed and, with the new `setdefault(date, []).extend(...)` merge, re-appended. On the first post-deploy cycle for an agent with an existing `funding_events.json`, every historical incoming/outgoing transfer is duplicated once, inflating `calculate_initial_investment_value`.

The new `optimism_withdrawals` key is unaffected (fresh key); exposure is on the existing `optimism` / `optimism_outgoing` keys. Fresh agents are unaffected.

*Fix:* a one-shot migration that drops entries lacking `transfer_id` before seeding, or back-fill `transfer_id` on load.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 89938537. `_track_erc20_transfers_optimism` now returns `None` on exception (the bare `except` path) and on first-page failure. `calculate_withdrawals_value` treats `None` as "don't cache" and skips the kv same-day write, so a transient API blip no longer pins `withdrawal=0` for the rest of the day — next cycle retries fresh. Test: `TestCalculateWithdrawalsValueSameDayCache::test_fetcher_failure_skips_kv_write`.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 89938537. Added a `fetch_failed` flag that gets set on any `success == False` from `_request_with_retries`. After the loop, if `fetch_failed` is true, the function discards the partial `new_by_date`, skips the merge + `store_funding_events()`, and returns `None`. This prevents the failure mode where page-1's transfers get persisted and seeded into the next cycle's seen-set, causing the early-stop to fire on a fully-seen page 1 and silently drop older pages. Test: `TestErc20OptimismFailureModes::test_mid_pagination_failure_discards_partial_and_returns_none`.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 89938537 for `_track_erc20_transfers_optimism`. A pre-pass over the page identifies outgoing-ERC20 entries (the only kind this fetcher cares about) and the counter increments only against that subset. The early-stop trips when all eligibles on the page are seen, regardless of unrelated incoming / ETH / non-ERC20 entries also on the page. The docstring's "early-stop after one page" claim now holds.

Left the same pattern in `_fetch_optimism_transfers_safeglobal` and `_fetch_outgoing_transfers_until_date_optimism` untouched: both had persistence (and the same early-stop limitation) pre-PR, so the behavior there is not a regression introduced by this change. Happy to apply the same eligible-only pattern in a follow-up if you'd prefer it consistent across all three fetchers — just confirm.

Test: `TestErc20OptimismFailureModes::test_eligible_only_early_stop_with_mixed_page`.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 89938537. New module-level helper `_drop_legacy_transfer_entries` strips persisted entries that lack the `transfer_id` field. Applied at both load sites for the legacy keys:
- `_fetch_all_transfers_until_date_optimism` (reads `funding_events["optimism"]`)
- `_fetch_outgoing_transfers_until_date_optimism` (reads `funding_events["optimism_outgoing"]`)

The migrated shape is written back to `self.funding_events` so subsequent loads don't need to re-filter and the on-disk file is cleaned of legacy entries the next time `store_funding_events()` fires. The first cycle after upgrade re-fetches the affected dates once; the early-stop keeps the cost bounded.

Chose drop-then-refetch over a backfill because the legacy fallback uid (`tx_hash:type:amount`) doesn't match what `_transfer_unique_id` will compute from a Safe API response (which uses `transferId`), so backfilling the legacy uid wouldn't actually dedup against API responses next cycle.

Tests: `TestDropLegacyTransferEntries::test_drops_entries_without_transfer_id` (helper), `test_applied_when_loading_optimism_outgoing` (end-to-end at the load site).

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3355,12 +3477,11 @@ def _fetch_all_transfers_until_date_optimism(
                 address, end_date, all_transfers_by_date, existing_optimism_data
             )
 
-            # Merge with existing data and save
+            # Merge new transfers into persisted data: append per-date so that
+            # a same-day transfer added after an earlier one already stored
+            # under the same date does not overwrite the earlier entries.
             for date, transfers in all_transfers_by_date.items():
```
LOCKhart07, line 3514: 🔴 **Blocking — migration + un-guarded partial-page persist can permanently delete `optimism` history.**

The C2 mid-pagination fix (discard partial data, return `None`) was applied only to `_track_erc20_transfers_optimism`. Here, the new `_drop_legacy_transfer_entries` migration (`:3452`) strips every legacy entry (none have `transfer_id`) and writes the emptied dict back to `self.funding_events["optimism"]`. The re-fetch then runs through `_fetch_optimism_transfers_safeglobal`, which on a mid-stream page failure does a silent `if not success: break` (`:3979`) — no exception. Execution falls through to this merge + `store_funding_events()`, persisting **page-1 only**.

Sequence: page 1 ok, page 2 fails → disk now holds only page-1; the stripped older pages are gone. Next cycle the seen-set is seeded from the page-1 entries (now carrying `transfer_id`), the early-stop fires on the fully-seen page 1, and pagination never reaches the missing pages — the history is never restored. This undercounts incoming transfers and deflates `calculate_initial_investment_value_from_funding_events`, inflating apparent profit.

Pre-PR this was harmless (page 1 was already stored, nothing was deleted) — the migration is what makes it destructive, so it's a regression introduced here, not the pre-existing limitation deferred earlier.

*Fix:* apply the same `fetch_failed` guard used in `_track_erc20_transfers_optimism` — skip the merge + `store_funding_events()` (and don't persist the stripped shape) when any page fetch failed mid-stream — or only write the migrated shape back after a fully-successful pagination.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4597,10 +4720,11 @@ def _fetch_outgoing_transfers_until_date_optimism(
                     break
                 transfers_url = cursor
 
-            # Merge new dates into persisted data and save
-            # (all_transfers only contains dates not already in existing_outgoing,
-            # because the fetch loop skips existing dates)
-            existing_outgoing.update(all_transfers)
+            # Merge new transfers into persisted data: append per-date so that
```
LOCKhart07, line 4816: 🔴 **Blocking — same migration + partial-persist data-loss as the `optimism` key, here for `optimism_outgoing`.**

`_drop_legacy_transfer_entries` (`:4589`) strips all legacy entries and writes the emptied dict back to `self.funding_events["optimism_outgoing"]`. On a mid-pagination page failure the loop does a silent `if not success: break` (`:4621`) and falls through to this merge + `store_funding_events()`, persisting only the partially-refetched pages and permanently dropping the stripped older history (the early-stop then prevents re-fetching it on later cycles).

*Fix:* same as the `optimism` fetcher — track a `fetch_failed` flag and skip the merge + store (and the write-back of the stripped shape) when a page fetch fails mid-stream.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4648,20 +4799,36 @@ def _track_erc20_transfers_optimism(
 
                 if not success:
                     self.context.logger.error("Failed to fetch Optimism transfers")
+                    fetch_failed = True
                     break
 
                 transfers = response_json.get("results", [])
 
                 if not transfers:
                     break
```
LOCKhart07, line 4902: 🟠 **Non-blocking — eligible-only early-stop still won't fire on pages with non-USDC outgoing ERC20s.**

`eligible_in_page` counts **all** outgoing `ERC20_TRANSFER`s, but `seen_tx_ids` only ever receives outgoing **USDC** transfers (the `.add` happens after the `symbol == "USDC"` and `amount > 0` checks). So a page containing any non-USDC outgoing ERC20 (or a zero-amount / future-dated one) can never reach `seen_eligible_in_page == eligible_in_page`, and the early-stop won't trip — the "early-stop after one page" claim only holds for USDC-only outgoing traffic.

Correctness is unaffected (the cumulative sum stays right and the kv cache still delivers the cycle-2 win). Consider narrowing the eligibility set to outgoing-USDC-ERC20 so it matches the seen-set, or softening the docstring claim.

LOCKhart07 (changes requested):

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4648,20 +4799,36 @@ def _track_erc20_transfers_optimism(
...
+                # consecutive-existing early-stop is counted against the
+                # outgoing-ERC20 subset only. Without this, unrelated entries
+                # (incoming, ETH, non-ERC20) reset the counter on every page
+                # and the early-stop almost never trips against realistic
+                # mixed Safe traffic.
+                eligible_indices = []
+                for idx, t in enumerate(transfers):
+                    if (
+                        t.get("from", "").lower() == safe_address.lower()
+                        and t.get("type", "") == "ERC20_TRANSFER"
```
OjusWiZard: 🟠 **N1 (re-review) — the eligible-only early-stop still can't fire for a safe that moves non-USDC tokens.**

The pre-pass denominator counts **all** outgoing ERC20s (`from == safe AND type == ERC20_TRANSFER`), but `seen_tx_ids` only ever contains **USDC** uids — it's seeded from the USDC-only `optimism_withdrawals` key and `seen_tx_ids.add()` at `:4890` runs *after* the `symbol.upper() != "USDC": continue` filter at `:4881`. So a non-USDC outgoing ERC20 (fee token, LP token, any other asset sent out) is counted in `eligible_in_page` but its uid can never enter `seen_tx_ids`, hence never increments `seen_eligible_in_page`.

Result: on any page containing ≥1 non-USDC outgoing ERC20, `seen_eligible_in_page == eligible_in_page` is unreachable → `stop_pagination` never trips → the fetcher walks the full history on every cache-miss. The commit's claim that "early-stop after one page now holds" only holds for a **pure-USDC-outgoing** safe; this is a narrowed version of the original I2 weakness.

Not a correctness/data bug (cumulative persisted data is still returned, and the same-day KV cache still gives the cycle-2+ win) — but it negates the headline cost claim for any safe with mixed outgoing tokens.

*Fix:* restrict the pre-pass denominator to USDC outgoing, e.g. add `and t.get("tokenInfo", {}).get("symbol", "").upper() == "USDC"` to the eligibility test so the denominator matches what actually becomes "seen". (Transfers with no inline `tokenInfo` need the async symbol lookup; the page response usually carries `tokenInfo`, so the common case is covered.)

Also: `TestErc20OptimismFailureModes` has no case with a non-USDC outgoing ERC20 alongside a seen USDC withdrawal on the same page — adding one would lock in this behavior. Minor: `eligible_indices` is built as a list but only its length is used; `sum(1 for t in transfers if ...)` is clearer.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4533,24 +4658,21 @@ def _fetch_outgoing_transfers_until_date_optimism(
                     if tx_date > current_date:
                         continue
 
-                    # Skip dates already persisted
-                    if tx_date in existing_outgoing:
+                    # Per-transfer dedup against persisted + in-cycle seen IDs.
```
OjusWiZard, line 4745: 🔵 **A1 (advisory) — eligible-only early-stop was applied to only one of the three fetchers.**

This fetcher (`_fetch_outgoing_transfers_until_date_optimism`) still uses the old `consecutive_existing >= len(transfers)` (`:4665`) against mixed incoming+outgoing `/transfers/` pages, so the original I2 weakness persists here — incoming/ETH entries reset the counter and the early-stop rarely trips. Lower priority since this isn't the same-day-cached hot loop, but worth the consistency. (The incoming fetcher at `:3997` is fine — it persists *all* incoming transfers, so its seen-set covers the full page.)

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -97,6 +97,79 @@
...
+    :return: the same shape with any legacy entries removed (dates that end up
+        empty are dropped).
+    """
+    if not date_keyed_data:
+        return date_keyed_data
+    migrated: Dict[str, Any] = {}
+    for date, transfers in date_keyed_data.items():
+        if not isinstance(transfers, list):
+            continue
+        kept = [t for t in transfers if isinstance(t, dict) and t.get("transfer_id")]
```
OjusWiZard: 🔵 **A2 (advisory) — `_drop_legacy_transfer_entries` can permanently drop unrecoverable history.**

Legacy entries (no `transfer_id`) are dropped on the expectation that the next fetch re-fetches them. But entries the Safe API can no longer return — beyond `MAX_PAGINATION_PAGES=100`, or pruned server-side — are dropped and never repopulated, understating `calculate_initial_investment_value` for very old safes. Low probability and a reasonable tradeoff; flag only. Consider logging a count of dropped-but-not-repopulated entries so the loss is observable.

OjusWiZard (commented):
Re-review of the fix commits (`89938537`+). C1, I1, and I3 are genuinely resolved — nice. One remaining issue in the I2 fix (N1, inline) plus two advisories. N1 is the only one I'd want addressed before calling this complete; it's a ~4-token change to the pre-pass.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e60a3d52. `_fetch_optimism_transfers_safeglobal` now returns a bool — `False` on any mid-pagination `success == False` or top-level exception. `_fetch_all_transfers_until_date_optimism` keeps the migrated shape in a local until the fetcher returns success; on failure it returns the previously-persisted raw shape and skips both the in-memory merge into `self.funding_events["optimism"]` and the `store_funding_events()` call. Disk state is untouched, so the next cycle re-migrates against the still-intact legacy entries and retries. New test: `TestFetchAllTransfersDateMerge::test_optimism_fetch_failure_does_not_persist_partial`.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e60a3d52, same shape as the incoming fetcher. `_fetch_outgoing_transfers_until_date_optimism` now keeps the migrated shape in a local `existing_outgoing` and the raw pre-migration shape in `raw_existing_outgoing`; the `self.funding_events["optimism_outgoing"]` write was removed from the top of the function. On any mid-pagination `success == False` (`fetch_failed` flag set) or the outer except, the function returns `raw_existing_outgoing` and never reaches the merge or `store_funding_events()`. Disk keeps the original legacy entries intact for next cycle to retry against. The existing `TestOutgoingOptimism::test_fail` now also asserts `store_funding_events.assert_not_called()` plus the un-stripped raw shape is returned.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e60a3d52. The pre-pass now narrows the eligibility filter to outgoing-USDC-ERC20 by also checking `tokenInfo.symbol.upper() == "USDC"` so it matches what the seen-set actually receives. When `tokenInfo` is absent on the API response we can't tell the symbol cheaply, so we count the entry as eligible — that errs on "don't stop early" and is safe because the seen-set will include it once the downstream USDC check passes anyway. The "early-stop after one page" claim now holds for mixed pages that include non-USDC outgoing ERC20s.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: The eligible-only pre-pass was actually narrowed to outgoing-USDC in the Lockhart C3 fix (e60a3d52) — the current pre-pass also filters on ``tokenInfo.symbol.upper() == "USDC"`` (or treats ``tokenInfo`` absence as eligible to err on "don't stop early"). What was missing was test coverage that locks in the behaviour for a mixed page containing a non-USDC outgoing ERC20 alongside a seen USDC withdrawal. Added that in 0a3e9b83: ``TestErc20OptimismFailureModes::test_eligible_only_early_stop_with_non_usdc_outgoing_erc20`` simulates ``(seen-USDC-out, unseen-AERO-out)`` on the same page and asserts the loop still stops after page 1.

Also took the minor — the pre-pass was already a simple counter (`eligible_in_page = 0; for ... eligible_in_page += 1`) after the Lockhart edit, so no `eligible_indices` list to clean up anymore.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 0a3e9b83. Applied the same eligible-only pre-pass to ``_fetch_outgoing_transfers_until_date_optimism``: eligibility there is outgoing ETHER_TRANSFER with ``value > 0``. The counter increments only on seen eligibles and the stop fires when ``seen_eligible_in_page == eligible_in_page``, matching the shape used in ``_track_erc20_transfers_optimism``. Replaces the previous ``consecutive_existing >= len(transfers)`` denominator that rarely fired against mixed ``/transfers/`` pages.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 0a3e9b83. ``_drop_legacy_transfer_entries`` now returns ``(migrated, dropped_count)``. Both call sites (``_fetch_all_transfers_until_date_optimism`` for the ``optimism`` key and ``_fetch_outgoing_transfers_until_date_optimism`` for ``optimism_outgoing``) log a single INFO line per cycle when ``dropped > 0``, including the count and a note that the entries will be re-fetched if still within Safe's pagination window. For very old safes some entries may be beyond that window or pruned server-side and therefore unrecoverable — the log line makes that loss observable. ``TestDropLegacyTransferEntries`` updated to assert on the new tuple shape.

LOCKhart07 (approved):
Re-reviewed at `0a3e9b83`. Both blocking findings are fixed correctly in `e60a3d52`: the legacy migration is now applied to a local var and the stripped shape is committed only after a fully-successful fetch — `_fetch_optimism_transfers_safeglobal` returns a bool, and both `_fetch_all_transfers_until_date_optimism` and `_fetch_outgoing_transfers_until_date_optimism` return the raw pre-migration data without storing on a mid-pagination failure, so disk history survives a transient blip and the next cycle retries. The non-USDC eligible-only gap (and Ojus N1/A1/A2) are addressed too: the ERC20 pre-pass now narrows eligibility to outgoing-USDC, A1 extends the eligible-only stop to the outgoing-ETH fetcher, and `_drop_legacy_transfer_entries` now reports a drop count so unrecoverable history loss is logged rather than silent.

Also checked: the failure-path return of the full cumulative dict does not double-count (`_calculate_chain_investment_value` recomputes from the priced cache rather than accumulating), and the two `@pytest.mark.skip`s correctly cover branches the Basius MVP bypasses, with re-enable pointers. New tests are falsifiable and pin each fix. CI green across all platforms / Python 3.10–3.14.

Superseding my earlier REQUEST_CHANGES — LGTM.

OjusWiZard (approved):
Approving. Both review rounds plus the second reviewer's blocking findings are addressed and verified against commits `e60a3d52` + `0a3e9b83`:

- **C1** transient-error cache poisoning, **I1** partial-pagination data loss, **I3** legacy double-count on upgrade — fixed (`None` sentinel + `fetch_failed` guards + `_drop_legacy_transfer_entries` migration).
- **N1** eligible-only early-stop narrowed to outgoing-USDC; **A1** same pre-pass applied to the outgoing-ETH fetcher; **A2** drop-count now logged.
- Migration + partial-persist guards are symmetric across the incoming and outgoing fetchers, so a mid-stream failure can no longer delete unrecoverable history.

Skipped tests (`test_not_profitable_fees`, `test_query_returns_none`) are legitimate — they cover TEMP-bypassed safety checks with clear WS2 2.6 / WS3 3.3 re-enable pointers. Out-of-scope TEMP bypasses are per the launch plan.

Clean from a correctness standpoint.

PR #329, opened 2026-06-02
fix(safe-api): TTL-based kv caches + proxy-safe pagination (WS1 1.3+1.4)

DIvyaNautiyal07:
## Summary

Workstream 1 PRs 1.3 and 1.4 from the Basius launch plan: https://docs.google.com/document/d/1R_ZuAV4wFSaiwXaQFo2TBXrtQMRCZAum8X9CzSWijiE/edit?tab=t.1xsq6yndq2gv

Branched off `hack/launch-base` after PR #328 merged.

### PR 1.3 — Shorten kv cache TTL from "today vs not today" to ~30 minutes

`calculate_withdrawals_value` and `calculate_initial_investment_value_from_funding_events` previously gated their recompute on `last_date == current_date`. With the same-day-skip dedup bug closed in PR #328, this calendar-day cache became the remaining barrier to same-day top-ups landing in ROI math: a 10 AM top-up + a 4 PM top-up on the same day would not appear in `calculate_initial_investment_value` until tomorrow's first cycle.

Replaced both day-equality checks with `0 <= (now - last_ts) < TTL` against new module-level constants `INITIAL_INVESTMENT_CACHE_TTL_SECONDS` and `WITHDRAWAL_CACHE_TTL_SECONDS`, both initially 30 minutes. Tunable in one place if we want fresher numbers (drop to 15) or more headroom (raise to 60). Negative ages (clock skew or malformed timestamps) are treated as expired so we don't serve a possibly-stale value.

### PR 1.4 — Proxy-safe pagination

Three Optimism transfer fetchers followed Safe's response `next` URL to advance pagination. That URL is absolute and points back at `safe-transaction-*.safe.global`, so under the planned proxy only page 1 would go through our infra — pages 2+ would bypass it with no API key. The proxy plan can't work end-to-end without this.

Replaced the next-URL chase with offset-based pagination built from `self.params.safe_api_v1_url`. Every page the agent issues now goes to the configured endpoint, whether that's the public Safe host today or the proxy URL after WS1 PR 1.5 lands. Termination still consults `response.get("next")` so the behaviour matches Safe's existing pagination contract. Deliberately doesn't gate on `len(transfers) < limit` because Safe is free to return short non-final pages.

Affected sites in `fetch_strategies.py`:
- `_fetch_optimism_transfers_safeglobal` (incoming-transfers)
- `_fetch_outgoing_transfers_until_date_optimism` (outgoing-ETH)
- `_track_erc20_transfers_optimism` (outgoing-USDC)

Added the existing `MAX_PAGINATION_PAGES = 100` as a safety backstop on all three loops so a misbehaving API can't trigger an unbounded walk.

## Per-cycle Safe API cost (Optimism, steady state)

After these fixes, on a 5-minute FSM cycle within a 30-minute TTL window:

| Cycle | balances | incoming | withdrawals | total |
|---|---|---|---|---|
| 1 (TTL miss, fresh window) | 1 | 1 | 1 | 3 |
| 2-5 (TTL hits) | 1 | 0 | 0 | 1 |
| 6 (TTL miss, new window) | 1 | 1 | 1 | 3 |

About 16 calls per hour per agent per chain. ~5x reduction vs no-cache, well inside Safe's authenticated free-tier ceiling. Same-day top-ups now appear in ROI within 30 minutes instead of waiting for the calendar day to roll.

## Tests

4 new tests:

- `TestProxySafePagination::test_withdrawal_fetcher_paginates_via_proxy_base_url` — page 1 advertises an absolute `next` pointing at Safe's host; asserts page 2 still goes through `safe_api_v1_url`.
- `TestProxySafePagination::test_pagination_respects_max_pagination_pages_cap` — exercises the backstop.
- `TestKvCacheTtlBoundaries::test_withdrawal_cache_negative_age_treated_as_expired` — future-dated `last_ts` (clock skew) recomputes rather than serving cached.
- `TestKvCacheTtlBoundaries::test_initial_investment_cache_just_inside_ttl_is_hit` — boundary check.

Updated 4 pre-existing tests in `TestCalculateInitialInvestmentValueFromFundingEvents` and `TestCalculateWithdrawalsValueSameDayCache` to use the new age semantics (renamed where the "same-day" framing no longer fit).

`pytest packages/valory/skills/liquidity_trader_abci/tests/` — 2730 pass, 2 skip (the WS2/WS3-bound safety-check tests inherited from launch-base).

Full lint suite (black-check, isort-check, flake8, mypy, pylint, darglint) all OK locally.

## Out of scope

- **PR 1.5** — per-chain Safe API URLs. Separate PR (needs production coordination on env-var names + proxy URL value).
- **Workstreams 2 and 3.**

## Test plan

- [ ] Run an Optimism agent against a real funded Safe for two FetchStrategies cycles ~5 minutes apart; confirm only 1 Safe API call fires on the second cycle (logs should show "Withdrawal cache hit Ns old" and equivalent for initial investment).
- [ ] Wait > 30 minutes, confirm cycle issues 3 Safe API calls again and rewrites both kv keys.
- [ ] Point `safe_api_v1_url` at a proxy host; confirm every paginated request in agent logs goes through the proxy (no calls to `safe-transaction-*.safe.global` even when Safe's response advertises an absolute `next` URL).


packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4175,11 +4222,14 @@ def _fetch_optimism_transfers_safeglobal(
...
-                if not cursor:
+                # Advance pagination by offset. Safe's DRF pagination signals
+                # "no more pages" via ``next: null``. Empty ``results`` is
+                # handled earlier in the loop. We deliberately don't gate on
+                # ``len(transfers) < SAFE_TRANSFERS_PAGE_LIMIT`` because Safe
+                # is free to return short non-final pages.
+                if response_json.get("next") is None:
                     break
-                transfers_url = cursor
+                offset += SAFE_TRANSFERS_PAGE_LIMIT
```
LOCKhart07: **Blocking — offset advance skips records on a short non-final page.**

This advances by a fixed `SAFE_TRANSFERS_PAGE_LIMIT` (100), but the number of records actually consumed is `len(transfers)`, which can be smaller. The termination comment just above explicitly assumes Safe may return short non-final pages ("we deliberately don't gate on `len(transfers) < SAFE_TRANSFERS_PAGE_LIMIT` because Safe is free to return short non-final pages"), and Safe also clamps `limit` to a server-side max. In either case `next` is non-null, the loop continues, and `offset += 100` jumps past the records that weren't returned — silently dropping funding/withdrawal transfers and corrupting ROI/withdrawal totals. The old next-URL chase couldn't hit this because Safe computed the offset itself.

```python
offset += len(transfers)
```
is strictly safer (dedup via `seen_transfer_ids` absorbs any overlap) and makes the offset logic consistent with the termination logic you already chose. Same fix needed at the two sibling fetchers (lines 4902 and 5142).

**Test gap (same root):** `test_*_paginates_via_proxy_base_url` build page 1 as a *single* transfer with a non-null `next` — exactly the short-page trigger — but only assert the host. They never assert the offset value, so they'd pass even though the code skips to `offset=100` after consuming 1 record. Add an assertion that page 2's offset equals the count consumed.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4800,11 +4892,14 @@ def _fetch_outgoing_transfers_until_date_optimism(
...
```
LOCKhart07: Same bug as the incoming fetcher (see comment on line 4232): advance by `offset += len(transfers)`, not a fixed `SAFE_TRANSFERS_PAGE_LIMIT`, or a short non-final page (which the termination comment above assumes is possible) silently skips records.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -5018,10 +5132,14 @@ def _track_erc20_transfers_optimism(
...
```
LOCKhart07: Same bug as the incoming fetcher (see comment on line 4232): advance by `offset += len(transfers)`, not a fixed `SAFE_TRANSFERS_PAGE_LIMIT`, or a short non-final page (which the termination comment above assumes is possible) silently skips records.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e7b68e73. Changed all three fetchers to `offset += len(transfers)`. Dedup absorbs any overlap on full pages; short non-final pages no longer skip records.

Test gap addressed: all three `test_*_paginates_via_proxy_base_url` tests now also assert page 2's offset equals the count consumed (1 in those scenarios), and added `test_short_non_final_page_does_not_skip_records` which builds a 50-row page-1 with `next != null` and pins `offset=50` on page 2. Mutation back to `offset += SAFE_TRANSFERS_PAGE_LIMIT` would fail that test.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e7b68e73. Changed all three fetchers to `offset += len(transfers)`. Dedup absorbs any overlap on full pages; short non-final pages no longer skip records.

Test gap addressed: all three `test_*_paginates_via_proxy_base_url` tests now also assert page 2's offset equals the count consumed (1 in those scenarios), and added `test_short_non_final_page_does_not_skip_records` which builds a 50-row page-1 with `next != null` and pins `offset=50` on page 2. Mutation back to `offset += SAFE_TRANSFERS_PAGE_LIMIT` would fail that test.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e7b68e73. Changed all three fetchers to `offset += len(transfers)`. Dedup absorbs any overlap on full pages; short non-final pages no longer skip records.

Test gap addressed: all three `test_*_paginates_via_proxy_base_url` tests now also assert page 2's offset equals the count consumed (1 in those scenarios), and added `test_short_non_final_page_does_not_skip_records` which builds a 50-row page-1 with `next != null` and pins `offset=50` on page 2. Mutation back to `offset += SAFE_TRANSFERS_PAGE_LIMIT` would fail that test.

LOCKhart07 (approved):
Approve. Reviewed at PR head `e7b68e73` (read the three fetchers and both cache paths in final state, not just the diff).

### PR 1.4 — offset pagination (my earlier blocking concern)
- All three fetchers now advance with `offset += len(transfers)`, not `+= SAFE_TRANSFERS_PAGE_LIMIT` (`fetch_strategies.py:4235`, `:4908`, `:5151`). Correctly handles a short non-final page (Safe returning `< limit` rows with `next != null`) without skipping un-returned records. `test_short_non_final_page_does_not_skip_records` pins it and is a real mutation-killer.
- Every page is built from `self.params.safe_api_v1_url`, so pages 2+ no longer leak to `safe-transaction-*.safe.global`. All three proxy-URL tests assert both the host and the `offset=1` increment.
- Termination uses `response_json.get("next") is None` — more precise than the old `if not cursor`, and consistent across all three loops.
- `MAX_PAGINATION_PAGES` backstop sets `fetch_failed = True` and skips persistence in all three (`:4245` returns `not fetch_failed`; `:4910` returns stale; `:5153` returns `None`). Cap-as-failure is the right call — persisting partial history would seed next cycle’s seen-set falsely-complete and permanently drop older pages.

### PR 1.3 — TTL caches
- Both paths use `0 <= age_seconds < TTL`; negative ages (clock skew) and unparseable timestamps are treated as expired and warned. Read and write both use `_get_current_timestamp()` (epoch seconds), so the comparison is timezone-safe — an improvement over the old `utcfromtimestamp` vs local-`now()` mix.
- The `else`-wrap is a real latent bug fix: the old code set `fetch_till_date = True` on a falsy same-day hit then unconditionally overwrote it back to `False` (plus a misleading "not today" log). `test_initial_investment_ttl_hit_falsy_passes_full_history_flag_to_mode` exercises the consuming (Mode) chain and pins `fetch_till_date is True`.

Tests are falsifiable, stale "same-day" wording cleaned up. `py_compile` clean.

### Non-blocking observations
1. First fetch of a Safe with >100 pages (~>10k transfers) hits the cap → `fetch_failed` → nothing persisted → next cycle restarts at `offset=0` with no forward progress. Realistically agent Safes never reach that volume, and failing beats persisting a falsely-complete seen-set — acceptable tradeoff, worth a louder alert if high-volume Safes are ever expected.
2. Offset pagination is not reorg-stable (a transfer removed between pages could skip a record), but Safe’s `next` is itself DRF offset-based, so no regression vs the old next-following; new arrivals are absorbed by the `seen_*_ids` dedup.

Did not run the suite locally (needs `autonomy packages sync` against `launch-base`); PR reports 2738 pass / lint clean and my static trace agrees.

OjusWiZard (commented):
Multi-agent review (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer). No blocking correctness bugs — TTL math, negative-age handling, offset arithmetic, and the proxy-URL guarantee are all correct. Inline notes below are test-coverage gaps and clarity/observability polish.

One flagged "data-loss" candidate was **rejected**: `offset += SAFE_TRANSFERS_PAGE_LIMIT` matches DRF `LimitOffsetPagination.get_next_link()` (which advances by the requested `limit`, with `next: null` exactly when `offset + limit >= count`), so it is behaviourally identical to the old next-URL following — no divergence, no dropped transfers.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3141,32 +3160,38 @@ def calculate_initial_investment_value_from_funding_events(
...
                         )
-                        last_date = "1970-01-01"
+                        age_seconds = INITIAL_INVESTMENT_CACHE_TTL_SECONDS
 
-                    # If last calculation was today, return cached value
-                    if last_date == current_date:
+                    # TTL-based cache: return cached value if the last
+                    # calculation was within INITIAL_INVESTMENT_CACHE_TTL_SECONDS.
+                    # Negative ages (clock skew) treated as expired.
+                    if 0 <= age_seconds < INITIAL_INVESTMENT_CACHE_TTL_SECONDS:
```
OjusWiZard, line 3198: **Test gap (Important).** The withdrawal path has `test_withdrawal_cache_just_outside_ttl_is_miss` pinning the strict `<` at the exact boundary (age == TTL → miss). The initial-investment path has the just-inside-TTL *hit* test but no exactly-at-TTL *miss* test. A `<`→`<=` regression on this line would go undetected.

Suggest adding `test_initial_investment_cache_exactly_at_ttl_is_miss`: `last_ts = now - INITIAL_INVESTMENT_CACHE_TTL_SECONDS`, assert `_load_chain_total_investment` is **not** called and the recompute path runs.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2431,20 +2443,26 @@ def calculate_withdrawals_value(self) -> Generator[None, None, Decimal]:
...
-                    last_date = datetime.utcfromtimestamp(int(ts)).strftime("%Y-%m-%d")
+                    age_seconds = now_ts - int(ts)
                 except (ValueError, TypeError):
-                    last_date = "1970-01-01"
-                if last_date == current_date:
+                    self.context.logger.warning(
+                        f"Unparseable last_withdrawals_calculated_timestamp "
+                        f"{ts!r}; treating cache as expired"
+                    )
+                    age_seconds = WITHDRAWAL_CACHE_TTL_SECONDS  # treat as expired
```
OjusWiZard: Two notes on this line:

**1. Test gap (Important).** This `except (ValueError, TypeError)` branch (set age=TTL → expired) is untested for `calculate_withdrawals_value`. The initial-investment path has `test_invalid_timestamp_format`, but the withdrawal path doesn't. A regression flipping the fallback to `age_seconds = 0` (spurious cache hit) wouldn't be caught. Suggest a test with `last_withdrawals_calculated_timestamp="not-a-number"` asserting the fetcher runs + kv is rewritten.

**2. Clarity (Advisory).** `age_seconds = WITHDRAWAL_CACHE_TTL_SECONDS` is correct *only* under the strict `< TTL` guard below. A future `<`→`<=` change would silently turn every parse-error into a cache **hit**. `age_seconds = WITHDRAWAL_CACHE_TTL_SECONDS + 1` (or `-1`, which the docstring already documents as "expired") makes the intent guard-independent. Same applies to the initial-investment sentinel at line 3171.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2454,7 +2472,8 @@ def calculate_withdrawals_value(self) -> Generator[None, None, Decimal]:
                     if cached_val is not None:
                         try:
                             self.context.logger.info(
-                                "Last withdrawal calculation was today, using cached value"
+                                f"Withdrawal cache hit ({age_seconds}s old, "
```
OjusWiZard, line 2493: **Test gap (Important).** Every TTL-hit test seeds a valid `"optimism_total_withdrawals": "12.5"`. The two fall-through paths inside this within-TTL branch are untested:
- `cached_val is None` (second `_read_kv` returns no value) → should fall through to recompute
- cached value unparseable (`InvalidOperation/ValueError/TypeError`) → should fall through to recompute

A regression that returned `Decimal("0")` instead of recomputing on a missing/garbage cached value would slip through. Suggest two small tests driving each fall-through.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -9809,3 +9793,620 @@ def test_applied_when_loading_optimism_outgoing(self):
...
+        obj._request_with_retries = req
+        obj.context.coingecko = MagicMock()
+        obj.params.sleep_time = 1
+
+        _drive(obj._track_erc20_transfers_optimism("0xAddr", 1704067200))
+
+        # Two API calls (page 1 + page 2); both go to the proxy base URL.
+        assert call_count[0] == 2
+        assert all(
+            url.startswith("https://safe-proxy.example.com/api/v1") for url in captured
```
OjusWiZard, line 9868: **Test hardening (Important, cheap).** All three proxy-pagination tests assert only the host prefix and absence of `safe.global`. They never assert the page-2 **offset**, so a typo like `offset += 1` (instead of `offset += SAFE_TRANSFERS_PAGE_LIMIT`) would still pass. Add `assert "offset=100" in captured[1]` (and `"limit=100"`) to actually pin the offset arithmetic these tests are meant to protect.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4672,15 +4748,33 @@ def _fetch_outgoing_transfers_until_date_optimism(
...
+            offset = 0
+            pages_processed = 0
             # Seed in-cycle dedup from already-persisted transfers so a new
             # transfer on an already-stored date is not silently dropped.
             seen_tx_ids = _collect_seen_transfer_ids(existing_outgoing)
             while True:
+                if pages_processed >= MAX_PAGINATION_PAGES:
+                    self.context.logger.warning(
+                        f"Hit MAX_PAGINATION_PAGES cap ({MAX_PAGINATION_PAGES}) for "
+                        "Optimism outgoing-transfers; treating as a failed fetch "
```
OjusWiZard, line 4789: **Observability (Advisory).** On cap-hit this fetcher returns stale `raw_existing_outgoing`. The caller `_track_eth_transfers_and_reversions` then computes `total_count` from that stale data, which equals the persisted `optimism_reversion_transfer_count`, so it serves the cached reversion **indefinitely** — the count can never grow past the cap to invalidate it. This only triggers on Safes exceeding 100 pages (~10k transfers) and is an inherent consequence of the (documented) "return stale on failure" contract, so not a blocker. But consider `logger.error` rather than `warning` on cap-hit so the stuck state is alarm-visible instead of silent.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -96,6 +95,19 @@
 # non-empty next-page cursor indefinitely.
 MAX_PAGINATION_PAGES = 100
 
+# Page size for offset-based pagination against Safe Transaction Service.
+# Matches Safe's default limit; the API caps at 100 anyway.
+SAFE_TRANSFERS_PAGE_LIMIT = 100
+
+# Time-to-live for the kv-side caches around the expensive ROI math. Set to
+# 30 minutes so a same-day top-up shows up on the next half-hour boundary
```
OjusWiZard: **Comment accuracy (Advisory)** — three small nits:
- "next **half-hour boundary**" implies clock-aligned :00/:30 slots; the code is a sliding 30-min window from the last compute. And "rather than waiting for the full TTL to expire" is self-referential (30 min *is* the TTL). Suggest: "each computation stays valid for up to 30 minutes after it ran."
- Line 99 `"the API caps at 100 anyway"` embeds an external fact the code doesn't depend on (Safe has used 200 in some versions) — rot risk; the `next == null` stop condition works regardless of the cap.
- The module-level `MAX_PAGINATION_PAGES` comment (lines 93–95) calls it a soft "backstop." True for the Mode loops, but the three Safe loops treat cap-hit as a **hard failure** (set `fetch_failed`, return None/stale). Worth clarifying the two roles.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in 41f9571e. Added `test_initial_investment_cache_exactly_at_ttl_is_miss` symmetric with the existing withdrawal version. Pins the strict `< TTL` boundary on the initial-investment path.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Both addressed in 41f9571e.

**1.** Added `test_withdrawal_cache_invalid_timestamp_format` — seeds a non-numeric `last_withdrawals_calculated_timestamp` and asserts the fetcher runs + kv is rewritten with a parseable value. A regression flipping the parse-error fallback to a spurious cache hit (e.g. `age_seconds = 0`) would now fail.

**2.** Changed both parse-error fallbacks (this site and the initial-investment site at line 3179) to `age_seconds = -1`. Negative age is already documented as expired by the guard, so a future `<`→`<=` regression on the boundary can't flip the fallback into a spurious cache hit. The intent is now guard-independent.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Both fall-through paths covered in 41f9571e:

- `test_withdrawal_cache_within_ttl_missing_cached_val_falls_through` — second `_read_kv` returns no `optimism_total_withdrawals`, asserts the fetcher runs (not `None` / not `Decimal("0")`).
- `test_withdrawal_cache_within_ttl_unparseable_cached_val_falls_through` — seeds `"garbage"` so `Decimal(str(cached_val))` raises `InvalidOperation`, asserts the except branch falls through to the fetcher rather than serving 0 or propagating.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
DIvyaNautiyal07: Note that the offset-arithmetic guarantee is also pinned by:
- Existing `assert "offset=0" in captured[0]` / `assert "offset=1" in captured[1]` added in e7b68e73 (a few lines below this comment).
- `test_short_non_final_page_does_not_skip_records` which builds a 50-row page-1 with `next != null` and pins `offset=50` on page 2.

The combination catches both the old `offset += SAFE_TRANSFERS_PAGE_LIMIT` and the `offset += 1` typo classes. The asserted page-2 value is `offset=1` rather than `offset=100` because pagination now advances by `len(transfers)` (Lockhart's fix in e7b68e73), so page-2 offset == count consumed.

Addressed the second half by adding `assert "limit=100" in captured[0] and "limit=100" in captured[1]` on all three proxy-pagination tests in 41f9571e — a `SAFE_TRANSFERS_PAGE_LIMIT` change would now be caught.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Done in 41f9571e — escalated this site to `logger.error`. Comment in the code spells out why this fetcher is special: it returns stale `raw_existing_outgoing` on failure (by design — the caller's reversion-cache key is derived from the transfer count, so serving stale keeps the cache valid through transient blips). On a Safe that exceeds the cap that stale count is permanently pinned, so the stuck state needs to be alarm-visible rather than buried in warning noise.

Kept the two sibling fetchers at `warning` — they signal failure to the caller via `None` / `False`, so the caller drops the data instead of getting stuck on it.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: All three nits addressed in 41f9571e:

- TTL constants block rewritten to drop the misleading "next half-hour boundary" (it's a sliding window from the last compute, not clock-aligned) and the self-referential "rather than waiting for the full TTL to expire". Now reads "each computation stays valid for up to 30 minutes after it ran."
- `SAFE_TRANSFERS_PAGE_LIMIT` comment dropped "the API caps at 100 anyway". Loop termination is gated on `next == null` and is robust to any server-side cap change.
- `MAX_PAGINATION_PAGES` comment now spells out the two roles: hard-failure trigger for the three Safe fetchers (sets `fetch_failed`, persistence is skipped), soft backstop for the Mode loops.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -9809,3 +9793,807 @@ def test_applied_when_loading_optimism_outgoing(self):
...
+                "0xAddr", "2099-01-01", all_transfers_by_date, existing
+            )
+        )
+
+        assert call_count[0] == 2
+        assert all(
+            url.startswith("https://safe-proxy.example.com/api/v1") for url in captured
+        ), captured
+        assert all("safe.global" not in url for url in captured), captured
+        # Offset advances by ``len(transfers)`` (1), not by ``limit`` (100).
```
DhairyaPatel7, line 9969: **This assertion doesn't actually guard the offset regression for the incoming fetcher.**

Page 1 returns 1 transfer, so the correct offset is `1` and the bug (`offset += SAFE_TRANSFERS_PAGE_LIMIT`) would produce `100`. But `"offset=1"` is a substring of `"offset=100"`, so `"offset=1" in captured[1]` is `True` even when the code regresses:

```python
>>> "offset=1" in "...&offset=100"
True
```

So a revert to `offset += SAFE_TRANSFERS_PAGE_LIMIT` in `_fetch_optimism_transfers_safeglobal` would pass this test. The only test that truly pins the increment is `test_short_non_final_page_does_not_skip_records` (offset=50, not a substring of offset=100), and it only drives `_track_erc20_transfers_optimism`. So the exact bug from the blocking review is regression-guarded for one of the three fetchers, not all three.

Fix: `assert captured[1].endswith("&offset=1")`, or add a short-page (offset=50) regression test for this fetcher. Same applies to the outgoing-ETH test below.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -9809,3 +9793,807 @@ def test_applied_when_loading_optimism_outgoing(self):
...
+        obj.params.sleep_time = 1
+
+        _drive(
+            obj._fetch_outgoing_transfers_until_date_optimism("0xaddr", "2099-01-01")
+        )
+
+        assert call_count[0] == 2
+        assert all(
+            url.startswith("https://safe-proxy.example.com/api/v1") for url in captured
+        ), captured
```
DhairyaPatel7, line 10013: Same weak-substring issue as the incoming-fetcher test above (line 9966): `"offset=1"` is a substring of `"offset=100"`, so this passes even if `_fetch_outgoing_transfers_until_date_optimism` regresses to `offset += SAFE_TRANSFERS_PAGE_LIMIT`. Use `.endswith("&offset=1")` or a short-page (offset=50) regression test.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4016,14 +4053,35 @@ def _fetch_optimism_transfers_safeglobal(
...
+            # following Safe's response ``next`` URL: that URL is absolute
+            # and points at safe-transaction-*.safe.global, so under a proxy
+            # only page 1 would go through our infra — page 2 onwards would
+            # bypass it. Building the URL ourselves from safe_api_v1_url
+            # keeps every page going through the configured endpoint.
+            transfers_base_url = f"{base_url}/safes/{address}/incoming-transfers/"
 
             processed_count = 0
+            offset = 0
+            pages_processed = 0
```
DhairyaPatel7, line 4076: **Non-blocking, low-probability:** this newly bounds a loop that was unbounded on `hack/launch-base` (the pre-existing cap only ever applied to the Mode fetchers, and there a cap-hit just `break`s and persists what it has). Here a cap-hit sets `fetch_failed=True`, so the caller skips persistence — and because the next cycle restarts from `offset=0` and hits the same cap, a Safe whose `/transfers/` history exceeds ~100 pages (~10k records) makes **zero incremental progress** and never syncs (before this PR it would, slowly).

The `fetch_failed` choice is the right one vs. the Mode "persist partial" approach (which has the falsely-complete-seen-set bug you're avoiding), so this is a real tradeoff, not necessarily a change to make here — just flagging the no-recovery path. If large Optimism Safes are ever in scope, consider raising the cap or an oldest-first backfill. Applies to all three fetchers.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4175,11 +4234,17 @@ def _fetch_optimism_transfers_safeglobal(
...
+                # return a short non-final page (server-side clamp or partial
+                # filter) with ``next != null``; advancing by ``limit`` would
+                # skip the un-returned records. Dedup via ``seen_transfer_ids``
+                # absorbs any overlap when Safe in fact returned a full page.
+                # Termination still consults ``next is None``; empty results
+                # are handled earlier in the loop.
+                if response_json.get("next") is None:
                     break
-                transfers_url = cursor
+                offset += len(transfers)
```
OjusWiZard, line 4250: **Test gap (Important).** `test_short_non_final_page_does_not_skip_records` proves the `offset += len(transfers)` advance only for `_track_erc20_transfers_optimism`. This **incoming** fetcher (`_fetch_optimism_transfers_safeglobal`) is exercised in `test_incoming_fetcher_paginates_via_proxy_base_url` with a *single* transfer on page 1, where `offset=1` coincidentally equals `len`. A regression here to `offset += SAFE_TRANSFERS_PAGE_LIMIT` would not be caught. Suggest a short-page test adapted to this fetcher: 50 transfers on page 1 with `next != null`, assert the 2nd request uses `offset=50` (not `offset=100`).

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4800,11 +4914,17 @@ def _fetch_outgoing_transfers_until_date_optimism(
...
+                # return a short non-final page (server-side clamp or partial
+                # filter) with ``next != null``; advancing by ``limit`` would
+                # skip the un-returned records. Dedup via ``seen_tx_ids``
+                # absorbs any overlap when Safe in fact returned a full page.
+                # Termination still consults ``next is None``; empty results
+                # are handled earlier in the loop.
+                if response_json.get("next") is None:
                     break
-                transfers_url = cursor
+                offset += len(transfers)
```
OjusWiZard, line 4919: **Test gap (Important).** Same as the incoming fetcher: the `offset += len(transfers)` advance for this **outgoing-ETH** fetcher (`_fetch_outgoing_transfers_until_date_optimism`) has no short-non-final-page test. `test_outgoing_eth_fetcher_paginates_via_proxy_base_url` uses a single transfer (`offset=1 == len`), so an `offset += SAFE_TRANSFERS_PAGE_LIMIT` regression here would slip through. Add a 50-row short-page test asserting `offset=50` on the 2nd request.

**Advisory (pre-existing, surfaced by this PR).** The `"Completed Optimism outgoing transfers: {processed_count} found"` log (line 4908) sits *inside* the `while` loop, so it fires on every page — same for the ERC20 fetcher (line 5151). The incoming fetcher logs its `"Completed"` line once *after* the loop (line 4249). Harmless (`processed_count` is monotonic) but misleading log noise now that multi-page fetches are realistic. Consider dedenting both to fire once after the loop, matching the incoming fetcher.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4175,11 +4234,17 @@ def _fetch_optimism_transfers_safeglobal(
                     )
                     break
 
-                # Check for next page and advance the URL
-                cursor = response_json.get("next")
-                if not cursor:
+                # Advance pagination by the number of records actually
```
OjusWiZard: **Advisory (cleanup).** This 8-line rationale comment is duplicated verbatim at three sites (here, line 4917, line 5160), varying only the `seen_*` variable name. It explains a genuinely non-obvious choice (advance-by-len vs advance-by-limit; why dedup absorbs overlap), so it's worth keeping — but once, near the `SAFE_TRANSFERS_PAGE_LIMIT` constant. Reduce the inline copies to a one-liner like `# advance by len(transfers); see SAFE_TRANSFERS_PAGE_LIMIT note`. Net-positive, no behavior change.

OjusWiZard (commented):
Re-review against current HEAD (41f9571e). All prior findings are resolved — `offset += len(transfers)`, the `age_seconds = -1` sentinel, `logger.error` on the outgoing cap-hit, the three comment fixes, and the new TTL fall-through/boundary tests all check out. No remaining Critical/Important correctness bugs. The fetchers' asymmetric failure-signalling (`bool` / stale-dict / `None`) is a documented, accepted trade-off — not flagged. Remaining items below are one test gap + two advisories.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
DIvyaNautiyal07: Good catch, fixed in 5b417302. Anchored all four offset checks with `endswith` (`&offset=0` / `&offset=1` / `&offset=50`) so the substring mismatch is caught directly. Also added `test_short_non_final_page_incoming_fetcher_does_not_skip_records` for this fetcher specifically — page 1 returns 50 rows with `next != null`, page 2 must hit `offset=50`. A revert to `offset += SAFE_TRANSFERS_PAGE_LIMIT` would now fail a dedicated test rather than slip through.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
DIvyaNautiyal07: Same fix in 5b417302 — `endswith("&offset=1")` for the proxy-base-url test and a new `test_short_non_final_page_outgoing_eth_fetcher_does_not_skip_records` (page 1 returns 50 ETHER_TRANSFER rows, asserts page-2 `offset=50`).

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Agreed on the tradeoff and deferring this one.

The cap-as-hard-failure is right for Basius scope. Today's funded Optimism Safe (`0x5a4B...86b2`) has ~3 pages of transfers; the new Base Safes for Basius launch will start from zero history. So the >10k scenario doesn't bite us yet.

If/when we onboard a Safe that crosses the cap, the right fix is oldest-first incremental backfill (persist with an `incomplete=true` marker that gets cleared once we walk back to the page that overlaps the previously persisted seen-set). That preserves the no-corruption invariant of this PR while making progress on each cycle. Filing as a follow-up rather than expanding this PR's scope. Thanks for spelling out the failure mode.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Already addressed in 5b417302 — `test_short_non_final_page_incoming_fetcher_does_not_skip_records` in `TestProxySafePagination` builds 50 incoming transfers on page 1 with `next != null` and asserts page 2 ends with `&offset=50`. A revert to `offset += SAFE_TRANSFERS_PAGE_LIMIT` would fail that test. Reviewed in parallel with my push, fully agreed it was a gap.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: **Test gap** — already addressed in 5b417302: `test_short_non_final_page_outgoing_eth_fetcher_does_not_skip_records` (50 ETHER_TRANSFER rows on page 1, asserts page-2 `&offset=50`). Same shape as the incoming version added in the same commit.

**Log noise** — fixed in fddccfc2. Dedented the per-page "Completed Optimism outgoing transfers" log to fire once after the loop, matching the incoming fetcher. Same for the "Completed Optimism ERC20 transfers" log at line 5151. Pre-existing as you noted; surfaced now that multi-page fetches are realistic.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Done in fddccfc2. Moved the full rationale to a block comment next to `SAFE_TRANSFERS_PAGE_LIMIT` at module scope (the genuinely non-obvious part: short non-final pages, dedup absorbs overlap, termination on `next is None`). Each inline site is now a single line: `# Advance by len(transfers); see SAFE_TRANSFERS_PAGE_LIMIT note.` Net 18 lines removed from the three loops.

OjusWiZard (approved):
Approving. All review findings — the blocking \`offset += len(transfers)\` short-page fix, TTL/cache fall-through tests, \`logger.error\` on cap-hit, and the proxy-safe pagination — are addressed. The two follow-up commits (\`5b417302\`, \`fddccfc2\`) closed the remaining test-coverage gaps (incoming + outgoing-ETH short-page regression tests with \`endswith\` offset anchoring) and the polish items (per-page log dedup, consolidated pagination rationale). No outstanding Critical/Important items. LGTM.

PR #331, opened 2026-06-02
feat(roi): freeze legacy ETH basis & remove ETH-reversion logic (WS2 2.0–2.3)

DhairyaPatel7:
## What & why

Workstream 2, tasks **2.0–2.6** in a single PR (the stack landed onto this branch via #332 and #333; what you're reviewing is the merged set, not just 2.0–2.3). Removes the Optimus ETH-reversion accounting that Basius doesn't need, replaces it with a one-time **frozen ETH basis** so ROI is unchanged for existing agents — including those that actually reverted ETH — and pulls in the opportunity-cache rollover invalidation, the slippage / fee / gas safety gate, and the re-enabled profitability gate.

```
initial_investment = frozen_eth_basis + Σ(live USDC)
```

`{chain}_total_investment` already equals `USDC + (incoming ETH − reverted ETH)`, so the freeze snapshots the correct netted value. It works whether or not an agent ever reverted, and is `$0` for USDC-only / Base agents (no special-casing).

## Changes by sub-task

**2.0 — frozen ETH basis.** `_freeze_eth_basis_if_needed` / `_load_frozen_eth_basis` / `_sum_usdc_investment_from_events`. Snapshots the net ETH basis into `funding_events.json` (idempotent; no-op when no ETH was ever received). Drops the dead `optimism_reversion_info` / `optimism_reversion_transfer_count` / `optimism_outgoing` keys.

**2.1 — drop revert-transaction generation.** `_check_and_create_eth_revert_transactions` + its period-0 caller removed.

**2.2 — chain investment uses frozen basis.** `_calculate_chain_investment_value` → `frozen_eth_basis + Σ USDC` (ETH skipped, no reversion subtraction).

**2.3 — drop reversion fetch/cache layer.** For optimism and mode, plus the `ETHER_TRANSFER` recording branch, and the now-unused imports.

**2.4 — opportunity-cache rollover invalidation** (from #332). `evaluate_strategy.py` now invalidates a cached opportunity that survived a period rollover (a successful entry already invalidates it, so a surviving cache means the entry failed). `_invalidate_cl_pool_cache` (in `base.py`) returns a bool, and the rollover and post-entry call sites in `evaluate_strategy.py` / `decision_making.py` log on `False`.

**2.5 — combined value-loss / slippage cap** (from #333). New `max_slippage_percentage` param in `models.py`. `check_if_route_is_profitable` rejects routes whose `(fromAmountUSD − toAmountUSD) / fromAmountUSD` exceeds the cap, on top of the existing fee/gas gates. Defaults are tuned in `service.yaml` and `agents/optimus/aea-config.yaml`.

**2.6 — re-enable profitability gate** (from #333). `check_if_route_is_profitable` is once again called on the route-selection path; the deliberately-disabled `test_not_profitable_fees` is replaced with the full new test family.

## Review fixes folded in

- **Direct-read investment path** (`address Ojuswi`, `address Divya Self-review`): leg-namespaced `priced_transfers` keys, narrowed `except` with traceback, warning-on-empty-funding-events, `_invalidate_cl_pool_cache` returns bool, plus the symmetric `(True, False)` save-guard test and at-cap boundary tests for the fee and gas gates.
- **Soft-KV bool fix** (`address LOCKhart07`): `_invalidate_cl_pool_cache` now captures `_write_kv`'s return value (it returns `False` on soft failure without raising) so the new `False`-handling at both call sites isn't dead.
- **Drop the post-loop cross-chain overwrite** (`address LOCKhart07`): the trailing `_save_chain_total_investment(chain, total_investment)` after the chain loop was writing the cross-chain sum to the last-iterated chain's key; per-chain saves inside `_calculate_chain_investment_value` already cover it.

## Validation

- Cross-checked against real Optimism agent safes via the Safe API, including a genuinely reverted agent (`0xE4EA…50A0`, reverted 0.005 ETH on 2025-06-20): the freeze reproduces today's ROI denominator.
- Full skill suite at **2733 tests passing, 1 skipped** at the current head.

## Before merge

- [x] `make format` and `autonomy packages lock` re-run after each review round.
- ℹ️ `test_decision_making.py::TestCheckIfRouteIsProfitable::test_not_profitable_fees` no longer fails — re-enabled along with the profitability gate (see 2.6 above).


DIvyaNautiyal07:
Quick question on the freeze before I go deeper. Why do we need to snapshot the ETH basis instead of computing it on the fly from `funding_events.json`?

The reverted ETH events are already persisted under `optimism_outgoing` and the incoming ETH under `optimism`, so `initial = sum(USDC_in) + sum(ETH_in priced) − sum(ETH_out priced)` reduces to a read from disk. No new state, no migration step, no dependency on whatever `_load_chain_total_investment` returns from kv. That path also avoids the asymmetric USDC-only filter in `_sum_usdc_investment_from_events` and survives a wiped kv.

Is there a case where `funding_events.json` isn't trustworthy as the source for legacy Optimism agents, or am I missing why the snapshot is needed?

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4378,457 +4235,117 @@ def _save_chain_total_investment(
...
+                f"(total ${current_total} - USDC ${usdc_total})"
             )
-            return raw_existing_outgoing
 
-        try:
-            # Use SafeGlobal API for Optimism transfers
-            base_url = self.params.safe_api_v1_url
-            transfers_url = f"{base_url}/safes/{address}/transfers/"
+        self.funding_events["frozen_eth_basis"] = frozen
+        self.funding_events["eth_basis_frozen"] = True
```
OjusWiZard: **Important — new agent permanently locks `eth_basis = 0`.** On a fresh agent's first run `current_total = 0` and `usdc_total = 0`, so the basis is frozen at 0 and `eth_basis_frozen = True` is persisted *before any funding arrives*. Since live ETH is hard-skipped (line ~3185) and the basis is now locked, any ETH deposited after this first run is never counted in the ROI denominator, permanently. Intentional for the existing-agent migration (USDC-only going forward), but currently undocumented and unguarded. Suggest logging a WARNING when `eth_basis == 0`, and documenting the "ETH after freeze is never counted" contract.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4378,457 +4235,117 @@ def _save_chain_total_investment(
...
 
-        if not address:
-            self.context.logger.warning(
-                "No address provided for fetching Optimism outgoing transfers"
+        frozen: Dict[str, float] = {}
+        for chain in self.params.target_investment_chains:
+            current_total = yield from self._load_chain_total_investment(chain)
+            usdc_total = yield from self._sum_usdc_investment_from_events(chain)
+            eth_basis = current_total - usdc_total
+            if eth_basis < 0:
```
OjusWiZard: **Important — negative `eth_basis` clamped silently.** `eth_basis < 0` means Σ(USDC) exceeds the stored total — a real accounting inconsistency (stale `current_total`, or a USDC key mismatch). It's clamped to `0.0` and the only log is the post-clamp INFO line showing `$0.0`, so the signal is lost and the wrong value is then frozen forever. Suggest a WARNING *before* clamping that includes the raw negative value, `current_total`, and `usdc_total`.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4378,457 +4235,117 @@ def _save_chain_total_investment(
...
+                if (
+                    chain == "mode"
+                    and self.params.airdrop_started
+                    and self.params.airdrop_contract_address
+                    and str(transfer.get("from_address", "")).lower()
+                    == self.params.airdrop_contract_address.lower()
+                ):
+                    continue
+                key = self._get_transfer_key(date, transfer, idx)
+                total += priced_transfers.get(key, 0.0)
```
OjusWiZard: **Important — freeze before USDC is fully priced overstates the ETH basis (one-shot, irreversible).** A USDC transfer present in `funding_events` but not yet in `priced_transfers` (e.g. a prior price-fetch failure) contributes `0.0` here, shrinking `usdc_total` and inflating the frozen `eth_basis`. For existing agents the prior run usually priced everything so the window is narrow — but the freeze is one-time and the error is permanent. Suggest logging a WARNING when a positive-amount USDC transfer has no priced entry, and/or deferring the freeze until pricing is complete.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4378,457 +4235,117 @@ def _save_chain_total_investment(
...
-                    )
-
-            if len(reversion_transfers) > 0:
-                historical_reversion_value = (
-                    yield from self._calculate_total_reversion_value(
-                        eth_transfers, reversion_transfers
-                    )
-                )
+            return float(basis.get(chain, 0.0) or 0.0)
+        except (TypeError, ValueError):
```
OjusWiZard: **Advisory — silent swallow.** A corrupt stored basis (non-numeric) is silently turned into `0.0`, understating investment with no trace. Add a WARNING log in the `except` block.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4378,457 +4235,117 @@ def _save_chain_total_investment(
...
-            return existing_outgoing
+                return {}
+        return {}
 
-        except Exception as e:
-            self.context.logger.error(
-                f"Error fetching Optimism outgoing transfers: {e}"
-            )
-            return raw_existing_outgoing
+    def _count_transfers(self, transfers_dict: Dict) -> int:
```
OjusWiZard: **Important — dead production code kept alive only by tests.** `_count_transfers` lost its only caller when `_track_eth_transfers_and_reversions` was deleted; it now passes the 100%-coverage gate purely because its unit test calls it directly. Same situation for `_load_funding_events_data` (just above) and `check_is_valid_safe_address` (~line 4711, orphaned by the removal of `get_master_safe_address`). Suggest deleting all three methods and their tests so coverage reflects code the system actually runs.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -9785,27 +8411,232 @@ def test_empty_input_returns_empty(self):
         assert _drop_legacy_transfer_entries({}) == ({}, 0)
         assert _drop_legacy_transfer_entries(None) == (None, 0)
 
-    def test_applied_when_loading_optimism_outgoing(self):
-        """Loading optimism_outgoing strips legacy entries lacking transfer_id."""
+
+class TestFreezeEthBasis:
```
OjusWiZard: **Advisory — test gaps in `TestFreezeEthBasis`.** Every `test_freeze_*` stubs `_sum_usdc_investment_from_events`, so the headline invariant `frozen = total - Σ(USDC events)` is only tested as arithmetic over opaque stubs. Worth adding: (1) an end-to-end test where `funding_events` holds real USDC transfers and the freeze reconstructs the *reverted-agent* denominator (the PR's marquee `0xE4EA…50A0` case); (2) coverage of the `delta`-field fallback `transfer.get("delta", ...)` — all current test transfers use `amount`; (3) a mixed `[ETH, USDC]` same-date transfer *without* `tx_hash`, the one case where `_get_transfer_key`'s index fallback can diverge between `_sum_usdc_investment_from_events` (USDC-only enumeration) and `_calculate_chain_investment_value` (all-transfer enumeration), zeroing the USDC match; (4) the exact-zero USDC-only/Base case asserted directly.

OjusWiZard (commented):
Automated multi-perspective review (code-review + silent-failure + test-coverage). The core reconstruction `eth_basis = stored_total - Σ(priced USDC)` is sound, idempotent, and the freeze reads a mutually-consistent prior-run snapshot. No blocking issues. Inline notes below are quality/robustness — none block the existing-agent migration.

DIvyaNautiyal07:
Thanks for the docs, the safety analysis lines up with what I see in the code. Agreed that the reversion data has to stay in some form for legacy Optimism agents, the 0xE4EA fixture confirms it.

The piece I want to push on is the choice of *freezing* the netted ETH basis vs treating `funding_events.json` as the canonical record and computing the initial investment from it directly.

Concretely, what I'd propose: leave `funding_events["optimism"]` (incoming ETH and USDC) and `funding_events["optimism_outgoing"]` (reverted ETH) in place, and compute `initial = sum(priced USDC_in) + sum(priced ETH_in) − sum(priced ETH_out)` each cycle. Both lists already exist on disk on the base branch with the right shape (symbol, amount, tx_hash). The `_priced_transfers` cache keys per transfer, so pricing is one-time per row, not per cycle.

We can still delete everything the PR deletes: the revert-transaction generation, `_track_eth_transfers_and_reversions`, `_fetch_outgoing_transfers_until_date_optimism`, `_track_eth_transfers_mode`, and the incoming ETHER_TRANSFER branch in `_fetch_optimism_transfers_safeglobal`. After this PR, no agent is funded with ETH and no agent ever sends ETH back, so the fetchers and the revert tx are dead weight regardless of approach. The difference is what backs the ETH portion of the ROI denominator going forward.

A few reasons I prefer the direct-read:

The agent never recalculates ETH reversion at runtime once the fetchers are gone. The reverted amount is a fixed number sitting on disk, we just sum it. There's no migration step that has to run on first boot, no `eth_basis_frozen` flag, no ordering concern about whether the migration ran before the first ROI compute, no `_sum_usdc_investment_from_events` filter that exists only to make `total − usdc` work.

It also doesn't depend on `optimism_total_investment` in kv being correct at migration time. If kv is wiped or reset before the first post-upgrade cycle (reinstall, store path change), the freeze computes `max(0 − usdc_sum, 0) = 0` and the ETH portion is silently erased. `funding_events.json` is the durable record, so a direct read survives that case.

And once `eth_basis_frozen` flips and `optimism_outgoing` is popped (line 4330), there's no way back if the snapshot was off. Reading from disk stays correctable because the source data is still there.

End state: `_calculate_chain_investment_value` keeps the ETH branch that's already there, the ETH-out lines get priced and subtracted the same way USDC-in is summed, and we drop the freeze, the migration, the flag, and the asymmetric USDC filter. Same behaviour for reverted agents, same `$0` for Basius, no new state.

Is there something pushing the design towards the kv snapshot that I'm missing? If `optimism_total_investment` is preferred as the source of truth here for a reason, happy to back the freeze, just want it captured.

DhairyaPatel7:
Thanks for the thorough review — all addressed, and the design moved in the process. After a parallel discussion we switched from the frozen-snapshot to **reading the net directly from `funding_events` each cycle**: `Σ priced(incoming USDC + ETH) − Σ priced(reverted ETH from optimism_outgoing)`, with one-time-per-row pricing via the existing `{chain}_priced_transfers` cache. That dissolves the four freeze comments:

- **new-agent locks `eth_basis=0` / negative-clamp / freeze-before-priced / silent-swallow** — there's no snapshot, flag, clamp, or `_load_frozen_eth_basis` anymore; the value is recomputed from the durable record, so a KV reset can't silently zero it and it stays correctable. Your durability point was the deciding factor.
- **dead production code (4346)** — deleted `_count_transfers`, `_load_funding_events_data`, and `check_is_valid_safe_address` along with their tests.
- **`TestFreezeEthBasis` gaps (4415)** — that class is gone; the ideas carried into `TestDirectReadInvestment`: an end-to-end **reverted-agent** case (the `0xE4EA…` shape: incoming USDC + ETH minus reverted ETH-out → net), the `delta`-field fallback, the exact-zero **USDC-only / Base** case, plus airdrop-exclusion and the price-failure / no-coingecko-id branches.

One thing worth flagging on the direct-read: `optimism_outgoing` holds **all** outgoing ETH (the master-safe filter was in the deleted reversion code) — for these agents that *is* the reversion, and subtracting all ETH that left the safe is the correct ROI treatment. The outgoing **fetcher** is deleted but its data stays as a static record read each cycle.

(Force-pushed: PR rebased onto the latest `hack/launch-base` after #328/#329; `fetch_strategies.py` back to 100% coverage, hashes re-locked.)

DIvyaNautiyal07 (commented):
Reviewing the WS2 stack (#331 → #332 → #333) together. Findings range
from one real semantic bug to a behavioural regression on Mode and several
grounded nits. None block on their own, but #2 and #3 below are worth
addressing before this lands.

**Top-level concern (not inline because it's PR metadata):** The PR title is
still `freeze legacy ETH basis & remove ETH-reversion logic` and the
description walks through `_freeze_eth_basis_if_needed`, `frozen_eth_basis`,
`eth_basis_frozen`, `_sum_usdc_investment_from_events`. None of those exist
in the code after the direct-read pivot. Worth rewriting so reviewers and
future `git bisect`-ers get the right mental model. The commit messages on
the latest two commits already describe direct-read, so the description is
the only thing left to update.

Inline comments follow.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3298,81 +3199,94 @@ def _save_priced_transfers(
...
 
-        # Load previously priced transfers so we never re-fetch their prices
+        :param all_transfers: newly-fetched incoming transfers (already merged on disk).
+        :param chain: the chain to value.
+        :param safe_address: the safe address (kept for signature stability).
+        :yield: None: generator yield for async operations.
+        :return: the chain's initial investment in USD.
+        """
+        if not self.funding_events:
+            self.funding_events = self.read_funding_events() or {}
```
DIvyaNautiyal07: **[P1 — real bug, currently masked]** `read_funding_events()` is declared `-> None` (`base.py:1046`) — it side-effects `self.funding_events` and returns `None`. So `self.read_funding_events() or {}` evaluates to `{}` and overwrites the dict that was just loaded from disk. The defensive branch doesn't actually work.

Currently masked because the upstream fetcher always populates `funding_events` before this is called, so the `if not self.funding_events` guard is dead in practice. But the test `test_reads_funding_events_when_empty` exercises this path and asserts `result == 0.0` against the broken implementation — it passes for the wrong reason rather than verifying a disk-load path that doesn't currently work.

Fix: drop the `or {}` — just call `self.read_funding_events()` for its side effect.
```python
if not self.funding_events:
    self.read_funding_events()
    if not self.funding_events:
        self.funding_events = {}
```

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3298,81 +3199,94 @@ def _save_priced_transfers(
     def _calculate_chain_investment_value(
```
DIvyaNautiyal07, line 3266: **[P1 — dead parameters]** `all_transfers` and `safe_address` are never read in the new body — the function reads `self.funding_events` directly. The docstring justifies them as "kept for signature stability" but there's only one caller in the same PR; dropping both is a one-line caller change.

The dead `all_transfers` is also actively misleading: the docstring describes it as "newly-fetched incoming transfers (already merged on disk)" which a future caller could read as a contract and rely on. Once it stops being passed at the call site, that misreading goes away.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3298,81 +3199,94 @@ def _save_priced_transfers(
...
+        incoming_value, inc_updated = yield from self._sum_priced_chain_transfers(
+            incoming, chain, priced_transfers, exclude_airdrop=True
+        )
+        # Reverted ETH that left the safe is a fixed record on disk; price and subtract.
+        outgoing = (self.funding_events or {}).get(f"{chain}_outgoing", {}) or {}
+        reverted_value, out_updated = yield from self._sum_priced_chain_transfers(
+            outgoing, chain, priced_transfers, exclude_airdrop=False
+        )
+
+        if inc_updated or out_updated:
```
DIvyaNautiyal07, line 3320: **[P1 — test gap]** This OR is real (two independent flags) but every test in `TestDirectReadInvestment` populates both legs with uncached rows, so both flags are `True` in every case. A mutation `or → and` would silently break persistence of outgoing-leg prices on the realistic scenario "incoming is fully cached from prior cycles, first outgoing arrives" and no test would catch it.

Add a case with `_load_priced_transfers` pre-populated for the incoming tx_hashes and one uncached outgoing row, then assert `_save_priced_transfers` was called.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3298,81 +3199,94 @@ def _save_priced_transfers(
...
-        cached_investment = 0.0
-        priced_updated = False
 
-        for date, transfers in all_transfers.items():
+        incoming = (self.funding_events or {}).get(chain, {}) or {}
+        incoming_value, inc_updated = yield from self._sum_priced_chain_transfers(
+            incoming, chain, priced_transfers, exclude_airdrop=True
+        )
+        # Reverted ETH that left the safe is a fixed record on disk; price and subtract.
+        outgoing = (self.funding_events or {}).get(f"{chain}_outgoing", {}) or {}
```
DIvyaNautiyal07: **[P2 — redundant guards]** The block at 3220-3221 guarantees `self.funding_events` is a non-falsy dict after that point, so the `(self.funding_events or {})` wrappings on this `.get()` and the one at line 3224 are dead defense. Plain `self.funding_events.get(...)` is correct.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3298,81 +3199,94 @@ def _save_priced_transfers(
...
+
+        total = incoming_value - reverted_value
+        yield from self._save_chain_total_investment(chain, total)
+        self.context.logger.info(
+            f"Total {chain} investment: ${total} "
+            f"(incoming: ${incoming_value}, reverted ETH: ${reverted_value})"
+        )
+        return total
+
+    def _sum_priced_chain_transfers(
```
DIvyaNautiyal07, line 3331: **[P2 — `exclude_airdrop=False` is dead on the outgoing call]** The airdrop filter inside this function is gated on `token_symbol.upper() == "USDC"`. The outgoing dict only contains ETH (both the deleted ETHER_TRANSFER recording and the historical persisted `optimism_outgoing` entries all carry `symbol: "ETH"`), so the filter never fires on the outgoing leg regardless of the flag value. Either drop the parameter and document this helper as incoming-aware only, or keep it but note the asymmetry.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -10721,3 +8962,209 @@ def fetch_mode(address, end_date, fetch_till_date):
...
+        obj.params.airdrop_started = False
+        obj._load_priced_transfers = _gen_return({})
+        obj._save_priced_transfers = _gen_none
+        obj._save_chain_total_investment = _gen_none
+        obj.get_coin_id_from_symbol = lambda s, c: "usd-coin"
+        obj._fetch_historical_token_price = _gen_return(1.0)
+        obj._fetch_historical_eth_price = _gen_return(2600.0)
+        return obj
+
+    def test_reverted_agent_net(self):
```
DIvyaNautiyal07, line 9371: **[P2 — opaque assertion]** `abs(result - (16 + 0.0057 * 2600)) < 1e-6` happens to equal the netted answer because `(16 + (0.0057+0.005)*2600) - 0.005*2600` collapses to that form. The algebra is correct, but the second incoming ETH row (`0.005` on `2025-06-11`) and the outgoing row (`0.005` on `2025-06-20`) are invisible from the asserted expression. A future maintainer who adds an incoming row and forgets to update the expected value gets a silently-passing wrong number.

Rewrite as something like:
```python
incoming_total = 16 + (0.0057 + 0.005) * 2600
outgoing_total = 0.005 * 2600
expected = incoming_total - outgoing_total
assert abs(result - expected) < 1e-6
```
so the netting is explicit and the test name (`reverted_agent_net`) matches what the assertion shows.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
OjusWiZard: **Important — the reverted-ETH leg reads a key (`{chain}_outgoing`) that nothing in the current tree writes.** A grep confirms no code writes `optimism_outgoing` or `mode_outgoing` (only `optimism_withdrawals`, a separate feature). Two consequences:

1. **Mode reversions are never subtracted.** The old base code computed Mode reversions live via `_track_eth_transfers_and_reversions` → `_track_eth_transfers_mode` and never persisted a `mode_outgoing` key. So for a Mode agent that received then reverted ETH, the new code counts the incoming ETH (Mode incoming ETH is still recorded, ~line 3858) but subtracts **0** → denominator inflated. Regression vs base.
2. **Any reversion that happens after this deploy** (any chain) is never recorded here and never subtracted. The leg only works on stale historical Optimism data left by the old code.

If freezing ETH-reversion as historical-only is intentional, please state that in the docstring and accept it; otherwise this subtraction is silently dead for Mode and for all future reversions.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3298,81 +3199,94 @@ def _save_priced_transfers(
...
-        # Track ETH transfers and reversions first
-        reversion_info = yield from self._track_eth_transfers_and_reversions(
-            safe_address, chain
-        )
-        reversion_amount = reversion_info.get("reversion_amount", 0)
-        historical_reversion_value = reversion_info.get(
-            "historical_reversion_value", 0.0
-        )
-        reversion_date = reversion_info.get("reversion_date")
+        funding_events is the canonical record: incoming USDC + ETH live under ``chain``
```
OjusWiZard: **Important — docstring describes a data model that's false for Optimism going forward.** It says "incoming USDC + ETH live under `chain`", but `_fetch_optimism_transfers_safeglobal` (~line 4077) now records **USDC only** (`if symbol.lower() != "usdc": continue`); incoming ETH is no longer captured for Optimism. Existing agents are fine (historical ETH rows persist on disk), but a *freshly-deployed* Optimism agent funded partly in ETH will silently understate its ROI denominator. This is likely the intended USDC-only direction — if so, please update the docstring to "USDC only (Optimism); ETH basis is historical-only" so no one assumes ETH is tracked live.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3383,41 +3297,24 @@ def _calculate_chain_investment_value(
...
+                        priced_transfers[transfer_key] = value
+                        updated = True
                         self.context.logger.info(
-                            f"{chain.upper()} NEW transfer on {date}: {amount} {token_symbol} @ ${price} = ${transfer_value}"
+                            f"{chain.upper()} priced {date}: {amount} {token_symbol} @ ${price} = ${value}"
                         )
                     else:
                         self.context.logger.warning(
-                            f"{chain.upper()} SKIPPED transfer on {date}: {amount} {token_symbol} — price fetch failed"
+                            f"{chain.upper()} SKIPPED {date}: {amount} {token_symbol} — price fetch failed"
```
OjusWiZard, line 3399: **Advisory — a persistently-unpriceable transfer understates the denominator forever, silently.** When pricing fails the row contributes 0, is **not** cached, and re-warns every cycle with no escalation — a token missing from `get_coin_id_from_symbol` (or a permanently failing price endpoint) permanently shrinks the ROI denominator with only a per-cycle WARNING. Pre-existing behavior carried into the new helper. Consider a bounded retry that caches a sentinel (and logs at ERROR) after N cycles so the loss is acknowledged rather than repeated silently.

OjusWiZard (commented):
Second-pass review of the reworked direct-read approach (findings beyond the 6 threads already open). The mechanism is correct for **existing Optimism agents** (their `funding_events.json` already holds `optimism` incoming USDC+ETH and `optimism_outgoing` reverted ETH from the old code). The inline notes flag the cases that aren't that exact one.

**Finding 3 (couldn't attach inline — the line is outside the diff): grand total overwrites the last chain's per-chain value — `fetch_strategies.py:3154`.** `_save_chain_total_investment(chain, total_investment)` sits *outside* the `for chain …` loop, so it writes the cross-chain grand total under the **last** chain's key, clobbering the correct per-chain value that the rework now saves inside `_calculate_chain_investment_value` (line ~3238). It's **pre-existing** (identical on `origin/hack/launch-base`) and harmless for today's single-chain deployments (grand total == chain total), but it's newly visible because of the added inner save, and it's wrong for any multi-chain config — it also poisons the cache-hit early-return at line ~3104. Fix: delete line 3154; per-chain saves already happen at ~3238.

DhairyaPatel7:
Thanks @DIvyaNautiyal07 — addressed the direct-read review in `891344e3`:

- **Signature** — `_calculate_chain_investment_value(chain)` now drops `all_transfers`/`safe_address`; it reads the full persisted `funding_events` each cycle and prices once per row via the `{chain}_priced_transfers` cache. The fetch still runs (it populates `funding_events`), but the valuation no longer depends on "new transfers this cycle".
- **Explicit components in the reverted-agent e2e** — `test_reverted_agent_net` now asserts against `incoming_total = 16 + (0.0057 + 0.005)*2600`, `outgoing_total = 0.005*2600`, `expected = incoming_total - outgoing_total`. This makes the reversion subtraction visible: the 0.005 ETH revert appears in both the incoming (2025-06-11) and outgoing (2025-06-20) ledgers and cancels, leaving USDC + the retained 0.0057 ETH. A second assert pins that this equals `16 + 0.0057*2600`.
- **Load-from-disk now exercised** — `test_reads_funding_events_when_empty` previously passed for the wrong reason (empty record → 0). It now mocks `read_funding_events` to populate a USDC row and asserts the loaded value (12.0) is what drives the result, so the disk-load path is actually covered.
- **OR-isolation** — added `test_save_called_when_only_outgoing_updates`: incoming row pre-cached, only the outgoing ETH row uncached → asserts `_save_priced_transfers` still fires (guards the `inc_updated or out_updated` save, so the newly-priced outgoing row isn't re-fetched every cycle).

100% line+branch coverage on `fetch_strategies.py`; full file suite green (493 tests).

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3296,83 +3200,103 @@ def _save_priced_transfers(
...
+        reversions are subtracted from legacy on-disk data only, and Mode (which never
+        persisted a ``mode_outgoing`` key) gets no reversion subtraction — accepted because
+        Mode ETH is gas-dust and no new reversions occur once the revert-tx path is gone.
+
+        :param chain: the chain to value.
+        :yield: None: generator yield for async operations.
+        :return: the chain's initial investment in USD.
+        """
+        if not self.funding_events:
+            self.read_funding_events()  # side-effects self.funding_events
```
DIvyaNautiyal07, line 3294: **[P2 — same pattern at sister sites]** Good fix here. The same broken form is still present at three other sites in this file (all preexisting on `hack/launch-base`, not introduced by this PR):

- `:3334` in `_fetch_all_transfers_until_date_mode` — `self.funding_events = self.read_funding_events()` overwrites the side-effect-populated dict with the `None` return, then the else-branch sets `{}`.
- `:3417` and `:3426` in `_fetch_all_transfers_until_date_optimism` — same pattern.
- `:4364` in `_track_erc20_transfers_optimism` — the original `read_funding_events() or {}` form.

Effect is loss of the incremental cache on cold start: each fetcher ends up re-fetching full history then re-saving, so it's a performance issue rather than a correctness one (the disk record gets repopulated by `store_funding_events` either way). Worth a mop-up commit in this PR or a follow-up so the fix you just made here doesn't drift back.

DIvyaNautiyal07 (approved):
All my prior P1s and P2s are addressed: the `read_funding_events()` clobber,
the dead `all_transfers`/`safe_address` params, the redundant `or {}` guards,
the dead `exclude_airdrop` flag, the OR-isolation test gap, the opaque
`test_reverted_agent_net` assertion, and the `test_reads_funding_events_when_empty`
test that was passing for the wrong reason. The Mode no-reversion behaviour is
now explicit in the docstring.

Approving. One P2 inline below about the same pattern at sister sites in the
same file; not blocking for this PR but worth a mop-up commit.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3240,14 +3141,17 @@ def calculate_initial_investment_value_from_funding_events(
...
 
+            # ``all_transfers`` is only this cycle's freshly-fetched delta (it merges into
+            # funding_events as a side effect). The valuation reads the full persisted
+            # funding_events directly, so recompute even when nothing new was fetched —
+            # historical rows still count.
             if not all_transfers:
-                self.context.logger.warning(f"No transfers found for {chain} chain")
-                continue
+                self.context.logger.info(
+                    f"No new {chain} transfers this cycle; valuing persisted history"
```
OjusWiZard, line 3197: **Test gap (important) — this control-flow change has no mutation coverage.** Dropping the old `continue` so an empty fetch still values persisted history is the PR's headline behavior, but no test proves it: `test_no_transfers` returns `{}` from the fetcher with an **empty** `funding_events`, so the real valuation returns 0 → `None` whether or not the `continue` is restored; and the integration tests that do return a value (`test_*:~5440/5458`) **stub** `_calculate_chain_investment_value`, bypassing this path. Re-inserting `continue` here passes green. Suggest one test: fetcher returns `{}`, `funding_events[chain]` pre-populated, assert the function returns the persisted-history value (and that `_calculate_chain_investment_value` was actually reached).

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -10721,3 +8962,261 @@ def fetch_mode(address, end_date, fetch_till_date):
...
+                "optimism": {
+                    "2025-01-01": [{"symbol": "USDC", "amount": 12, "tx_hash": "0xR"}]
+                }
+            }
+
+        obj.read_funding_events = _load_from_disk
+        result = _drive(obj._calculate_chain_investment_value("optimism"))
+        assert result == 12.0
+
+    def test_save_called_when_only_outgoing_updates(self):
```
OjusWiZard, line 9515: **Test gap (important) — save-guard FALSE case untested.** This covers `inc_updated=False, out_updated=True → save`. The both-cached arm (`inc_updated=False and out_updated=False → _save_priced_transfers NOT called`) has no call-count assertion, so removing the `if inc_updated or out_updated:` guard at `fetch_strategies.py:3245` (an extra KV write every warm cycle) wouldn't be caught. Add a fully-cached case that asserts `_save_priced_transfers` is not invoked.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py
```diff
@@ -10721,3 +8962,261 @@ def fetch_mode(address, end_date, fetch_till_date):
...
+        obj.get_coin_id_from_symbol = lambda s, c: None
+        obj.funding_events = {
+            "optimism": {
+                "2025-01-01": [{"symbol": "WEIRD", "amount": 3, "tx_hash": "0xW"}]
+            }
+        }
+        result = _drive(obj._calculate_chain_investment_value("optimism"))
+        assert result == 0.0
+
+    def test_price_fetch_failure_warns(self):
```
OjusWiZard, line 9472: **Test gap (important) — "not cached on failure" not asserted.** This checks the skip + warning, but not that the failed key is absent from `priced_transfers` afterward. A mutation caching `0` on price-fetch failure (which would permanently suppress retries once the price API recovers) still yields `result == 0.0` and passes. Assert the transfer key is not present in the cache after the call.

OjusWiZard (approved):
Approving — this commit cleanly addresses the earlier review: dead `all_transfers`/`safe_address` params dropped, `exclude_airdrop` removed (filter now gated on `chain==mode & symbol==USDC`), the `read_funding_events() or {}` masking fixed, and the historical-only ETH/reversion semantics documented. No production correctness issues found at this HEAD.

Remaining items are **test-coverage gaps** (non-blocking) — inline notes below. None require a code change to merge, but finding 1 leaves the PR's headline behavior change without mutation coverage and is worth a follow-up test.

_Still-open advisory (carryover): unpriceable-transfer silent understatement at `fetch_strategies.py:3318`, and the pre-existing grand-total overwrite at `fetch_strategies.py:3157` (single-chain-safe)._

DIvyaNautiyal07 (commented):
## Review findings — P1 and P2

### P1 (Should Fix)

**1. Shared `priced_transfers` cache silently mispriced on a (date, tx_hash) collision between incoming and outgoing legs.**
`fetch_strategies.py:3278-3291`. `_calculate_chain_investment_value` passes the same `priced_transfers` dict to two `_sum_priced_chain_transfers` calls — one for `funding_events[chain]` (incoming USDC/ETH) and one for `funding_events[f"{chain}_outgoing"]` (reverted ETH). `_get_transfer_key` (line 3215) returns `f"{date}_{tx_hash}"` whenever `tx_hash` is present. If a single Safe transaction recorded both ledger sides with the same `tx_hash` (multisend, internal-transfer scenarios, or legacy on-disk data from prior bugs), whichever leg runs first writes its USD value to the shared key; the second leg short-circuits on the cache hit at line 3337 and silently uses the wrong leg's value. The reversion subtraction then produces a silently incorrect basis. Fix: namespace the keys by leg (e.g., `f"in_{date}_{tx_hash}"` vs `f"out_{date}_{tx_hash}"`) or use separate cache dicts.

### P2 (Consider)

**2. Silent disk-read failure produces `$0` investment with no operator signal.**
`fetch_strategies.py:3274-3277`, `base.py:1216-1227`. `_read_data` only `setattr`s `self.funding_events` on a successful JSON load. On `JSONDecodeError` or `PermissionError/OSError`, it logs and returns without updating the attribute. The new caller does `if not self.funding_events: self.funding_events = {}`, silently masking the failure. Distinguishability matches the PR #267 incident shape — `$0` from "no history" vs `$0` from "disk failed" are indistinguishable to ROI consumers. Add an explicit warning before `self.funding_events = {}` so operators see when the empty basis came from a disk failure.

**3. `_sum_priced_chain_transfers` `except Exception` swallows stack trace, masks systemic bugs.**
`fetch_strategies.py:3367-3371`. Per-row catch logs only `str(e)`; no traceback. A schema change or refactor that breaks `transfer.get("symbol")` on every row would silently produce a zero basis. Use `self.context.logger.exception(...)` and narrow the catch to documented failure modes (`ValueError, TypeError, KeyError`) so unexpected exceptions surface.

**4. `_invalidate_cl_pool_cache` failure invisible at the call site.**
`base.py:2028-2029`, `evaluate_strategy.py:3659-3660`. The base method catches all exceptions and returns void. The caller hits `continue` unconditionally. If the kv write fails, the cache entry persists, and the next cycle hits the same `cached_period != current_period` path, fails to invalidate again, and skips the chain forever. Return a bool from `_invalidate_cl_pool_cache` and let the caller decide whether to skip-and-retry or fall through to fresh evaluation.

**5. OR-isolation test gap: `(inc_updated=True, out_updated=False)` is uncovered.**
`test_fetch_strategies.py` (`TestDirectReadInvestment`). `test_save_called_when_only_outgoing_updates` covers `(False, True)`. `test_save_skipped_when_both_legs_cached` covers `(False, False)`. The symmetric `(True, False)` case — only incoming priced fresh, outgoing fully cached — isn't tested. A mutation changing the guard to `if out_updated:` (dropping `inc_updated`) would pass everything. Add the symmetric test.

**6. Fee and gas at-cap boundary tests missing.**
`test_decision_making.py`. `test_slippage_at_cap_passes` pins `>` (not `>=`) for the slippage gate. The fee gate (`max_fee = 5%`) and gas gate (`max_gas = 25%`) have no equivalent at-cap test. Mutating `>` to `>=` on either fee or gas branch is unobserved. Add `test_fee_at_cap_passes` and `test_gas_at_cap_passes`.

**7. Two dead test patches reference `_check_and_create_eth_revert_transactions` after its production deletion.**
`test_fetch_strategies.py:242`, `test_fetch_strategies.py:5401`. The method is removed in this PR's `fetch_strategies.py` diff. The two `obj._check_and_create_eth_revert_transactions = _gen_none` patches are now no-ops setting an attribute on objects whose class no longer defines that name. Delete the patch lines.

**8. Two `TestDirectReadWs1EdgeCoverage` tests assert nothing.**
`test_fetch_strategies.py:9571` (`test_safeglobal_transfer_without_unique_id`) and `test_fetch_strategies.py:9597` (`test_erc20_outgoing_usdc_without_unique_id`). Both invoke a function and end without assertions. Either assert the observable consequence (transfer not appended, `seen_tx_ids` unchanged) or drop the tests — branch exercise alone catches no mutations.

**9. Default profitability thresholds put the slippage cap below the sum of fee + gas caps.**
`service.yaml`, `skill.yaml` (`max_fee_percentage = 0.02`, `max_gas_percentage = 0.25`, `max_slippage_percentage = 0.1`). Since LiFi's `toAmountUSD` is already net of fees, a route at the gas cap (25%) reports ≥25% value loss in the slippage check, which is rejected at the 10% cap. The slippage gate is effectively the combined cap; the gas gate alone can never reject a route the slippage gate didn't already reject (for fee + gas > 10%). The `models.py:1342-1349` comment mentions overlap but doesn't surface the concrete default conflict. Either tighten `max_gas_percentage`, raise `max_slippage_percentage`, or document the intended dominance explicitly.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -2035,12 +2042,12 @@ def _invalidate_cl_pool_cache(self, chain: str) -> Generator[None, None, None]:
                 {kv_key: json.dumps({"invalidated": True}, ensure_ascii=True)}
             )
 
-            self.context.logger.info(
-                f"Invalidated CL pool cache for chain {chain} after successful pool entry"
-            )
+            self.context.logger.info(f"Invalidated CL pool cache for chain {chain}")
+            return True
```
LOCKhart07: **Bool return doesn't reflect a failed KV write.** `_write_kv` returns `False` *without raising* when the KV store responds with a non-`SUCCESS` performative (`_do_connection_request` returns the message; `_write_kv` does `return response == SUCCESS`). Here the result is `yield from`-ed but not captured, and the function unconditionally `return True`. So a soft write failure is reported as success, contradicting the docstring and leaving the new `False`-handling at both call sites (`decision_making.py` post-entry and `evaluate_strategy.py` rollover) dead for everything except an exception.

```python
success = yield from self._write_kv(
    {kv_key: json.dumps({"invalidated": True}, ensure_ascii=True)}
)
if success:
    self.context.logger.info(f"Invalidated CL pool cache for chain {chain}")
else:
    self.context.logger.error(
        f"KV write failed invalidating CL pool cache for {chain}"
    )
return success
```

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3287,14 +3188,17 @@ def calculate_initial_investment_value_from_funding_events(
...
+                )
 
             # Calculate investment value for this chain
-            chain_investment = yield from self._calculate_chain_investment_value(
-                all_transfers, chain, safe_address
-            )
+            chain_investment = yield from self._calculate_chain_investment_value(chain)
             total_investment += chain_investment
 
         yield from self._save_chain_total_investment(chain, total_investment)
```
LOCKhart07: **(Pre-existing, not introduced here — flagging since the rewrite sits on top of it.)** This runs once *after* the `for chain in target_investment_chains` loop, so `chain` is the **last** iterated chain and `total_investment` is the **cross-chain sum**. `_calculate_chain_investment_value` already persists the correct per-chain value internally, so for a multi-chain config (`service.yaml` → base/optimism/mode) this overwrites `mode_total_investment` with the sum of all chains; a later cache-hit `_load_chain_total_investment("mode")` then over-counts. Harmless for the single-chain agent config (`target_investment_chains: ["optimism"]`) where it's a redundant identical write. Worth dropping this line (the per-chain saves already cover it) while the function is being reworked.

LOCKhart07:
**PR description understates the actual scope.** The body describes this as WS2 **2.0–2.3** (ETH-reversion removal, "−641 lines in `fetch_strategies.py`") and frames a stack "PR1 (this) → PR2 (2.4 opportunity cache) → PR3 (2.5 slippage + 2.6 profitability gate)".

But diffing against a freshly-fetched `hack/launch-base`, #332 and #333 were merged **into this branch**, so this PR actually lands all of **2.0–2.6**:

- **2.4** — opportunity-cache rollover invalidation (`evaluate_strategy.py`, `base.py`)
- **2.5 / 2.6** — re-enabled profitability gate + combined value-loss/slippage cap (`decision_making.py`, `models.py`, `services/optimus/service.yaml`, `agents/optimus/aea-config.yaml`)

The "stack" framing and the −641-line figure are misleading for anyone reading this later. Please update the description to reflect that this PR merges the full 2.0–2.6 set.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Good catch — fixed in 5579df4d. `_invalidate_cl_pool_cache` now captures `_write_kv`'s return and propagates it, so the `False`-handling at both call sites is no longer dead for the soft-failure path. The `try` still catches actual exceptions (no `_write_kv` invocation, e.g.) and returns False there too. Added `test_soft_kv_failure_returns_false` to pin it.

Minor naming: kept the `_write_kv` return as `success` (matching the param name on other call sites in the file) rather than `ok` or `wrote` — happy to rename if you'd prefer.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Confirmed and removed in 5579df4d. You're right that the per-chain saves inside `_calculate_chain_investment_value` already cover it — dropping the trailing call is the clean fix.

Added `test_multi_chain_per_chain_total_saved_not_cross_chain_sum` as a regression guard: configures a `["optimism", "mode"]` target list with 50.0 + 25.0 per-chain returns and asserts no `("mode", 75.0)` save fires. The outer return value (75.0) still surfaces the cross-chain sum for the calling round, just doesn't get persisted to a single chain's KV key.

Kept the comment block explaining why the line is gone so this isn't re-added by accident.

DIvyaNautiyal07:
Updated the description — thanks for flagging. The body now reflects that this PR delivers the full **2.0–2.6** set (with #332 / #333 folded in by the rebase onto current `hack/launch-base`), with sub-sections per task and an explicit "review fixes folded in" callout for the three review rounds. Dropped the misleading −641-line figure and the stacked-PR framing.

LOCKhart07 (approved):
Re-reviewed `3280ba2` → `7de63b29`. All three findings from the prior round are resolved, each with a falsifiable test:

- **`_invalidate_cl_pool_cache` soft-KV-failure** (`5579df4d`) — now captures `_write_kv`'s return and propagates it, so the `False`-handling at both call sites is live. Pinned by `test_soft_kv_failure_returns_false` + return-value asserts on `test_success`/`test_exception`.
- **Post-loop cross-chain-sum overwrite** — trailing `_save_chain_total_investment` dropped; the per-chain saves inside `_calculate_chain_investment_value` are the only writes. Pinned by `test_multi_chain_per_chain_total_saved_not_cross_chain_sum`.
- **PR description scope** — updated to reflect the full 2.0–2.6 set.

Remaining hunks in `7de63b29` are pure black-26.1 reformatting and IPFS-hash re-locks. No new issues. LGTM.

PR #332, opened 2026-06-02
fix(evaluate): invalidate stale CL-pool opportunity cache on period rollover (WS2 2.4)

DhairyaPatel7:
**Stacked on #331 (PR1).** Base = `feat/ws2-remove-reversions`.

## WS2 Task 2.4 — invalidate stale cached opportunity

The Velodrome CL opportunity cache (`velodrome_cl_pool_{chain}`) was invalidated **only on a successful pool entry** (`decision_making.py:480`). When the agent **failed to invest** in a period, the cached opportunity survived and was blindly retried in later cycles with stale data (price moved, balances changed) until the 24-hour expiry — "tripping later cycles".

### Fix
- `_cache_cl_pool_data` now records the `period` the opportunity was cached in.
- `_check_and_use_cached_cl_opportunity` invalidates any cache that survives a **period rollover**. Since a successful entry already invalidates the cache, a survivor means the invest failed — so we re-evaluate fresh instead of retrying stale data.
- Within-period reuse is preserved (same `period_count`); old caches without a `period` field are unaffected (backward-compatible).

This is the common BabyDegen workflow fix — it benefits the wider agent family, not just Basius.

### Notes
- The FSM runs `EvaluateStrategy` once per period and settles all actions (swaps + enter-pool) within that period via the `DecisionMaking ↔ PostTxSettlement` loop, so a cache surviving into the next period is unambiguously a failed/un-entered opportunity. The period-rollover check therefore also covers the "on failure" trigger (no separate settlement-failure hook exists).

### Tests
- Added period-rollover (invalidate) and matching-period (reuse) tests; `evaluate_strategy.py` new lines at 100% coverage; **410 tests pass** in the module.

WS2 stack: PR1 (#331, reversion removal) → **PR2 (this, 2.4)** → PR3 (2.5 slippage + 2.6 profitability gate).


packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
```diff
@@ -3636,6 +3636,20 @@ def _check_and_use_cached_cl_opportunity(
                 if not cached_data:
                     continue
 
+                # Invalidate a stale opportunity left over from a previous period. A cache
+                # that survives a period rollover was never successfully entered (a
+                # successful entry invalidates it), so the agent failed to invest it — drop
+                # it and re-evaluate fresh instead of blindly retrying stale data.
+                cached_period = cached_data.get("period")
+                current_period = self.synchronized_data.period_count
+                if cached_period is not None and cached_period != current_period:
```
DIvyaNautiyal07: **P2 — Backward-compat skips existing caches.**

The `cached_period is not None` clause means any cache written by an existing agent before this PR (no `period` field) falls through to the normal expiry validation. For an Optimism agent that has a pre-PR cache AND keeps failing to invest, the bug being fixed isn't actually fixed for up to 24h post-deploy (until `_should_use_cached_cl_data` expires the cache by timestamp). For fresh Basius deployments this is a no-op so not a launch blocker.

Cleanest fix is one line: drop the `is not None` guard and treat absent-period as stale.

```python
if cached_period != current_period:
    ...invalidate...
```

Also worth a one-line comment somewhere here noting that the collapse "on failure + on period rollover" → "on period rollover only" relies on `synchronized_data.period_count` always incrementing between `EvaluateStrategy` invocations. If a future FSM change ever breaks that invariant (e.g. a failed transition that doesn't advance the period), the cache survives forever. Flagging in code makes the assumption explicit.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_evaluate_strategy.py
```diff
@@ -3713,6 +3713,35 @@ def _bad_check():
...
+        assert invalidated  # stale cross-period cache was invalidated
+
+    def test_cache_used_when_period_matches(self):
+        """A cache from the current period is used normally (no invalidation)."""
+        b = _mk()  # synchronized_data.period_count == 1
+        b.current_positions = []
+        b._get_cached_cl_pool_data = _gen_return({"pool_address": "0x1", "period": 1})
+        b._should_use_cached_cl_data = MagicMock(return_value=True)
+        b._update_cl_pool_round_tracking = _gen_none
+        b._reconstruct_actions_from_cached_cl_pool = _gen_return(
```
DIvyaNautiyal07, line 3739: **P2 — Test gap on the backward-compat path.**

`test_cache_invalidated_on_period_rollover` and `test_cache_used_when_period_matches` both seed a `period` field. The third path — cache without `period` (pre-PR shape) — has no test. If we go with the cleaner guard (`cached_period != current_period` without the `is not None`) the existing tests would still pass while changing observable behavior for legacy caches. Add `test_cache_without_period_field_invalidated_or_falls_through` so whichever policy you settle on is pinned.

packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
OjusWiZard: **Test gap (Important): the backward-compat path is untested.** This guard uses `cached_period is not None`, so a pre-PR cache with no `period` key correctly falls through to time-based expiry instead of being invalidated — exactly the behavior the PR description promises. But no test exercises it.

A refactor to a truthiness check (`if not cached_period`) would silently break it — `period: 0` is falsy — and every existing test would still pass. Please add a `test_cache_without_period_field_uses_time_expiry` that feeds a cache dict with no `period` key and asserts `_invalidate_cl_pool_cache` is **not** called and the cache is used normally.

packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
```diff
@@ -3892,6 +3906,7 @@ def _cache_cl_pool_data(
             "pool_finalization_timestamp": current_timestamp,
             "last_round_timestamp": current_timestamp,
             "round_count": 1,  # First round
+            "period": self.synchronized_data.period_count,
```
OjusWiZard, line 3918: **Test gap (Important): this write is never asserted.** This line is half the fix, but the `TestCacheClPoolData` tests only drive `_cache_cl_pool_data` to completion without inspecting the `_write_kv` payload — deleting this line would not fail any test. Please capture the `_write_kv` argument and assert `json.loads(payload)["period"]` equals the expected `period_count`.

OjusWiZard (commented):
Automated multi-perspective review. Core logic is **correct** — the period-rollover guard is sound, types are JSON-safe, the `period` field survives the `_update_cl_pool_round_tracking` rewrite, and successful invalidation terminates cleanly (the `invalidated` marker is honored in `_get_cached_cl_pool_data`). No correctness bugs. The findings are test gaps (inline) plus a few accuracy nits below.

**Advisory — not on diff lines, so noted here:**

- **Stale docstring + misleading log in `_invalidate_cl_pool_cache` (`base.py:1652, 1666`).** The method now has two callers, but its docstring still says *"Called when enter pool succeeds"* and it unconditionally logs *"…invalidating … after successful pool entry."* When invoked from the new rollover path this is factually wrong and prints right after the correct rollover log — it will mislead log-based diagnosis. Generalize the message or pass a reason.
- **`period`-preservation through `_update_cl_pool_round_tracking` is untested (`evaluate_strategy.py:3868`).** Within-period reuse depends on that rewrite keeping `period`. It does today, but `TestUpdateClPoolRoundTracking` only asserts `round_count`/`last_round_timestamp`; a future `cached_data.pop("period")` would silently break the guard with no test failure. Assert `period` survives in the written payload.
- **`_invalidate_cl_pool_cache` ignores the `_write_kv` return value (`base.py:1661`).** Pre-existing, not introduced here — but this PR adds a caller that relies on the write landing. If the write fails the stale cache is re-read next cycle. Optional: log on `False`.
- **Broad `except Exception` logs `str(e)` only (`evaluate_strategy.py:3680`).** Pre-existing; consider `exc_info=True` for a stack trace and chain context. Minor.

DhairyaPatel7:
Thanks both — addressed, with one design call to flag.

**Guard policy (@DIvyaNautiyal07 vs @OjusWiZard).** You proposed opposite handling of a legacy cache with no `period` field: invalidate-as-stale vs fall-through-to-time-expiry. I went with **invalidate** — `cached_period != current_period`, so an absent `period` (`None`) never equals the int and is dropped. Rationale: a legacy cache that survives to a post-deploy cycle *is* exactly the stale-retry the PR fixes, and the cost of dropping a still-valid one is a single re-evaluation (the swaps already happened; re-eval just re-prices with current balances). It also fixes the bug immediately rather than leaving legacy agents exposed for up to 24h. I used `!=` (not a truthiness check), so `period == 0` stays valid within its own period — @OjusWiZard's falsy-0 concern doesn't apply.

I added the **invariant comment** (the collapse to "rollover only" relies on `period_count` advancing between `EvaluateStrategy` invocations).

**Test gaps** (both): added `test_cache_without_period_field_invalidated` (pins the policy above), an assertion that `_cache_cl_pool_data` writes `period == period_count`, and that `period` **survives** `_update_cl_pool_round_tracking`. The existing cache-flow tests (`test_cache_expired`, valid-with-actions, actions-none) are pinned to the current period.

**`_invalidate_cl_pool_cache` docstring/log** (@OjusWiZard): generalized — it no longer claims "after successful pool entry" since the rollover path also calls it.

The `_write_kv`-return and broad-`except` items are pre-existing; left as-is to keep the diff scoped.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_evaluate_strategy.py
```diff
@@ -3713,6 +3713,52 @@ def _bad_check():
         result = _drive(b._check_and_use_cached_cl_opportunity())
         assert result is None
 
+    def test_cache_invalidated_on_period_rollover(self):
```
DIvyaNautiyal07, line 3716: **[P1 — period-0 case untested]** The docstring on the production check explicitly promises that `!=` (not a truthy check) keeps `period == 0` valid within its own period. But `_mk()` hardcodes `period_count=1` and no test overrides it to 0. A mutation `cached_period != current_period` → `not cached_period` would silently invalidate valid period-0 caches and the suite wouldn't catch it.

Add a case where the synced data has `period_count=0` and the cached dict has `"period": 0`, then assert the cache is used (not invalidated).

DIvyaNautiyal07, line 3716: **[P2 — mutation budget]** This case uses `period=0, current=1`, which is also `current - 1` and `< current`. The check `!=` is indistinguishable from `<` or `== current - 1` from the test side. Adding a second rollover case like `period=2, current=3` closes that — proves the check is membership-based, not just adjacent-lag.

OjusWiZard (approved):
Approving. The period-rollover invalidation logic is correct, types are JSON-safe, the `period` field survives the round-tracking rewrite, and invalidation terminates cleanly. All review findings from the multi-perspective pass have been addressed (backward-compat path tested, period write asserted, round-tracking preservation tested, docstring/log corrected, FSM invariant documented).

Two optional test-hardening items remain open (period-0 validity and a non-adjacent rollover case) — mutation coverage only, not blockers. LGTM. 🚀

DhairyaPatel7:
Thanks @DIvyaNautiyal07 @OjusWiZard — addressed in `33751b82` (+ the guard already landed earlier).

**Policy decision on the legacy/no-`period` cache.** The two suggestions point in opposite directions, so flagging the call explicitly:
- @DIvyaNautiyal07 — drop the `is not None` guard so an absent `period` is treated as stale.
- @OjusWiZard — keep it so a pre-PR cache falls through to time-based expiry.

I went with **dropping the guard** (`if cached_period != current_period:`). Rationale: for a pre-PR Optimism agent that has a stale cache *and* keeps failing to invest, treating absent-period as stale fixes the bug immediately rather than leaving it broken for up to 24h until time-expiry — which is the whole point of the PR. For fresh Basius it's a no-op either way. So @OjusWiZard, I did **not** add `test_cache_without_period_field_uses_time_expiry` (it would assert the behaviour we deliberately dropped); instead the legacy path is pinned to the chosen policy below. Happy to flip if you'd rather preserve the fall-through.

Pinned tests:
- `test_cache_without_period_field_invalidated` — legacy cache (no `period`) → `_invalidate_cl_pool_cache` **is** called.
- `test_cache_used_when_period_zero` — `period_count=0` + cached `period=0` → used, not invalidated (pins the `!=`-not-truthiness promise; a `not cached_period` mutation would fail it). [@DIvyaNautiyal07 P1]
- `test_cache_invalidated_non_adjacent_period` — `period=2, current=3` → invalidated, proving the check is membership-based, not adjacent-lag (`<` / `current-1`). [@DIvyaNautiyal07 P2]
- `test_basic_cache` already asserts `json.loads(payload)["period"] == period_count`, so deleting the write would fail a test. [@OjusWiZard, evaluate_strategy.py:3918]

Also added the in-code note that the "invalidate on failure + rollover" → "rollover only" collapse relies on `synchronized_data.period_count` advancing between `EvaluateStrategy` invocations.

DIvyaNautiyal07 (approved):
Period-0 case (`test_cache_used_when_period_zero`) and the
non-adjacent rollover (`test_cache_invalidated_non_adjacent_period`,
`period=2, current=3`) both address the test-gap asks. The first guards the
`!=`-vs-truthiness mutation for first-period agents; the second proves the
guard is membership-based rather than adjacent-lag. Approving.

PR #333, opened 2026-06-02
feat(safety): re-enable profitability gate + add swap slippage cap (WS2 2.5/2.6)

DhairyaPatel7:
**Stacked on #332 (PR2).** Base = `feat/ws2-opportunity-cache`.

### 2.6 — Profitability gate
Re-enables the fee/gas gate in `check_if_route_is_profitable` (`decision_making.py`) that was commented out for the Basius MVP. Routes whose fee% or gas% exceed `max_fee_percentage` / `max_gas_percentage` are rejected again. **This also fixes the pre-existing `test_not_profitable_fees` failure** that was red on the branch (and inherited by PR1/PR2).

### 2.5 — Slippage protection (config restore + swap cap)
- **Config restore:** `service.yaml` `slippage_tolerance` default `1.0 → 0.01` (still `${SLIPPAGE_TOLERANCE:...}`-overridable). `1.0` = 100% made LP add/remove min-amounts ≈ 0, i.e. protection off. The final production value is calibrated in **WS3 3.5**.
- **Swap-level cap:** new `max_slippage_percentage` param + an agent-side value-loss reject in `check_if_route_is_profitable` using `(fromAmountUSD − toAmountUSD)/fromAmountUSD`, independent of LiFi's own slippage handling. Wired through `models.py`, `skill.yaml`, `aea-config.yaml`, `service.yaml` (default `0.1`, env `MAX_SLIPPAGE_PERCENTAGE`).

### Tests
- `_make_behaviour` sets the new param; added `test_not_profitable_slippage` (passes fee/gas gate, exceeds the value-loss cap → rejected); `test_models` config updated. `decision_making.py` at **100% coverage**; full skill suite green except one **pre-existing, unrelated** failure (`test_velodrome.py::…::test_query_returns_none` — the WS3-3.3 disabled Aerodrome check, also red on the base branch).

### Notes
- The slippage cap default (`0.1`) and `slippage_tolerance` (`0.01`) are interim/protective; **WS3 3.4/3.5** calibrate the production values for thin Base stables.

WS2 stack: PR1 (#331) → PR2 (#332) → **PR3 (this, 2.5 + 2.6)**.

### DIFF
--- a/packages/valory/services/optimus/service.yaml
+++ b/packages/valory/services/optimus/service.yaml
@@ -6,7 +6,7 @@ aea_version: '>=2.0.0, <3.0.0'
 license: Apache-2.0
 fingerprint: {}
 fingerprint_ignore_patterns: []
-agent: valory/optimus:0.1.0:bafybeicu4cu2sldqmluyqg2nxa45zsp2hxyet7ryl4lh3tgaccfbyavhbe
+agent: valory/optimus:0.1.0:bafybeiaspcsi72v37xd6q7kycjm6x3g2o3ya5jhbojpewygca6muheuoci
 number_of_agents: 1
 deployment:
   agent:
--- a/packages/valory/services/optimus/service.yaml
+++ b/packages/valory/services/optimus/service.yaml
@@ -6,7 +6,7 @@ aea_version: '>=2.0.0, <3.0.0'
 license: Apache-2.0
 fingerprint: {}
 fingerprint_ignore_patterns: []
-agent: valory/optimus:0.1.0:bafybeiev3v5ke2tbnz66u7bzonhku7n5v3saacxgfhojbs3uehavwbi5da
+agent: valory/optimus:0.1.0:bafybeicu4cu2sldqmluyqg2nxa45zsp2hxyet7ryl4lh3tgaccfbyavhbe
 number_of_agents: 1
 deployment:
   agent:
--- a/packages/valory/services/optimus/service.yaml
+++ b/packages/valory/services/optimus/service.yaml
@@ -93,6 +93,7 @@ models:
       min_investment_amount: ${MIN_INVESTMENT_AMOUNT:int:1}
       max_fee_percentage: ${MAX_FEE_PERCENTAGE:float:0.02}
       max_gas_percentage: ${MAX_GAS_PERCENTAGE:float:0.25}
+      max_slippage_percentage: ${MAX_SLIPPAGE_PERCENTAGE:float:0.1}
       balancer_graphql_endpoints: ${BALANCER_GRAPHQL_ENDPOINTS:str:{"optimism":"https://api.studio.thegraph.com/query/75376/balancer-optimism-v2/version/latest","base":"https://api.studio.thegraph.com/query/24660/balancer-base-v2/version/latest","mode":"https://api.studio.thegraph.com/query/75376/balancer-mode-v2/version/latest"}}
       allowed_chains: ${ALLOWED_CHAINS:list:["base","optimism","mode"]}
       target_investment_chains: ${TARGET_INVESTMENT_CHAINS:list:["base","optimism","mode"]}
@@ -111,7 +112,7 @@ models:
       genai_api_key: ${GENAI_API_KEY:str:""}
       airdrop_started: ${AIRDROP_STARTED:bool:false}
       airdrop_contract_address: ${AIRDROP_CONTRACT_ADDRESS:str:""}
-      slippage_tolerance: ${SLIPPAGE_TOLERANCE:float:1.0}
+      slippage_tolerance: ${SLIPPAGE_TOLERANCE:float:0.01}
       genai_model: ${GENAI_MDOEL:str:gemini-2.5-flash}
       slippage_for_swap: ${SLIPPAGE_FOR_SWAP:float:0.005}
       use_x402: ${USE_X402:bool:true}
--- a/packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
+++ b/packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
@@ -2239,9 +2239,32 @@ class DecisionMakingBehaviour(LiquidityTraderBaseBehaviour):
             f"Details: total_fee={total_fee}, total_gas_cost={total_gas_cost}, from_amount_usd={from_amount_usd}, to_amount_usd={to_amount_usd}"
         )
 
-        # TEMP (Basius MVP): profitability gate bypassed, re-enable per
-        # workstream 2 PR 2.6 by gating on fee_percentage/gas_percentage
-        # against allowed_fee_percentage/allowed_gas_percentage.
+        if (
+            fee_percentage > allowed_fee_percentage
+            or gas_percentage > allowed_gas_percentage
+        ):
+            self.context.logger.error(
+                f"Route is not profitable: fee {fee_percentage:.2f}% "
+                f"(allowed {allowed_fee_percentage:.2f}%), gas {gas_percentage:.2f}% "
+                f"(allowed {allowed_gas_percentage:.2f}%)"
+            )
+            return False, None, None
+
+        # Combined value-loss cap: (fromAmountUSD - toAmountUSD) / fromAmountUSD bounds the
+        # total value lost on the route. LiFi's toAmountUSD is net of protocol/integrator
+        # fees, so this captures fees + price impact together (an agent-side backstop
+        # independent of LiFi's own slippage param) — it is NOT market slippage alone, and
+        # can fire even when the fee/gas gates above pass.
+        slippage_percentage = (
+            (from_amount_usd - to_amount_usd) / from_amount_usd
+        ) * 100
+        allowed_slippage_percentage = self.params.max_slippage_percentage * 100
+        if slippage_percentage > allowed_slippage_percentage:
+            self.context.logger.error(
+                f"Route slippage too high: {slippage_percentage:.2f}% of input "
+                f"(allowed {allowed_slippage_percentage:.2f}%)"
+            )
+            return False, None, None
 
         self.context.logger.info("Route is profitable!")
         return True, total_fee, total_gas_cost
--- a/packages/valory/skills/liquidity_trader_abci/models.py
+++ b/packages/valory/skills/liquidity_trader_abci/models.py
@@ -479,6 +479,11 @@ class Params(BaseParams):
         self.min_investment_amount = self._ensure("min_investment_amount", kwargs, int)
         self.max_fee_percentage = self._ensure("max_fee_percentage", kwargs, float)
         self.max_gas_percentage = self._ensure("max_gas_percentage", kwargs, float)
+        # Stored as a FRACTION (0.1 == 10%), not a percent — multiplied by 100 at the
+        # use-site. Matches max_fee_percentage / max_gas_percentage above.
+        self.max_slippage_percentage = self._ensure(
+            "max_slippage_percentage", kwargs, float
+        )
         self.balancer_graphql_endpoints = json.loads(
             self._ensure("balancer_graphql_endpoints", kwargs, str)
         )
--- a/packages/valory/skills/liquidity_trader_abci/models.py
+++ b/packages/valory/skills/liquidity_trader_abci/models.py
@@ -481,6 +481,12 @@ class Params(BaseParams):
         self.max_gas_percentage = self._ensure("max_gas_percentage", kwargs, float)
         # Stored as a FRACTION (0.1 == 10%), not a percent — multiplied by 100 at the
         # use-site. Matches max_fee_percentage / max_gas_percentage above.
+        # NOTE: this gate bounds COMBINED value loss (fromAmountUSD - toAmountUSD), which
+        # includes protocol/integrator fees as well as price impact, so it overlaps the
+        # fee gate. Set it with max_fee_percentage + max_gas_percentage headroom in mind:
+        # a route at the fee/gas limits already reports ~that much value loss here even at
+        # zero price impact, so a value too close to those limits rejects otherwise-fine
+        # routes. See check_if_route_is_profitable for the use-site.
         self.max_slippage_percentage = self._ensure(
             "max_slippage_percentage", kwargs, float
         )
--- a/packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
+++ b/packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
@@ -62,8 +62,10 @@ def _make_behaviour(**overrides: Any) -> DecisionMakingBehaviour:
     params.lifi_fetch_step_transaction_url = "https://li.fi/step"
     params.lifi_advance_routes_url = "https://li.fi/routes"
     params.waiting_period_for_status_check = 0
-    params.max_fee_percentage = 0.05
-    params.max_gas_percentage = 0.05
+    # Distinct so a production-side threshold swap (fee checked against the gas limit
+    # or vice-versa) is observable in TestCheckIfRouteIsProfitable.
+    params.max_fee_percentage = 0.05  # 5%
+    params.max_gas_percentage = 0.25  # 25%
     params.max_slippage_percentage = 0.1
     params.slippage_for_swap = 0.03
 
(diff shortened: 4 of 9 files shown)


packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
```diff
@@ -2239,9 +2239,25 @@ def check_if_route_is_profitable(
...
         )
 
-        # TEMP (Basius MVP): profitability gate bypassed, re-enable per
-        # workstream 2 PR 2.6 by gating on fee_percentage/gas_percentage
-        # against allowed_fee_percentage/allowed_gas_percentage.
+        if (
+            fee_percentage > allowed_fee_percentage
+            or gas_percentage > allowed_gas_percentage
+        ):
+            self.context.logger.error("Route is not profitable!")
```
DIvyaNautiyal07: **Nit — error message asymmetry.**

`"Route is not profitable!"` doesn't say whether fee or gas tripped, or by how much, or against what threshold. The slippage error a few lines below (`f"Route slippage too high: {slippage_percentage:.2f}% of input (allowed {allowed_slippage_percentage:.2f}%)"`) does. For oncall / debug-from-logs consistency, either split into two checks with their own messages, or fold `fee_percentage`, `gas_percentage`, and their allowed thresholds into this one. The existing `logger.info` two lines above already prints them; this error just doesn't carry them through.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
```diff
@@ -2239,9 +2239,25 @@ def check_if_route_is_profitable(
...
+            return False, None, None
+
+        # Slippage / value-loss cap: reject routes that lose more than the allowed
+        # percentage of the input value, independent of LiFi's own slippage handling.
+        slippage_percentage = (
+            (from_amount_usd - to_amount_usd) / from_amount_usd
+        ) * 100
+        allowed_slippage_percentage = self.params.max_slippage_percentage * 100
+        if slippage_percentage > allowed_slippage_percentage:
```
OjusWiZard, line 2262: **Important — confirm what `toAmountUSD` includes.**

This value-loss cap runs *after* the fee/gas gate. If LiFi's `toAmountUSD` is already **net of protocol/integrator fees**, then `(fromAmountUSD − toAmountUSD)` captures *fees + price impact combined* — so a route can pass the 2% fee gate and 25% gas gate yet still be rejected here by the 10% cap (or vice-versa). The three caps then interact in a way the name `max_slippage_percentage` doesn't make obvious.

(Gas is paid in native token and normally isn't deducted from `toAmount`, so the gas overlap is likely a non-issue — it's the *fee* overlap that matters.)

If this is intentionally a **combined value-loss** guard (as the PR description reads), a one-line comment saying so would prevent future confusion. If it's meant to be market-slippage-only, add `total_fee`/`total_gas_cost` back before computing the ratio.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
```diff
@@ -2710,6 +2705,17 @@ def test_zero_amounts(self):
         result = _exhaust(b.check_if_route_is_profitable(route))
         assert result[0] is False
 
+    def test_not_profitable_slippage(self):
```
OjusWiZard, line 2731: **Important — the re-enabled gate's gas branch is untested.** The gate is `fee_percentage > allowed or gas_percentage > allowed`, but `test_not_profitable_fees` uses `fee=50, gas=0`, so only the fee branch is exercised. A gas-only rejection would never be caught (e.g. if `or` were ever changed to `and`):

```python
def test_not_profitable_gas_only(self):
    b = _make_behaviour()  # max_gas_percentage = 0.05
    b._get_step_transactions_data = _make_gen_method([{"fee": 0.5, "gas_cost": 10}])
    route = {"fromAmountUSD": "100", "toAmountUSD": "99", "steps": []}
    assert _exhaust(b.check_if_route_is_profitable(route))[0] is False
```

**Advisory — two cheap regression guards for the new slippage cap:**
- *Exact boundary:* the cap uses strict `>`, so 10% loss (`100 → 90`) must **pass** — guards against an accidental `>=`.
- *Negative slippage:* a value-gaining route (`toAmountUSD > fromAmountUSD`) correctly passes — asserting it guards against a sign flip in the subtraction.

Also consider asserting `result[1]`/`result[2]` (fee/gas) in `test_profitable`.

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -479,6 +479,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.min_investment_amount = self._ensure("min_investment_amount", kwargs, int)
         self.max_fee_percentage = self._ensure("max_fee_percentage", kwargs, float)
         self.max_gas_percentage = self._ensure("max_gas_percentage", kwargs, float)
+        self.max_slippage_percentage = self._ensure(
```
OjusWiZard, line 490: **Advisory — name vs. value.** `max_slippage_percentage` is stored as a fraction (`0.1`) and multiplied by `*100` in the body, but the name reads as "percentage" — an operator could set `10.0` expecting 10% and get 1000%. This is **consistent with the existing** `max_fee_percentage`/`max_gas_percentage`, so it's inherited, not introduced here. Not worth diverging in this one PR; a one-line unit comment on the attribute (or a repo-wide rename later) would close the ambiguity.

OjusWiZard (commented):
Multi-perspective review of the WS2 2.5/2.6 safety checks. No confirmed Critical bug — the slippage division is safely guarded by the pre-existing `if not from_amount_usd` early return, the fee-rejection/zero/happy paths are correct, and `decision_making.py` is at 100% coverage. Two Important items inline (one semantics question, one test gap) plus one advisory on param naming.

DhairyaPatel7:
Thanks — all addressed.

- **Error-message asymmetry** (@DIvyaNautiyal07): the gate rejection now logs the fee/gas percentages and their thresholds, matching the slippage message.
- **What `toAmountUSD` includes** (@OjusWiZard): confirmed it's intentionally a **combined value-loss** guard — `(fromAmountUSD − toAmountUSD)/fromAmountUSD` captures fees + price impact (LiFi's `toAmountUSD` is net of protocol/integrator fees), an agent-side backstop independent of LiFi's slippage param. Added a comment saying so, including that it can fire even when the fee/gas gates pass.
- **Gate gas branch + slippage guards** (@OjusWiZard): added `test_not_profitable_gas_only` (exercises the `or gas` branch), `test_slippage_at_cap_passes` (exactly 10% loss passes — guards against `>=`), `test_value_gaining_route_passes` (negative loss passes — guards a sign flip), and asserted `result[1]`/`result[2]` (fee/gas) in `test_profitable`.
- **Param name vs value** (@OjusWiZard): added a one-line comment on `max_slippage_percentage` noting it's a fraction (0.1 == 10%), consistent with `max_fee_percentage`/`max_gas_percentage`.

`decision_making.py` stays at 100% coverage.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
```diff
@@ -64,6 +64,7 @@ def _make_behaviour(**overrides: Any) -> DecisionMakingBehaviour:
     params.waiting_period_for_status_check = 0
     params.max_fee_percentage = 0.05
     params.max_gas_percentage = 0.05
+    params.max_slippage_percentage = 0.1
```
DIvyaNautiyal07, line 69: **[P1 — fee/gas variable swap is undetectable]** `max_fee_percentage` and `max_gas_percentage` are both set to `0.05` in this fixture (lines 65-66, just above). That means a production-side swap like `fee_percentage > allowed_gas_percentage` (checking the wrong threshold) would silently pass every test in `TestCheckIfRouteIsProfitable`. The two gates being re-enabled by this PR are exactly the kind of safety check where a variable swap is a real risk.

Fix: set them to distinct values, e.g. `max_fee_percentage = 0.05` and `max_gas_percentage = 0.25`, and tune the test data so the gas-only and fee-only cases each fail against their own threshold but not the other's.

packages/valory/services/optimus/service.yaml
```diff
@@ -111,7 +112,7 @@ models:
       genai_api_key: ${GENAI_API_KEY:str:""}
       airdrop_started: ${AIRDROP_STARTED:bool:false}
       airdrop_contract_address: ${AIRDROP_CONTRACT_ADDRESS:str:""}
-      slippage_tolerance: ${SLIPPAGE_TOLERANCE:float:1.0}
+      slippage_tolerance: ${SLIPPAGE_TOLERANCE:float:0.01}
```
DIvyaNautiyal07, line 115: **[P2 — `slippage_tolerance` change has wider blast radius than LP]** Worth knowing this param is also read by `pools/velodrome.py:147-150` and `:233` for the price-monitoring guard (price-movement %, not min-amounts). Going from `1.0` → `0.01` tightens that gate from "100% movement allowed" to "1% triggers monitoring", which can block Velodrome LP entries on modest volatility.

Agents running with `SLIPPAGE_TOLERANCE=1.0` in their env keep the loose behavior; agents without an env override pick up the tight `0.01`. The change is the right direction, but operators should be told to either remove the stale override or set it explicitly to their intended value before this deploys.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
```diff
@@ -2239,9 +2239,32 @@ def check_if_route_is_profitable(
...
+        # Combined value-loss cap: (fromAmountUSD - toAmountUSD) / fromAmountUSD bounds the
+        # total value lost on the route. LiFi's toAmountUSD is net of protocol/integrator
+        # fees, so this captures fees + price impact together (an agent-side backstop
+        # independent of LiFi's own slippage param) — it is NOT market slippage alone, and
+        # can fire even when the fee/gas gates above pass.
+        slippage_percentage = (
+            (from_amount_usd - to_amount_usd) / from_amount_usd
+        ) * 100
+        allowed_slippage_percentage = self.params.max_slippage_percentage * 100
+        if slippage_percentage > allowed_slippage_percentage:
```
DIvyaNautiyal07, line 2262: **[P2 — slippage cap conceptually double-counts fees]** The in-code comment correctly flags this as a combined value-loss guard, not pure market slippage. But the practical implication is that a route with `fee = max_fee_percentage`, `gas = max_gas_percentage`, `price_impact = 0` reports `fee + gas` as value loss and gets rejected here even though both gates above just passed it.

Not a bug, but operators setting `MAX_SLIPPAGE_PERCENTAGE` need to know the effective combined threshold is tighter than the number implies. Worth either a sentence in the param's docstring in `models.py`, or a comment here naming the relationship: `max_slippage_percentage` should be set with `max_fee_percentage + max_gas_percentage` headroom in mind.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
```diff
@@ -2671,21 +2672,17 @@ class TestCheckIfRouteIsProfitable:
...
+        assert result[1] == 0.1  # total_fee
+        assert result[2] == 0.1  # total_gas_cost
 
-    @pytest.mark.skip(
-        reason=(
-            "Profitability gate is bypassed for the Basius MVP — re-enable per "
-            "workstream 2 PR 2.6 and un-skip this test alongside that change."
-        )
-    )
     def test_not_profitable_fees(self):
```
DIvyaNautiyal07, line 2688: **[P2 — test would silently route through slippage cap if fee gate weakened]** The route uses `fromAmountUSD=100, toAmountUSD=50` → 50% value loss, far over the 10% slippage cap. Fee gate fires first, so the assertion is right today. But if the fee gate were ever weakened or removed, this test would still pass via the slippage path and the name (`test_not_profitable_fees`) would lie about what it covers.

Pin to the fee gate by using `toAmountUSD="99"` — that's 1% value loss (under the 10% cap), so only the fee gate can reject it.

OjusWiZard (approved):
All review findings addressed and verified in code:
- Combined value-loss cap now documented (states `toAmountUSD` is net of fees and can fire even when fee/gas gates pass).
- Gas-only branch, exact-boundary, and value-gaining (negative-slippage) cases all covered; happy path now asserts returned fee/gas.
- `max_slippage_percentage` fraction-vs-percent convention documented.

No Critical/Important issues remaining. LGTM 🚢

DhairyaPatel7:
Thanks @DIvyaNautiyal07 @OjusWiZard — addressed in `e515cb44`.

**Threshold identity (was undetectable with both at 0.05).** Fixture now uses distinct values — `max_fee_percentage=0.05` (5%), `max_gas_percentage=0.25` (25%) — and two complementary canaries pin which threshold each gate reads:
- `test_not_profitable_fees` — fee 10% (between the two limits), gas 0, 1% value loss → only the fee gate can reject. A swap to `fee_percentage > allowed_gas_percentage` (10% > 25% → False) would let it through. Also pins the test to the fee gate via `toAmountUSD="99"` so it can't pass through the slippage cap if the fee gate were ever weakened. [@DIvyaNautiyal07 P1 / P2]
- `test_gas_within_gas_limit_passes` — gas 10% (between the limits), fee 0 → must pass. A swap to `gas_percentage > allowed_fee_percentage` (10% > 5% → True) would wrongly reject it. [@OjusWiZard, the `or gas` branch]
- `test_not_profitable_gas_only` retuned to gas 30% > 25%.

**Combined value-loss overlap.** Confirmed `toAmountUSD` is net of LiFi protocol/integrator fees, so the cap is a combined value-loss guard (fees + price impact), not market slippage — kept the use-site comment and added a `models.py` note that `max_slippage_percentage` should be set with `max_fee_percentage + max_gas_percentage` headroom. [@OjusWiZard models.py:484 / @DIvyaNautiyal07 P2]

**`slippage_tolerance` blast radius — confirmed, and it's more than monitoring.** [@DIvyaNautiyal07 service.yaml:115] The param is dual-purpose:
1. **LP min-amount protection** (`velodrome.py` `_calculate_slippage_protection` and the stable/volatile add/exit paths): `amount_min = expected * (1 - slippage_tolerance)`. At the old default `1.0` this is `min = 0` — i.e. slippage protection was effectively **disabled**. `0.01` restores real 1% protection, so this is a correctness fix, not just a tightening.
2. **Velodrome price-movement monitoring** (`velodrome.py:147-150, 233, 725, 751`): `price_moved = |Δ%| > slippage_tolerance*100`, so `1.0 → 0.01` moves the trigger from 100% → 1% and can defer LP entries on modest volatility, as you noted.

**Operator guidance:** agents still carrying `SLIPPAGE_TOLERANCE=1.0` in their env keep both the (unsafe) loose LP protection and the loose monitoring; agents without an override pick up `0.01`. Operators should remove the stale override or set it explicitly to their intended value before this deploys. I've left `service.yaml` comment-free (the file carries no inline comments today) and documented this in the WS2 change notes instead.

DIvyaNautiyal07 (approved):
Fee/gas thresholds are now distinct (5% / 25%), the two canary
tests (`test_not_profitable_fees` at fee 10% / gas 0, and
`test_gas_within_gas_limit_passes` at fee 0 / gas 10%) catch a swap on either
side without depending on the other gate firing. The `max_slippage_percentage`
overlap with fee/gas is documented in `models.py`. Approving.

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -479,6 +479,17 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.min_investment_amount = self._ensure("min_investment_amount", kwargs, int)
         self.max_fee_percentage = self._ensure("max_fee_percentage", kwargs, float)
         self.max_gas_percentage = self._ensure("max_gas_percentage", kwargs, float)
+        # Stored as a FRACTION (0.1 == 10%), not a percent — multiplied by 100 at the
+        # use-site. Matches max_fee_percentage / max_gas_percentage above.
+        # NOTE: this gate bounds COMBINED value loss (fromAmountUSD - toAmountUSD), which
+        # includes protocol/integrator fees as well as price impact, so it overlaps the
+        # fee gate. Set it with max_fee_percentage + max_gas_percentage headroom in mind:
```
OjusWiZard, line 486: **Advisory (docs-only, non-blocking).** This contradicts the `decision_making.py` comment added in this same PR, which says the cap *"captures fees + price impact together"* — i.e. **gas excluded**. That one is accurate: LiFi gas is paid in the native token from the wallet, not deducted from `toAmountUSD`, so `fromAmountUSD − toAmountUSD` reflects fees + price impact, not gas.

Advising `max_fee_percentage + max_gas_percentage` headroom over-provisions, which nudges toward a *looser* (less protective) slippage cap. Suggest aligning the two comments:

> Set it with `max_fee_percentage` headroom in mind (plus a price-impact margin) — a route at the fee limit already reports ~that much value loss here even at zero price impact.

Not worth blocking the merge over.

OjusWiZard (approved):
Re-reviewed the newer `e515cb44` changes — the fee/gas threshold-identity canaries are correct and close the last open gap, package hashes are consistent, working tree clean. Approving. One documentation-only nit inline (non-blocking).

PR #334, opened 2026-06-03
feat(safe-api): route Safe API through proxy and strengthen 429 retry

DIvyaNautiyal07:
## Summary

First half of the Safe API routing work. Caching changes are split into a follow-up PR to keep this one small and reviewable.

**URL routing (replaces two existing params with a cleaner pair):**
- New `safe_api_url` (the root URL the agent talks to for Safe Transaction Service) and `safe_api_chain_slugs` (chain → slug map). Both env-overridable via `SAFE_API_URL` and `SAFE_API_CHAIN_SLUGS` so the routing target or chain identifiers can be swapped without redeploying.
- Helper `_build_safe_api_url(chain, version, path)` in `LiquidityTraderBaseBehaviour` constructs every Safe API URL from those, raising a clear `KeyError` if a chain has no slug.
- Five call sites migrated: the three Optimism transfer fetchers, the two `safes/<addr>/` Safe-check probes, and the v2 balances fetch (also now accepts an explicit `chain` arg).
- Defaults: `https://rpc-gate.autonolas.tech/safe/tx-service` for the root + `{"optimism":"oeth","base":"base"}` for the slug map.

**Stronger 429 retry in `_request_with_retries`:**
- Honour `Retry-After` (integer-seconds form; clamped to 30s).
- When the header is absent, backoff schedule `(5s, 10s, 15s)` plus `[0, 2)`-second jitter per attempt.
- Wait computation factored into `_compute_rate_limit_wait` for direct unit testing.
- Replaces the hardcoded `sleep(60)` that earlier review flagged as too long and too rigid.

## Out of scope (follow-up PR)

- Use Safe API's `fiat_balance` field instead of re-fetching from CoinGecko in `_calculate_safe_balances_value`.
- New balances cache (kv-keyed by chain) with TTL.
- Per-cycle RPC multicall on whitelisted balances to detect external inflows and invalidate the balances + incoming-transfers caches.
- `PostTxSettlementRound` hook to invalidate balances + outgoing-transfers caches after every successful agent transaction.

## Test plan

- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — all green locally
- [x] `tomte tox -e bandit` — green after the `# nosec B311` on `random.uniform` for the retry jitter
- [x] `tomte tox -e py3.10-darwin` for `TestProxySafePagination`, `short_non_final_*`, and `test_models` — 66 passed
- [x] `autonomy packages lock --check` — verification successful
- [ ] Verify CI green


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -555,11 +569,37 @@ def _get_optimism_balances_from_safe_api(
...
+        try:
+            slug = self.params.safe_api_chain_slugs[chain]
+        except KeyError as exc:
+            raise KeyError(
+                f"No Safe API slug configured for chain {chain!r}; "
+                f"check safe_api_chain_slugs"
+            ) from exc
+        return f"{self.params.safe_api_url}/{slug}/api/{version}/{path}"
+
     def _fetch_safe_balances_with_pagination(
```
OjusWiZard, line 600: 🔴 **Critical — broken tests, not in this PR's diff.** This signature is now `(self, chain, safe_address)`, but `tests/behaviours/test_base.py` was not updated. `TestFetchSafeBalancesWithPagination` calls `_fetch_safe_balances_with_pagination("0xSafe")` with a single positional arg at lines **2837, 2845, 2853** and **4467** → `TypeError: missing 1 required positional argument: 'safe_address'`. Also `_make_behaviour` (test_base.py:63) still sets the removed `params.safe_api_base_url` and never sets `safe_api_url` / `safe_api_chain_slugs`. CI enforces 100% coverage, so this must be fixed before merge: pass `"optimism"` as the first arg at those four sites and update line 63 to the new param names.

packages/valory/skills/liquidity_trader_abci/skill.yaml
```diff
@@ -278,7 +278,8 @@ models:
       store_path: data
       assets_info_filename: assets.json
       pool_info_filename: current_pool.json
-      safe_api_base_url: https://safe-transaction-optimism.safe.global/api/v2/safes
+      safe_api_url: https://rpc-gate.autonolas.tech/safe/tx-service
```
OjusWiZard, line 281: 🟠 **Important — sibling config still points straight at safe.global.** This file, `aea-config.yaml`, and `service.yaml` were updated, but `packages/valory/skills/optimus_abci/skill.yaml` was missed: lines **327/339** still set `safe_api_base_url` and `safe_api_v1_url` → direct `safe-transaction-optimism.safe.global` URLs. AEA deep-merges overrides so this won't crash (the new keys are still supplied here), but it leaves dead keys *and* a direct, proxy-bypassing endpoint in config — the exact thing this PR removes. Replace both keys with `safe_api_url` + `safe_api_chain_slugs` to match.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4064,12 +4063,13 @@ def _fetch_optimism_transfers_safeglobal(
...
-            # and points at safe-transaction-*.safe.global, so under a proxy
-            # only page 1 would go through our infra — page 2 onwards would
-            # bypass it. Building the URL ourselves from safe_api_v1_url
-            # keeps every page going through the configured endpoint.
-            transfers_base_url = f"{base_url}/safes/{address}/incoming-transfers/"
+            # following Safe's response ``next`` URL so every page request
+            # uses the URL we construct from the configured params; the
+            # response ``next`` is absolute and can drop the agent off the
+            # endpoint it was configured to talk to.
+            transfers_base_url = self._build_safe_api_url(
```
OjusWiZard, line 4070: 🟠 **Important — a misconfigured/new chain silently degrades.** `_build_safe_api_url` raises `KeyError` when `chain` is absent from `safe_api_chain_slugs`. All five call sites (this one + 4318, 4401, 4773, 4995) sit inside broad `except Exception` handlers that turn it into `fetch_failed=True` / empty result / stale data / `return False` — indistinguishable from a transient network blip, so a real misconfig looks recoverable and never gets fixed. Today every site hardcodes `"optimism"` (which is configured), so this only bites on a config typo or an added chain, and `safe_api_chain_slugs` gets no validation. Suggest validating it in `Params.__init__` (is-a-dict, non-empty str values, covers `target_investment_chains`) so bad config fails loud at startup instead of mid-round.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1472,6 +1514,36 @@ def _calculate_rate_limit_wait_time(self) -> int:
 
         return 0
 
+    def _compute_rate_limit_wait(self, response: Message, attempt: int) -> int:
```
OjusWiZard, line 1520: 🟠 **Important — new methods have no direct tests.** Neither `_compute_rate_limit_wait` nor `_build_safe_api_url` is exercised directly. Dark branches: Retry-After integer parse, the 30 s clamp, the `ValueError`/HTTP-date fall-through, the schedule index clamp (`attempt ≥ 3`), jitter, and the `KeyError` path. The existing 429 test passes a `MagicMock` as `headers`, so `.splitlines()` returns a mock and the parse branch is never meaningfully hit. With the 100% coverage gate, add `TestComputeRateLimitWait` + `TestBuildSafeApiUrl` (plain synchronous calls — no generator plumbing needed).

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1472,6 +1514,36 @@ def _calculate_rate_limit_wait_time(self) -> int:
...
+        ``RATE_LIMIT_BACKOFF_SCHEDULE`` indexed by ``attempt``, plus
+        a small uniform jitter so two retries from the same starting
+        second don't land at exactly the same later second. Always
+        clamped to ``MAX_RATE_LIMIT_WAIT_SECONDS``.
+
+        :param response: the rate-limited HTTP response.
+        :param attempt: zero-indexed retry attempt counter.
+        :return: sleep duration in seconds.
+        """
+        headers_blob = getattr(response, "headers", "") or ""
```
OjusWiZard: 🟡 **Advisory — silent fall-through on the rate-limit path.** Two spots drop to the backoff schedule with no log: `getattr(response, "headers", "") or ""` here treats a missing/`None` `headers` attribute identically to "no header", and the `except ValueError: break` below (line 1538) silently discards a non-integer Retry-After. If upstream ever sends an HTTP-date Retry-After of e.g. 120 s, it's silently replaced by a 5 s retry — and a framework change to the response shape would be invisible. A `debug`/`warning` log before each fall-through would make these diagnosable.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -82,6 +83,17 @@
 MAX_RETRIES_FOR_ROUTES = 3
 MAX_RETRIES_FOR_STATUS_CHECK = 5
 MAX_SWAP_CONFIRMATION_RETRIES = 60
+
+# Upper bound for any single sleep on a 429 response, whether the wait
+# came from the Retry-After header or from the backoff schedule. Sized
+# to stay well inside FSM round budgets.
+MAX_RATE_LIMIT_WAIT_SECONDS = 30
+
+# Backoff schedule used when the rate-limit response has no Retry-After
```
OjusWiZard, line 92: 🟡 **Advisory — comment vs. code drift.** This says sleep on attempt `i` is `RATE_LIMIT_BACKOFF_SCHEDULE[i] + jitter`, but the implementation clamps `idx = min(attempt, len-1)` (line 1540), so for `attempt ≥ 2` it saturates at the last entry rather than indexing directly. Worth noting the clamp here. Relatedly, with the current `max_retries=3` the 3rd 429 returns before sleeping, so `SCHEDULE[2]` (15 s) is unreachable in practice.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -555,11 +569,37 @@ def _get_optimism_balances_from_safe_api(
         )
         return balances
 
+    def _build_safe_api_url(self, chain: str, version: str, path: str) -> str:
```
OjusWiZard: 🟡 **Advisory (type tightening).** Only `"v1"`/`"v2"` are ever passed — `version: Literal["v1", "v2"]` lets mypy reject a bad call site at no cost (no call changes needed). Minor: the `try/except KeyError` re-raise (just below) is mostly redundant with a plain `dict[chain]` lookup, which already raises `KeyError(chain)`; keep it only if you want the custom message — though if you add the startup validation suggested in `fetch_strategies.py`, this guard becomes belt-and-suspenders.

OjusWiZard (commented):
Automated multi-agent review (code / comments / tests / silent-failures / types / simplify). Direction is sound — the proxy refactor and the backoff logic are correct (I checked the `retries - 1` → schedule-index mapping; it's right). Inline findings below: **1 critical** (broken tests, will fail CI), **3 important**, plus advisories. No security/data-loss issues.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in fbfa143a. Updated `_make_behaviour` to set `safe_api_url` + `safe_api_chain_slugs` (and dropped the removed `safe_api_base_url`). All four `_fetch_safe_balances_with_pagination("0xSafe")` call sites now pass `"optimism"` as the first arg. Verified locally: full test_base.py file plus the new `TestBuildSafeApiUrl` and `TestComputeRateLimitWait` classes all pass.

packages/valory/skills/liquidity_trader_abci/skill.yaml
DIvyaNautiyal07: Fixed in fbfa143a. `packages/valory/skills/optimus_abci/skill.yaml` lines 327 and 339 now use `safe_api_url` + `safe_api_chain_slugs` matching the sibling configs. No dead keys and no direct safe-transaction host left behind.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in fbfa143a. Added shape validation in `Params.__init__` right after `safe_api_chain_slugs` is parsed: must be a dict, every value must be a non-empty string, otherwise raises `ValueError` with the offending chain name and value. A typo like `{"optimism": ""}` or a top-level list now fails loud at agent startup instead of becoming a silent `fetch_failed=True` mid-round.

Didn't cross-check against `target_investment_chains` because Mode is a legitimate target chain without a Safe API slug (the Mode code path uses the explorer API), so a strict subset-check would produce a false positive on Mode-enabled deployments.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in fbfa143a. Added two synchronous unit test classes in `tests/behaviours/test_base.py`:

- `TestBuildSafeApiUrl`: 4 cases covering v1 URL structure, v2 URL structure, different-chain slug substitution, and the `KeyError` path with assertion on the message content.

- `TestComputeRateLimitWait`: 6 cases covering integer-seconds Retry-After (returned directly), Retry-After over the cap (clamped), Retry-After in HTTP-date form (falls through to schedule), no header (walks the schedule + jitter at attempt=0,1,2), attempt past the schedule length (saturates at last entry), and missing-`headers`-attribute (treated as no header). Each test asserts the actual return value or a strict `[base, base+jitter+1)` range so a regression of the parse, the clamp, or the saturate would fail.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Done in fbfa143a. Both fall-through paths now emit a `debug` log:
- Missing/empty `headers` attribute: `"Rate-limit response had no headers; using backoff schedule"`.
- `Retry-After` value that doesn't parse as an integer: includes the offending value (`{value!r}`) so an HTTP-date or other format is visible in logs.

Used `debug` rather than `warning` so a flood of 429s during a burst doesn't spam log warnings, but the diagnostic info is there if anyone goes looking.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in fbfa143a. Comment on `RATE_LIMIT_BACKOFF_SCHEDULE` now reads `RATE_LIMIT_BACKOFF_SCHEDULE[min(i, len-1)] + uniform(0, jitter)` and notes that attempts past the last index saturate at the final entry, matching the actual clamp.

Re: `SCHEDULE[2]` being unreachable with `max_retries=3` — agreed for that single call path. But `max_retries` is a per-caller parameter (defaults to 3) and other call paths could request more, so I kept `SCHEDULE` at three entries so the helper behaves correctly if/when that happens. The new `test_attempt_past_schedule_saturates_at_last` covers the saturation path independent of `max_retries`.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Done in fbfa143a. `version` parameter now typed `Literal["v1", "v2"]`. Mypy will reject a bad call site at compile time as you suggested.

Kept the `try/except KeyError` re-raise on the slug lookup because with the startup validation now in place it really is belt-and-suspenders, but the explicit message names `safe_api_chain_slugs` rather than just the bare key, which made the failure noticeably easier to triage when I was debugging earlier. Happy to drop it if you'd rather have the leaner version.

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -544,7 +544,25 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        # Root URL of the Safe Transaction Service the agent talks to.
+        # Each fetcher derives its endpoint as
+        # ``{safe_api_url}/{slug}/api/{version}/safes/{addr}/...`` where the
+        # per-chain slug comes from ``safe_api_chain_slugs``.
+        self.safe_api_url = self._ensure("safe_api_url", kwargs, str)
+        self.safe_api_chain_slugs = json.loads(
+            self._ensure("safe_api_chain_slugs", kwargs, str)
+        )
+        if not isinstance(self.safe_api_chain_slugs, dict):
+            raise ValueError(
```
OjusWiZard, line 556: 🟠 **Important — these two `ValueError` branches are untested → the 100% coverage gate will fail.** Every test passes a well-formed `safe_api_chain_slugs` (`test_models.py:619` uses `{"optimism":"oeth","base":"base"}`), so neither this raise (556-559) nor the empty-slug raise (562-565) is ever executed. CI enforces 100% coverage on `packages/valory/`, so these uncovered lines block merge — the same gate this validation was added to satisfy. Add two `TestParams` cases:
```python
# non-dict
safe_api_chain_slugs = json.dumps([])  # -> pytest.raises(ValueError, match="must be a JSON object")
# empty / non-str slug
safe_api_chain_slugs = json.dumps({"optimism": ""})  # -> pytest.raises(ValueError, match="non-empty")
```

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -544,7 +544,25 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+                "safe_api_chain_slugs must be a JSON object mapping chain "
+                f"to slug; got {type(self.safe_api_chain_slugs).__name__}"
+            )
+        for chain, slug in self.safe_api_chain_slugs.items():
```
OjusWiZard, line 560: 🟡 **Advisory — validation is shape-only, not coverage.** This checks the dict is well-formed but doesn't verify every chain in `target_investment_chains` has a slug. A config like `target_investment_chains: ["mode"]` without a `"mode"` entry still surfaces as a mid-round `KeyError` from `_build_safe_api_url`, not a startup `ValueError` — the residual half of the original concern. Not exploitable today (call sites hardcode `"optimism"`, which is in the default config), but a one-line cross-check here (`set(target_investment_chains) <= set(self.safe_api_chain_slugs)`) would close it fully. Optional.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_base.py
```diff
@@ -2644,6 +2645,128 @@ def test_existing_cache_invalid_json(self) -> None:
...
+    def test_missing_headers_attribute_treated_as_no_header(self) -> None:
+        """A response object without ``headers`` falls through to backoff."""
+        from packages.valory.skills.liquidity_trader_abci.behaviours.base import (
+            RATE_LIMIT_BACKOFF_SCHEDULE,
+        )
+
+        b = _make_behaviour()
+        # Plain object with no ``headers`` attribute at all.
+        wait = b._compute_rate_limit_wait(object(), attempt=0)
+        assert wait >= RATE_LIMIT_BACKOFF_SCHEDULE[0]
```
OjusWiZard: 🟡 **Advisory — assertion bounds only one side.** This passes for any large value; the sibling cases (`test_no_header_uses_schedule_with_jitter`, `test_attempt_past_schedule_saturates_at_last`) bound both sides. For consistency add the upper bound:
```python
assert RATE_LIMIT_BACKOFF_SCHEDULE[0] <= wait < RATE_LIMIT_BACKOFF_SCHEDULE[0] + RATE_LIMIT_BACKOFF_JITTER_SECONDS + 1
```

OjusWiZard (commented):
Re-review of the post-fix state (HEAD `fbfa143a`). All 7 earlier findings are cleanly resolved. Reviewing what the fix commit added: one merge-blocker (CI coverage) plus two low-priority polish items below. No correctness/security/data-loss issues — the logic is sound.

packages/valory/skills/liquidity_trader_abci/models.py
LOCKhart07, line 556: 🔴 **Blocking — new validation has no negative tests; fails the 100% coverage gate.**

These two `raise ValueError` branches (non-dict slugs, and empty/non-string slug value at line 562) were added in `fbfa143a` to make bad config fail loud at startup. But `test_models.py` was only updated for the happy path (`_make_kwargs`) — neither failure branch is exercised.

`.coveragerc` has `branch = True`, `fail_under = 100`, `source = packages/valory`, and `models.py` is not omitted, so the untaken True-branches and the `raise` lines drop coverage below 100% and the CI gate fails — the same gate the earlier review invoked for the broken-test finding.

Fix: add two `pytest.raises(ValueError)` cases to `TestParams` — one with a non-dict `safe_api_chain_slugs` (e.g. `"[]"`), one with a dict carrying an empty/non-string slug (e.g. `{"optimism": ""}`).

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1472,6 +1517,45 @@ def _calculate_rate_limit_wait_time(self) -> int:
...
+        if not headers_blob:
+            self.context.logger.debug(
+                "Rate-limit response had no headers; using backoff schedule"
+            )
+            headers_blob = ""
+        for line in headers_blob.splitlines():
+            name, _, value = line.partition(":")
+            if name.strip().lower() == "retry-after":
+                try:
+                    return min(int(value.strip()), MAX_RATE_LIMIT_WAIT_SECONDS)
```
LOCKhart07: 🟡 **Advisory — negative `Retry-After` isn't floored at zero.**

`min(int(value.strip()), MAX_RATE_LIMIT_WAIT_SECONDS)` clamps the upper bound but not the lower, so a malformed `Retry-After: -5` yields `-5` → `self.sleep(-5)`. Per spec Retry-After is non-negative so this is low risk, but `max(0, min(...))` is a cheap guard. The schedule path (line 1557) can't go negative, so only this header path is affected.

packages/valory/skills/liquidity_trader_abci/models.py
DIvyaNautiyal07: Fixed in 75600848. Three negative-path tests on ``TestParams``:

- ``test_safe_api_chain_slugs_non_dict_raises`` — `json.dumps([])` fails with `"must be a JSON object"`.
- ``test_safe_api_chain_slugs_empty_string_raises`` — `{"optimism": ""}` fails with `"non-empty"`.
- ``test_safe_api_chain_slugs_non_string_value_raises`` — `{"optimism": 1}` covers the type half of the same guard.

Each uses `pytest.raises(ValueError, match=...)` against the actual message text so a wording change has to update the test explicitly.

packages/valory/skills/liquidity_trader_abci/models.py
DIvyaNautiyal07: Acknowledged but leaving as-is. Mode is a legitimate target chain that doesn't (and can't) have a Safe API slug — the Mode code path in ``_calculate_safe_balances_value`` routes to the Mode explorer API instead. A strict subset check (`set(target_investment_chains) <= set(safe_api_chain_slugs)`) would false-positive on any Mode-targeting deployment.

Could do it with a Mode-shaped exclusion (`set(target) - {"mode"} <= set(slugs)`) but that bakes a chain name into the validation that ought to be data. Happy to add it if the call sites ever stop hardcoding `"optimism"` and start pulling chain from config; until then the loss-of-load is theoretical.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_base.py
DIvyaNautiyal07: Done in 75600848. ``test_missing_headers_attribute_treated_as_no_header`` now bounds both sides: `base <= wait < base + jitter + 1`. The previous one-sided `wait >= SCHEDULE[0]` would have passed for any large value — your point exactly.

packages/valory/skills/liquidity_trader_abci/models.py
DIvyaNautiyal07: Fixed in 75600848. Added three negative-path tests on ``TestParams`` covering the non-dict branch (`json.dumps([])`), the empty-string slug branch (`{"optimism": ""}`), and the non-string-value branch (`{"optimism": 1}`). Each uses `pytest.raises(ValueError, match=...)` against the message text. Both `raise` branches at `models.py:556` and `models.py:562` are now exercised, restoring the 100% coverage gate.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 75600848. ``_compute_rate_limit_wait`` now wraps the Retry-After parse with `max(0, min(...))` so `Retry-After: -5` produces a 0-second sleep rather than `self.sleep(-5)`. Added ``test_negative_retry_after_is_floored_at_zero`` to pin it. Agreed it's spec-non-conforming input, but the guard is cheap and the failure mode (negative sleep argument) would have been fun to debug otherwise.

LOCKhart07 (approved):
Re-reviewed at `75600848`. Both my prior findings are resolved:

- **Coverage gap (models.py validation)** — closed by three falsifiable `TestParams` negative tests (non-dict, empty-string slug, non-string slug value), each hitting a distinct raise branch.
- **Negative `Retry-After`** — floored with `max(0, min(...))` and covered by `test_negative_retry_after_is_floored_at_zero`.

The declined "target-chain ⊆ slugs" check is reasonable — Mode routes to the explorer API, not Safe API, and all `_build_safe_api_url` call sites are Optimism-only. LGTM.

DhairyaPatel7 (approved):
Re-reviewed end to end. Migration is clean — `git grep` confirms no `safe_api_base_url` / `safe_api_v1_url` references remain anywhere in `packages/` (all five call sites plus the sibling `optimus_abci/skill.yaml`). The earlier blocking/important points from @OjusWiZard and @LOCKhart07 (broken tests, sibling config, untested validation branches, negative `Retry-After`) are all addressed with covering tests, and the declined `target_investment_chains` cross-check is reasonable (Mode is a legit target chain with no Safe slug). LGTM.

PR #335, opened 2026-06-03
feat(safe-api): balances cache, RPC change-detect, post-tx invalidation, Safe USD field

DIvyaNautiyal07:
## Summary

Stacked on #334. Second half of the Safe API routing work described in `safe_api_proxy_routing.md`. Once #334 lands the base switches to `hack/launch-base` automatically.

**Cache + freshness**
- New `SAFE_BALANCES_CACHE_TTL_SECONDS = 3 * 3600`. `_get_safe_balances_from_safe_api(chain)` reads `safe_balances_{chain}` + `last_safe_balances_calculated_timestamp_{chain}` from kv at the top of each cycle; fresh cache → no Safe API call. On miss, hits Safe API and persists both.
- `_detect_and_invalidate_on_inflow(chain, safe_address)`: before the cache check, reads native ETH + every `WHITELISTED_ASSETS[chain]` balance via existing RPC helpers. Compares to `last_known_safe_balances_{chain}` snapshot. Any balance strictly greater → external inflow → resets `last_safe_balances_calculated_timestamp_{chain}` and `last_initial_value_calculated_timestamp_{chain}` so the next cycle re-fetches and ROI reflects the new state within one FSM tick.
- `_invalidate_caches_after_tx(chain)`: called from `PostTxSettlementBehaviour` (`chain` comes from `synchronized_data.chain_id`) after every successful settlement. Resets `last_safe_balances_calculated_timestamp_{chain}` and `last_withdrawals_calculated_timestamp_{chain}` so agent-initiated changes (swap / enter / exit / withdrawal / approval / reversion) show up on the next cycle.

**Generic chain dispatch**
- `_get_optimism_balances_from_safe_api` → `_get_safe_balances_from_safe_api(chain)`. Safe address from `safe_contract_addresses[chain]`, URL from the helper.
- Dispatch in `get_positions` and `_calculate_safe_balances_value` routes by membership in `safe_api_chain_slugs` rather than equality against `Chain.OPTIMISM.value`. Optimism and Base both get the Safe API path; Mode keeps the explorer fallback.

**Chain-scoped kv keys**

The codebase iterates every chain in `target_investment_chains` per cycle, so every cache key this PR touches is per-chain. The six keys involved:

| Old key | New key |
|---|---|
| `safe_balances` (new) | `safe_balances_{chain}` |
| `last_safe_balances_calculated_timestamp` (new) | `last_safe_balances_calculated_timestamp_{chain}` |
| `last_known_safe_balances` (new) | `last_known_safe_balances_{chain}` |
| `last_withdrawals_calculated_timestamp` (PR #329) | `last_withdrawals_calculated_timestamp_{chain}` |
| `last_initial_value_calculated_timestamp` (PR #329) | `last_initial_value_calculated_timestamp_{chain}` |
| `optimism_total_withdrawals` (PR #329) | `total_withdrawals_{chain}` |

Producers in `calculate_withdrawals_value` and `calculate_initial_investment_value_from_funding_events` were updated to match the new shape.

**Use Safe's USD field**
- `_get_safe_balances_from_safe_api` carries `fiatBalance` and `fiatConversion` from the Safe API response into each normalised balance entry.
- `_calculate_safe_balances_value` prefers `balance["fiat_conversion"]` as the per-unit price when Safe shipped one. Falls back to `_fetch_token_price` (CoinGecko) when:
  - The field is absent (reward tokens, on-chain backstop entries).
  - The decimal parse fails.
  - The value is non-positive — Safe returns `fiatConversion: "0.0"` for trusted-but-unpriced tokens, so the explicit guard prevents valuing the token at `balance * 0`.

## End-state freshness

| Event | Visible after |
|---|---|
| Agent swap / enter / exit / withdrawal | Next cycle (~6 min, on-tx invalidate) |
| External funding to whitelisted token | Next cycle (~6 min, RPC change-detect) |
| External funding to non-whitelisted token | Up to 3h (TTL safety floor) |
| Nothing changed | Cache served, zero Safe API calls |

## Test plan

- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint -e bandit` — all green locally
- [x] `tomte tox -e py3.10-darwin` — full `test_base.py`, `test_fetch_strategies.py`, `test_post_tx_settlement.py` files pass
- [x] `autonomy packages lock --check` — verification successful
- [x] New test classes covering the new helpers (per the coverage gate)
  - `TestDetectAndInvalidateOnInflow` × 5 (no-change, inflow triggers, None native balance, unparseable snapshot, write failure)
  - `TestInvalidateCachesAfterTx` × 2 (chain-scoped write, write failure)
  - `TestGetOptimismBalances` TTL cache path × 3 (hit-skips-fetch, unparseable timestamp, unparseable cached payload)
  - `TestCalculateSafeBalancesValue` Safe USD branch × 3 (valid use, zero fallback, malformed decimal fallback)
- [ ] Verify CI green
- [ ] Manual smoke test against a funded safe once deployed


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,236 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
+        # If a whitelisted balance went up since last cycle, the helper
+        # invalidates ``last_safe_balances_calculated_timestamp`` so the
+        # cache lookup below falls through to a fresh SafeApi fetch.
+        yield from self._detect_and_invalidate_on_inflow(chain, safe_address)
+
+        # TTL cache: skip the SafeApi fetch if a recent result is on disk.
+        cached_blob: Dict[str, Any] = {}
+        try:
+            read = yield from self._read_kv(
+                ("safe_balances", "last_safe_balances_calculated_timestamp")
```
DhairyaPatel7: 🔴 **Blocking — kv cache keys aren't chain-scoped, but dispatch is now multi-chain.**

`safe_balances`, `last_safe_balances_calculated_timestamp`, and `last_known_safe_balances` (lines 523/567) are all global keys, but `get_positions` and `_calculate_safe_balances_value` now route **every** chain in `safe_api_chain_slugs` through this method. The deployed `service.yaml:98` defaults `target_investment_chains=["base","optimism","mode"]`, so base **and** optimism both hit this path in a single cycle.

With two slug-chains:
- `_detect_and_invalidate_on_inflow` compares one chain's on-chain balances against the *other* chain's `last_known_safe_balances` snapshot → spurious inflow nearly every cycle → resets `last_initial_value_calculated_timestamp` to `"0"` each tick (thrashes the initial-value cache).
- The TTL cache is defeated, and in some read/write interleavings one chain's cached `safe_balances` is served for another.

The pre-rename key was `safe_balances_optimism` (chain-scoped); the rename to the bare `safe_balances` dropped the scoping at the same time dispatch went multi-chain. The PR description's “one cache per agent, since each agent runs against one chain” doesn't hold for the shipped service config.

Fix: scope the balances cache keys per chain — `safe_balances`, `last_safe_balances_calculated_timestamp`, and `last_known_safe_balances` — e.g. `f"safe_balances_{chain}"`, so each chain keeps its own cache + inflow baseline.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,236 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
 
-    def _get_optimism_balances_from_safe_api(
-        self,
-    ) -> Generator[None, None, List[Dict[str, Any]]]:
-        """Get Optimism balances using SafeApi with pagination"""
-        safe_address = self.params.safe_contract_addresses.get("optimism")
-        if not safe_address:
-            self.context.logger.error("No safe address set for Optimism chain")
-            return []
+    def _detect_and_invalidate_on_inflow(
```
DhairyaPatel7, line 531: 🔴 **Blocking — new code paths are untested; the 100% branch-coverage gate will fail.**

`.coveragerc` has `branch = True`, `fail_under = 100`, `source = packages/valory`, and none of `base.py` / `fetch_strategies.py` / `post_tx_settlement.py` are omitted. These additions have no covering tests:

- **`_detect_and_invalidate_on_inflow`** — only ever stubbed (`test_base.py:958`, `:1119`), never driven. Every branch (native/token `None`, unparseable `last_known_safe_balances`, inflow-detected vs not, the three write `except` paths) is uncovered.
- **`_invalidate_caches_after_tx`** — no test references it and `test_post_tx_settlement.py` wasn't touched; at minimum the `except` branch is uncovered.
- **TTL cache-hit block** in `_get_safe_balances_from_safe_api` (the `use_cache` True path, the `json.loads` `except`, and the unparseable-timestamp `warning` branch) — `_make_optimism_b` stubs `_read_kv` to `{}`, so `use_cache` is always False and the hit path never runs.
- **Safe `fiat_conversion` pricing branch** in `_calculate_safe_balances_value` (the `Decimal(...)` path + `InvalidOperation` `except`) — no test seeds `fiat_balance`/`fiat_conversion`.

Please add direct tests for these — same coverage-gate issue the earlier #334 review hit.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2326,17 +2325,34 @@ def _calculate_safe_balances_value(
...
+                # ``fiat_balance`` and ``fiat_conversion`` from the API
+                # response. Reward-token entries (OLAS/VELO) and on-chain
+                # backstop entries don't have these fields, so we still
+                # fall back to ``_fetch_token_price`` for those.
+                token_price = None
+                safe_fiat_balance = balance.get("fiat_balance")
+                safe_fiat_conversion = balance.get("fiat_conversion")
+                if safe_fiat_balance is not None and safe_fiat_conversion is not None:
+                    try:
+                        token_price = Decimal(str(safe_fiat_conversion))
```
DhairyaPatel7, line 2363: 🟠 **`fiatConversion: "0.0"` is treated as a real $0 price and skips the CoinGecko fallback.**

The Safe Transaction Service returns `fiatConversion: "0.0"` / `fiatBalance: "0.0"` for trusted tokens it has no price feed for (the `trusted=true` filter gates spam, not price availability). Here both are non-`None`, so `token_price = Decimal("0")`, the `if token_price is None:` fallback is skipped, and the final `is None` guard also passes → the token is valued at `balance * 0 = $0` instead of falling back to `_fetch_token_price`. That silently under-reports portfolio value / ROI for any held-but-Safe-unpriced token.

Fix: treat a non-positive conversion as “no Safe price”:
```python
if token_price is not None and token_price <= 0:
    token_price = None
```

DhairyaPatel7 (changes requested):

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in e0dd04d8 + d36aa439. The codebase is multichain — `get_positions` and the balance/initial-investment/withdrawal flows iterate every chain in `target_investment_chains` per cycle — so I extended the chain-scoping to the pre-existing PR #329 cache keys too:

- `safe_balances` → `safe_balances_{chain}`
- `last_safe_balances_calculated_timestamp` → `last_safe_balances_calculated_timestamp_{chain}`
- `last_known_safe_balances` → `last_known_safe_balances_{chain}`
- `last_withdrawals_calculated_timestamp` → `last_withdrawals_calculated_timestamp_{chain}`
- `last_initial_value_calculated_timestamp` → `last_initial_value_calculated_timestamp_{chain}`
- `optimism_total_withdrawals` → `total_withdrawals_{chain}`

`_invalidate_caches_after_tx` and `_detect_and_invalidate_on_inflow` both produce the chain-scoped keys; the producers in `calculate_withdrawals_value` and `calculate_initial_investment_value_from_funding_events` were updated to match. PostTxSettlement passes `self.synchronized_data.chain_id`. Tests bulk-renamed.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in e0dd04d8. Added direct test classes for the new helpers:

**`TestDetectAndInvalidateOnInflow`** (5 cases):
- `test_no_change_no_invalidation` — snapshot matches readings, no invalidate write.
- `test_inflow_triggers_invalidation` — current > previous, chain-scoped timestamp set to `"0"`.
- `test_native_balance_none_is_skipped` — `_get_native_balance` returns `None` without crashing.
- `test_unparseable_last_known_treated_as_empty` — garbage in snapshot, treated as empty, still runs.
- `test_write_failure_swallowed` — kv write exception is logged, doesn't propagate.

**`TestInvalidateCachesAfterTx`** (2 cases): chain-scoped balances key + un-chained withdrawals key both set to `"0"`; write-failure swallowed.

**TTL cache-hit path on `TestGetOptimismBalances`** (3 cases):
- `test_ttl_cache_hit_skips_safeapi_fetch` — within-TTL hit returns cached payload without touching SafeApi.
- `test_unparseable_timestamp_treated_as_expired_then_refetches` — bad timestamp falls through to SafeApi.
- `test_unparseable_cached_payload_falls_back_to_refetch` — `json.loads` failure on cached payload falls through.

**Safe `fiat_conversion` pricing on `TestCalculateSafeBalancesValue`** (3 cases):
- `test_safe_fiat_conversion_used_directly` — valid conversion used as price, no CoinGecko call.
- `test_safe_fiat_conversion_zero_falls_back_to_coingecko` — pinned the fix from your other comment.
- `test_safe_fiat_conversion_invalid_decimal_falls_back_to_coingecko` — `InvalidOperation` falls back.

All branches above now exercised — 100% gate should hold.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Real bug, fixed in e0dd04d8. The CoinGecko branch now runs whenever the Safe price is non-positive:

```python
if token_price is not None and token_price <= 0:
    token_price = None
```

Pinned by `test_safe_fiat_conversion_zero_falls_back_to_coingecko` in `TestCalculateSafeBalancesValue`: feeds a balance with `fiat_conversion: "0.0"` and asserts the returned USD value comes from the stubbed CoinGecko ($2.5), not `balance * 0`.

DhairyaPatel7 (approved):
Re-reviewed at `d36aa439`. All three prior findings resolved and verified against the branch:

1. ✅ **Chain-scoped cache keys** — all balances / initial-value / withdrawal keys are now `…_{chain}`; `git grep` confirms no un-scoped variant remains (reads/writes consistent), and `synchronized_data.chain_id` (from `from_chain`) is a chain-name string matching `target_investment_chains`, so `_invalidate_caches_after_tx` hits the same key the reader uses.
2. ✅ **Coverage** — `TestDetectAndInvalidateOnInflow`, `TestInvalidateCachesAfterTx`, `test_ttl_cache_hit_skips_safeapi_fetch`, and `test_safe_fiat_conversion_zero_falls_back_to_coingecko` cover the previously-dark branches.
3. ✅ **`fiatConversion <= 0` fallback** — guard added and pinned by test.

One non-blocking note: `_invalidate_caches_after_tx` keys off the settled tx's `chain_id`, but `calculate_withdrawals_value` only ever reads `last_withdrawals_calculated_timestamp_{target_investment_chains[0]}`. These coincide for single-chain deployments; in a true multi-chain deployment a tx settling on a chain other than `target[0]` won't invalidate the withdrawals cache that's actually read. The balances and initial-value paths don't have this asymmetry (both iterate all chains). Harmless today since withdrawals is effectively `target[0]`-only — flagging for awareness, not blocking.

LGTM.

LOCKhart07 (approved):
Reviewed against a freshly-fetched `hack/launch-base` (which already contains #334). The three prior blocking issues are genuinely resolved: kv keys are chain-scoped throughout (no un-scoped key left in production code), the new helpers have falsifiable tests, and the `fiatConversion: "0.0"` → CoinGecko fallback is guarded + tested.

Also verified the dispatch (`chain in safe_api_chain_slugs` checks dict keys), that `synchronized_data.chain_id` is a chain **slug** so `_invalidate_caches_after_tx` writes the correct `..._{chain}` key, and that the new `post_tx_settlement` call is covered by the existing `async_act` tests.

LGTM. A few **non-blocking** notes for a follow-up:

1. **Inflow detection runs on every call, including TTL cache hits.** `_detect_and_invalidate_on_inflow` does ~6 sequential RPC reads and runs *before* the cache check, so a cache hit still pays the full RPC cost. It's invoked from both `get_positions` and `_calculate_safe_balances_value` each cycle, per safe-api chain (~24 RPC reads/cycle for base+optimism), and `_supplement_with_onchain_whitelisted_balances` re-reads the same whitelisted balances again. The cache saves the paid SafeApi/CoinGecko calls but not the RPC load — worth gating inflow-detection to once per cycle or reusing the readings.

2. **Baseline clobbered on partial RPC failure.** A token whose `_get_token_balance` returns `None` is dropped from `current`, which then overwrites `last_known_safe_balances_{chain}`. Next cycle the token reappears with `previous=0` → spurious inflow → an unnecessary invalidation + refetch. Merging into the prior snapshot instead of overwriting would avoid it.

3. **Dead assignment.** In the `use_cache` branch of `_get_safe_balances_from_safe_api`, `api_success = True` is set but never read afterward — removable.

4. `_invalidate_caches_after_tx` also fires for checkpoint/KPI vanity txs (harmless extra refetch).

None of these block merge.

OjusWiZard (commented):
Automated multi-perspective review (correctness, silent-failures, tests, comments, types, simplification). No critical bugs or security issues — the cache-key migration is clean and the `fiatConversion` fallback never values a token at `balance × 0`. Inline notes below: 2 worth resolving before merge (1, the inflow side-effect; the stale docstrings), the rest advisory.

Verified and **not** reported: the `Dict[str, Any].get()` → `json.loads` path type-checks fine (no mypy error), and `chain_id=None` is unreachable on the `SETTLE` path.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
 
-        if api_success:
-            # Cache fresh data on successful API fetch (including empty — clears stale cache)
+        if inflow_token is not None:
             try:
                 yield from self._write_kv(
-                    {"safe_balances_optimism": json.dumps(all_balances)}
+                    {
+                        f"last_safe_balances_calculated_timestamp_{chain}": "0",
+                        f"last_initial_value_calculated_timestamp_{chain}": "0",
```
OjusWiZard: **Important — inflow detector resets the initial-investment cache on agent-caused increases, not just external deposits.**

`_detect_and_invalidate_on_inflow` compares on-chain balances to the `last_known_safe_balances_{chain}` snapshot and fires on *any* strict increase. It can't distinguish an external deposit from the agent's own activity — after an ETH→USDC swap (or a pool exit returning tokens) the next cycle sees USDC up vs. the pre-tx snapshot, logs `External inflow detected…`, and zeroes `last_initial_value_calculated_timestamp_{chain}`, forcing `calculate_initial_investment_value_from_funding_events` back into a full transfer-history scan.

Correctness is fine (a swap output isn't an incoming transfer, so `total_investment` stays right), but it's wasted work + a misleading log on every token-increasing agent action. Note `_invalidate_caches_after_tx` deliberately *omits* this key (a tx doesn't change initial investment) — which makes resetting it here the inconsistent half.

Suggestion: drop the `last_initial_value_calculated_timestamp_{chain}` reset from this method (its own 30-min TTL still catches genuine external deposits), or at least soften the unconditional "External inflow detected" log since it fires on internal rebalances too.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
-        self.context.logger.info(
-            f"Fetching Optimism balances from SafeApi for safe: {safe_address}"
-        )
+        Reads the agent safe's on-chain balances for native ETH plus the
+        whitelisted tokens on ``chain``, compares to the per-token snapshot
+        persisted in ``last_known_safe_balances``, and if any token's
+        on-chain balance is strictly greater than what we last saw,
+        invalidates two kv timestamps:
+
+        - ``last_safe_balances_calculated_timestamp`` so the next cycle
```
OjusWiZard: **Important — docstring names the pre-rename (un-scoped) kv keys.** The whole point of this PR is chain-scoping, but this docstring still says `last_safe_balances_calculated_timestamp` (line 496) and `last_initial_value_calculated_timestamp` (line 498); the code writes the `…_{chain}` variants (lines 558–559). Same drift at line 491 (`last_known_safe_balances` → `last_known_safe_balances_{chain}`) and the inline comment at line 621. Append `_{chain}` to each.

packages/valory/skills/liquidity_trader_abci/behaviours/post_tx_settlement.py
```diff
@@ -48,6 +48,13 @@ def async_act(self) -> Generator:  # type: ignore[override]
             ):
                 yield from self.fetch_and_log_gas_details()
 
+            # The agent just changed safe state (swap/enter/exit/withdrawal/
+            # approval/etc.) on ``chain_id``, so that chain's kv-cached
+            # balances and the (un-chained) withdrawal value are no longer
```
OjusWiZard: **Important (comment accuracy) — the withdrawal value is no longer "un-chained".** `_invalidate_caches_after_tx` writes `last_withdrawals_calculated_timestamp_{chain}` and `calculate_withdrawals_value` reads the same chain-scoped key, so the `(un-chained)` parenthetical now contradicts the code. Suggest: "… balances and the withdrawal value for that chain are no longer accurate."

packages/valory/skills/liquidity_trader_abci/behaviours/post_tx_settlement.py
```diff
@@ -48,6 +48,13 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+            # accurate. Reset both timestamps so the next FetchStrategies
+            # cycle picks up the new state instead of serving cached data.
+            yield from self._invalidate_caches_after_tx(self.synchronized_data.chain_id)
```
OjusWiZard: **Important (test) — this call is line-covered but has no behavioral assertion.** `test_async_act_not_checkpoint_submitter` drives `async_act` through this line (so the 100% gate passes) but only asserts `set_done.assert_called_once()`; deleting this `yield from` would still pass. Add a `_write_kv` capture and assert `last_safe_balances_calculated_timestamp_optimism == "0"` and `last_withdrawals_calculated_timestamp_optimism == "0"`.

While here: `synchronized_data.chain_id` is `Optional[str]`. It's unreachable-as-None on the `SETTLE` path today, but a one-line guard (or an explicit test) would keep a future caller from silently writing `…_None` keys.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
+        re-fetch instead of serving cached data.
+
+        :param chain: chain identifier of the just-settled tx.
+        :yield: None
+        """
+        try:
+            yield from self._write_kv(
+                {
+                    f"last_safe_balances_calculated_timestamp_{chain}": "0",
+                    f"last_withdrawals_calculated_timestamp_{chain}": "0",
```
OjusWiZard: **Advisory (maintainability) — stringly-typed chain-scoped keys split across files.** `_invalidate_caches_after_tx` (here) *writes* `last_withdrawals_calculated_timestamp_{chain}` and `last_initial_value_calculated_timestamp_{chain}`; their *readers* live in `fetch_strategies.py` (2487, 3178, 3266). Two raw f-strings in different files, no shared definition, no test asserting both sides produce the same string — exactly the missed-rename bug class. This PR got it right; the next one can silently break it. Extract small key-builders (e.g. `_kv_withdrawals_ts(chain)`) in `base.py` and reuse them.

(Related: the read-ts → parse → age-check block is duplicated 3× — base.py ~636, fetch_strategies.py ~2490 and ~3189 — a shared `_cache_age_seconds(ts_key, raw)` helper would unify them.)

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
+                    self.context.logger.debug(
+                        "Failed to cache safe balances to KV store"
+                    )
+            else:
+                # API failed AND the TTL cache wasn't usable; reach for the
+                # last-known persisted payload before the on-chain backstop.
+                if cached_payload:
+                    try:
+                        self.context.logger.warning(
+                            "SafeApi failed, using stale cached balances"
```
OjusWiZard, line 772: **Advisory (test gap) — this stale-fallback branch (API failed *and* cache expired) isn't exercised.** `test_api_failure_falls_back_to_cache` sets `_get_current_timestamp → 0`, giving `cache_age = 0`, so it actually hits the within-TTL path above, not this one. Add a test: timestamp beyond TTL + `_fetch_safe_balances_with_pagination → (False, [])` + assert the stale payload is returned. Separately, `TestDetectAndInvalidateOnInflow` has no ERC-20 (non-native) inflow case — every current case trips on native ETH first.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2326,17 +2325,41 @@ def _calculate_safe_balances_value(
...
+                # and avoids a paid CoinGecko call per token per cycle.
+                # ``_get_safe_balances_from_safe_api`` carries forward
+                # ``fiat_balance`` and ``fiat_conversion`` from the API
+                # response. Reward-token entries (OLAS/VELO) and on-chain
+                # backstop entries don't have these fields, so we still
+                # fall back to ``_fetch_token_price`` for those.
+                token_price = None
+                safe_fiat_balance = balance.get("fiat_balance")
+                safe_fiat_conversion = balance.get("fiat_conversion")
+                if safe_fiat_balance is not None and safe_fiat_conversion is not None:
```
OjusWiZard: **Advisory — `safe_fiat_balance` is read only as a guard, never used.** Price comes solely from `fiat_conversion` (line below); `fiat_balance` (= balance×conversion) is never consumed. If Safe ever returns a valid `fiatConversion` with a null `fiatBalance`, the Safe price is silently dropped and an unnecessary CoinGecko call runs. Change the guard to `if safe_fiat_conversion is not None:` and drop the unused local.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
-        else:
-            # API failed — fall back to cached balances
+                self.context.logger.debug(
+                    "Failed to invalidate cache timestamps after inflow detection"
+                )
+
+        try:
+            yield from self._write_kv({last_known_key: json.dumps(current)})
+        except Exception:  # nosec B110
+            self.context.logger.debug(f"Failed to persist {last_known_key}")
```
OjusWiZard: **Advisory — silent cache-degradation logged at DEBUG.** This `except` (and the post-tx invalidation at line 593, and the TTL-cache read at line 633) swallow KV failures at `debug`. None corrupt values, but each is an invisible degradation in a money-handling path: a persistent snapshot-write failure here means every cycle re-detects the same inflow and re-invalidates (cache effectively off until KV recovers); a failed post-tx invalidation serves pre-tx balances as "fresh" within TTL. Suggest raising these to `warning`/`error` so operators can see why the cache stops working.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
+        cached_payload = cached_blob.get(balances_key)
+        use_cache = (
+            cache_age is not None
+            and 0 <= cache_age < SAFE_BALANCES_CACHE_TTL_SECONDS
+            and cached_payload
+        )
+        if use_cache:
+            try:
+                all_balances = json.loads(cached_payload)
+                api_success = True
```
OjusWiZard: **Advisory (dead store).** `api_success` is only read at line 678, inside the `if not use_cache:` block that reassigns it on line 672. This `api_success = True` on the cache-hit path is never read — safe to remove.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3161,17 +3175,12 @@ def calculate_initial_investment_value_from_funding_events(
...
 
                 if (
                     last_calculated_timestamp
-                    and (
-                        timestamp := last_calculated_timestamp.get(
-                            "last_initial_value_calculated_timestamp"
-                        )
-                    )
+                    and (timestamp := last_calculated_timestamp.get(ii_ts_key))
                     and timestamp is not None
```
OjusWiZard: **Advisory (dead clause).** The walrus `(timestamp := …)` two lines up already short-circuits the `and` chain when the value is falsy (including `None`), so this `and timestamp is not None` can never be reached as `False`. Pre-existing, but it's in the diff context — safe to drop.

OjusWiZard (commented):
Automated multi-perspective review (correctness, silent-failures, tests, comments, types, simplification). No critical bugs or security issues — the cache-key migration is clean and the `fiatConversion` fallback never values a token at `balance × 0`. Inline notes below: 2 worth resolving before merge (1, the inflow side-effect; the stale docstrings), the rest advisory.

Verified and **not** reported: the `Dict[str, Any].get()` → `json.loads` path type-checks fine (no mypy error), and `chain_id=None` is unreachable on the `SETTLE` path.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in bae64d79. Dropped the `last_initial_value_calculated_timestamp_{chain}` reset from `_detect_and_invalidate_on_inflow`. The increase signal can't distinguish an external deposit from agent activity (swap output, pool exit, OLAS reward), and zeroing the initial-investment cache on those was thrashing a recompute that was already correct. The initial-investment TTL still picks up genuine external deposits.

Also softened the log from "External inflow detected" to "Balance increase detected … refreshing safe balances cache" so the operator-facing string matches what the signal actually conveys.

Pinned with `test_inflow_does_not_invalidate_initial_value_cache` — asserts no write payload includes the initial-value key after a positive delta.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in bae64d79. Docstrings on `_detect_and_invalidate_on_inflow` (lines 491/496) and `_invalidate_caches_after_tx` (lines 575+) all now name the chain-scoped keys (`…_{chain}` everywhere they appear). Same treatment for the inline comment at the TTL-cache read site.

packages/valory/skills/liquidity_trader_abci/behaviours/post_tx_settlement.py
DIvyaNautiyal07: Fixed in bae64d79. Comment now reads "that chain's kv-cached balances and withdrawal value are no longer accurate" — dropped the `(un-chained)` parenthetical since `_invalidate_caches_after_tx` writes `last_withdrawals_calculated_timestamp_{chain}` and the producer in `calculate_withdrawals_value` reads the same key.

packages/valory/skills/liquidity_trader_abci/behaviours/post_tx_settlement.py
DIvyaNautiyal07: Both done in bae64d79:

**Behavioural assertion:** `test_async_act_not_checkpoint_submitter` now stubs `_write_kv` to capture each payload and asserts the two chain-scoped reset entries are actually written (`last_safe_balances_calculated_timestamp_optimism: "0"` and `last_withdrawals_calculated_timestamp_optimism: "0"`). Deleting the `yield from self._invalidate_caches_after_tx(...)` line would now fail the test, not just slip through the coverage gate.

**`chain_id is None` guard:** added in `async_act` — the call site now checks `chain_id` is truthy before calling, logs a warning otherwise, and skips invalidation. New `test_async_act_skips_invalidation_when_chain_id_missing` pins this: asserts none of the captured writes have keys ending in `_None`.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Done in bae64d79. Module-level builders for all six chain-scoped keys (`safe_balances_kv_key`, `safe_balances_ts_kv_key`, `last_known_safe_balances_kv_key`, `withdrawals_ts_kv_key`, `total_withdrawals_kv_key`, `initial_value_ts_kv_key`) live alongside the TTL constant in `base.py`. Producers in `fetch_strategies.py` import them; consumers in `base.py` use them; tests assert on the literal strings, so a typo at any call site fails mypy rather than silently writing a parallel key.

Left the read-ts → parse → age-check helper as a follow-up — it's a different shape across the three sites (one has a separate cache-value lookup, another reads kv inside the same call) so unifying them needs more design and isn't a coverage gate.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Both holes closed in bae64d79.

**Stale-fallback (API failed AND TTL expired):** `test_api_failure_with_expired_cache_uses_stale_payload` sets `_get_current_timestamp` to `now`, stores a `last_safe_balances_calculated_timestamp_optimism` that's `now - TTL - 1` seconds old, stubs `_fetch_safe_balances_with_pagination` to return `(False, [])`, and asserts the stale cached payload (`balance=55`) is what comes back. Distinct from `test_api_failure_falls_back_to_cache` which trips on the within-TTL branch.

**ERC-20 inflow:** `test_inflow_on_whitelisted_erc20_triggers_invalidation` seeds the snapshot with native ETH = 0 AND USDC = 0 explicitly, stubs `_get_token_balance` to return 1_000_000 for USDC and 0 otherwise. Asserts the invalidate fires from the USDC delta, not the native ETH path.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Done in bae64d79. Dropped the unused `safe_fiat_balance` local and tightened the guard to `if safe_fiat_conversion is not None:`. A valid `fiatConversion` with a null `fiatBalance` now uses the Safe price instead of silently falling back to CoinGecko.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Done in bae64d79. Raised three kv-failure logs from `debug` to `warning`:

- `_detect_and_invalidate_on_inflow` invalidation write — now warns about the safe-balances cache potentially serving stale data this cycle.
- `_detect_and_invalidate_on_inflow` snapshot persist — now warns that the next cycle will re-detect the same balance increase.
- `_invalidate_caches_after_tx` write — now warns that pre-tx balances may be served as fresh within TTL.
- TTL-cache read in `_get_safe_balances_from_safe_api` — now warns that we'll fetch from SafeApi.
- SafeApi-success persist — now warns that the next cycle will refetch.

Each one was an invisible cache-off-until-kv-recovers state at `debug`; they're operator-facing now.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Removed in bae64d79. The cache-hit `api_success = True` is gone; `api_success` is now initialised to `False` once at the top of the function and only reassigned by `_fetch_safe_balances_with_pagination` on the not-use-cache path. Mypy still types it correctly.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Done in bae64d79. Dropped the redundant `and timestamp is not None` clause from `calculate_initial_investment_value_from_funding_events`. The walrus already short-circuits the `and` chain when the value is falsy (`None` included), so the extra check was dead. Added a short comment noting why so a future contributor doesn't "helpfully" re-add it.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
             try:
-                cached = yield from self._read_kv(("safe_balances_optimism",))
-                if cached and cached.get("safe_balances_optimism"):
-                    self.context.logger.warning(
-                        "SafeGlobal API failed, using cached balances"
-                    )
-                    all_balances = json.loads(cached["safe_balances_optimism"])
-            except Exception:
+                cache_age = now_ts - int(last_ts_raw)
+            except (ValueError, TypeError):
```
OjusWiZard, line 713: **Important (consensus safety) — `positions` now flows through a per-agent, wall-clock-TTL *local* cache, under a same-until-threshold round.**

`GetPositionsRound` is a `CollectSameUntilThresholdRound` (`states/get_positions.py:37`): agents must submit byte-identical `positions` (`GetPositionsPayload`, `get_positions.py:74`) or the round returns `NO_MAJORITY`. `positions` is built from this method.

- **Before:** this fetched Safe API live on every call (cache only on API failure) → all agents saw the same on-chain truth within a round → naturally convergent.
- **After:** it serves from a **per-agent local KV** cache (`_read_kv`/`_write_kv` → `KV_STORE_CONNECTION`, each agent's own store) whenever `cache_age < TTL`, where `now_ts = int(time.time())` (`base.py:2091`) is **per-agent wall-clock**.

So two honest agents in the same round can serve balances captured at different times (one fresh, one up to 3 h stale) → different `positions` → `NO_MAJORITY`. Steady-state is fine (identical stable value), but divergence triggers exist: (1) external inflow + per-agent RPC timing skew — `_detect_and_invalidate_on_inflow` runs against each agent's own node, so one invalidates and refetches while another serves the pre-inflow snapshot; (2) agent restart / cold cache; (3) wall-clock skew across the TTL boundary.

The #329/#334 TTL caches use the same `time.time()` pattern but cache *derived value numbers* — this is the first to put a per-agent cache directly in front of the same-until-threshold payload, where value-equality is mandatory.

Options: drive the TTL off the consensus timestamp (`self.round_sequence.last_round_transition_timestamp.timestamp()`, already used in this file) so hit/miss is identical across agents; or keep the cache only for the non-consensus USD-value path and let `positions` fetch live; or document explicitly why per-agent caching is consensus-safe here (e.g. balances static across a round window) so the next reviewer doesn't have to re-derive it.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +482,239 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
+                        try:
+                            last_known[key.lower()] = int(val)
+                        except (ValueError, TypeError):
+                            continue
+            except (json.JSONDecodeError, TypeError):
+                self.context.logger.warning(
+                    f"Unparseable {last_known_key}; treating as empty"
+                )
+
+        inflow_token: Optional[str] = None
```
OjusWiZard, line 593: **Advisory — the "per-cycle cost is small" justification undercounts: this runs up to 3× per chain per FSM period.** `_get_safe_balances_from_safe_api` (which calls this inflow-detect) is reached from three behaviours per period — `GetPositionsBehaviour.get_positions` (`get_positions.py:60`), `DecisionMakingBehaviour.get_positions` (`decision_making.py:174`), and `FetchStrategiesBehaviour._calculate_safe_balances_value` (`fetch_strategies.py:2285`). Each runs `1 native + len(WHITELISTED_ASSETS[chain])` sequential RPC reads. With `base` now live alongside `optimism`, that's ~`2 chains × 3 calls × ~6 reads ≈ 36 RPC reads/period` just for inflow detection — and the 2nd/3rd passes are near-pure overhead, since the 1st already rewrote `last_known_safe_balances_{chain}` to the current reading so they detect nothing. It also multiplies the per-agent divergence surface in the consensus finding above. Consider hoisting inflow-detection to run once per chain per period (e.g. top of `FetchStrategies.async_act`) instead of inside the shared balances getter.

OjusWiZard (commented):
Second-pass review — two architecture-level findings the first pass didn't cover (consensus determinism + per-period invocation count). New only; the 10 earlier inline notes still stand. Also verified clean this pass: base-chain wiring (`safe_api_chain_slugs` + `WHITELISTED_ASSETS["base"]` align), no surviving old un-scoped kv keys, and the aea-config/service.yaml diff is hash-only.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: You're right that the cache shape isn't consensus-safe for an N>1 service — both the TTL clock (`time.time()` is per-agent) and `_detect_and_invalidate_on_inflow` (per-RPC) can desync the round payload. Optimus runs `number_of_agents: 1` in `packages/valory/services/optimus/service.yaml`, so threshold is 1 and the agent always agrees with itself, which is why `GetPositionsRound` / `DecisionMakingRound` / `FetchStrategiesRound` don't trip in practice. Holding off on the bigger redesign (consensus-timestamp anchoring, removing the per-RPC invalidator) preemptively on a single-agent service, but happy to revisit if Optimus ever moves to N>1.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Confirmed, even on a single-agent service the 3×-per-period multiplier is real waste, and you're right that the 2nd/3rd passes are near no-ops because the 1st already rewrote the snapshot. Fixed in d8fd8f54 by hoisting `_detect_and_invalidate_on_inflow` to a loop at the top of `FetchStrategiesBehaviour.async_act` keyed off `target_investment_chains`, so it runs once per chain per cycle. The three downstream getters keep their TTL cache and post-tx invalidation contract, they just no longer redo the on-chain reads.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -454,15 +501,15 @@ def contract_interact(
...
     def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
-        """Get positions using SafeApi for Optimism, Mode Explorer for Mode"""
+        """Get positions using SafeApi for chains with a slug, Mode Explorer for Mode"""
         all_balances = defaultdict(list)
 
         for chain in self.params.target_investment_chains:
-            if chain == Chain.OPTIMISM.value:
-                # Use SafeApi for Optimism
-                balances = yield from self._get_optimism_balances_from_safe_api()
+            if chain in self.params.safe_api_chain_slugs:
```
OjusWiZard, line 509: **Robustness (important) — `balances` is uninitialized when a chain matches neither branch.** This `if … elif chain == Chain.MODE.value` has no `else` and `balances` isn't initialized before the loop body, yet `if balances:` is read below. For the current config (`optimism`/`base`, both in `safe_api_chain_slugs`) the `if` always matches so it's latent — but a chain that's neither Safe-API nor Mode (e.g. a future mainnet entry added before its slug is wired) makes the read either raise `NameError` (first iteration) or **append the previous chain's balances to this chain** (cross-chain contamination). The PR changed exactly this dispatch. Suggest `balances = []` at the top of each iteration plus an `else` that logs and skips.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -474,53 +521,245 @@ def get_positions(self) -> Generator[None, None, List[Dict[str, Any]]]:
...
+                    "balances cache"
+                )
+                break
+
+        # Atomic write: timestamp reset and baseline-snapshot update go in
+        # the same kv payload. If the write raises, NEITHER lands — the
+        # next cycle re-detects the same delta and re-tries. Splitting the
+        # writes would let the baseline advance on a half-failed cycle and
+        # silently lose the inflow signal for the rest of the TTL.
+        payload: Dict[str, str] = {last_known_key: json.dumps(current)}
```
OjusWiZard: **Correctness (important) — a partial RPC read overwrites the full baseline → spurious invalidation every cycle.** `current` only holds tokens whose reads succeeded (`None` is skipped above), but the baseline is then **replaced** with `current` unconditionally. If one read (e.g. native ETH) times out this cycle, the persisted snapshot loses that key; next cycle the real balance reads `> last_known.get(addr, 0) == 0` → false "inflow" → cache invalidated + Safe API refetched, repeating until a single fully-successful cycle restores the complete snapshot. This silently defeats the PR's quota-saving goal during RPC hiccups and widens the per-agent divergence surface from the consensus thread. Suggest merging instead of replacing: `merged = {**last_known, **current}` so unread tokens keep their prior baseline.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -530,6 +769,8 @@ def _get_optimism_balances_from_safe_api(
                         "asset_type": "native",
                         "address": to_checksum_address(ZERO_ADDRESS),
                         "balance": int(balance),
```
OjusWiZard: **Advisory — `int(balance)` is unguarded against JSON `null`** (here and at the ERC-20 branch ~`:790`). `balance = balance_data.get("balance", "0")` only defaults when the key is *absent*; a present `null` gives `int(None)` → `TypeError` that propagates out of `get_positions`. Unlikely from the live API but reachable via a corrupt **stale-cache** payload. Wrap each `int(balance)` in `try/except (ValueError, TypeError)` and skip the entry with a warning.

OjusWiZard (approved):
Approving — the Safe-API balances cache, post-tx invalidation (chain_id is the slug, so it hits the chain-scoped keys correctly), generic multi-chain dispatch, and the fiatConversion path all look correct, and the earlier review round was thoroughly addressed. Three new robustness notes below (inline); none block merge — finding 2 is the one most worth doing, since it can quietly undermine the cache during RPC flakiness.

_Still-open from before: the per-agent wall-clock-TTL cache consensus-safety thread — fine for the current single-agent service, revisit if Optimus goes N>1._

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 12f3e05c. Initialised `balances = []` at the top of each iteration and added an `else` that warns and `continue`s, so a chain that's neither in `safe_api_chain_slugs` nor Mode is skipped cleanly instead of either crashing (NameError on iter 0) or inheriting the previous chain's balances. Dropped the `# pragma: no branch` on the `elif` since the dispatch now has three real arms. Test `test_chain_dispatch_skips_chain_with_no_handler` pins the contract: an `unknown` chain produces no entry, and the next iteration's chain doesn't carry over its balances.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 12f3e05c with the merge you suggested. Replaced the wholesale write with `merged_snapshot = {**last_known, **current}` so a timed-out token read keeps its prior baseline instead of getting silently dropped and triggering a spurious inflow on the next cycle. Updated the docstring to spell out the invariant so the next reader doesn't re-introduce the wholesale-replace. Test `test_partial_rpc_read_preserves_baseline_keys` sets up a USDC-timeout cycle with ETH flat, asserts USDC retains its prior 999 in the persisted snapshot, and asserts no invalidate timestamp was written.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 12f3e05c. Wrapped the balance parse in `try/except (ValueError, TypeError)` once at the top of the loop, so a present-but-null `balance` (or any non-numeric string, reachable via a corrupt cached payload) is logged and skipped instead of crashing `get_positions`. Both native and ERC-20 append sites now use the same pre-parsed int. Tests `test_skips_entry_with_null_balance` and `test_skips_entry_with_non_numeric_balance` cover both paths.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2326,17 +2348,55 @@ def _calculate_safe_balances_value(
...
+                        token_price = None
+
+                # Prefer Safe's precomputed ``fiat_balance`` for the total:
+                # equal to ``adjusted_balance * token_price`` but skips the
+                # decimals-divide round-trip we already did above. Only
+                # trust it when the Safe price was usable (non-null,
+                # positive) so we don't carry forward a stale or zeroed
+                # total.
+                token_value_usd: Optional[Decimal] = None
+                safe_fiat_balance = balance.get("fiat_balance")
```
OjusWiZard, line 2382: **Advisory (negligible) — asymmetric zero-guard.** The price path just above deliberately treats `fiat_conversion <= 0` as "no Safe price" and falls back to CoinGecko, but here the precomputed total is trusted whenever `safe_fiat_balance is not None`. A Safe-reported `fiat_balance: "0.0"` alongside a positive price would be used as a `$0` total instead of recomputing `adjusted_balance * token_price`. Impact is effectively nil (a `0.0` total with a positive price implies a dust balance worth ≈$0), but for symmetry with the guard above:
```python
if token_value_usd is not None and token_value_usd <= 0:
    token_value_usd = None  # fall back to adjusted_balance * token_price
```
Take it or leave it — not blocking.

OjusWiZard (approved):
Re-approving at 4f80e615. The two follow-up commits (`12f3e05c`, `4f80e615`) cleanly address all second-pass findings — each got a fix **and** a dedicated test:
- `get_positions` / `_calculate_safe_balances_value`: per-iteration `balances` init + `else`/`continue` (`test_chain_dispatch_skips_chain_with_no_handler`)
- inflow baseline: merge instead of replace (`test_partial_rpc_read_preserves_baseline_keys`)
- `int(balance)` guarded (`test_skips_entry_with_null_balance` / `…non_numeric_balance`)
- `_write_kv` falsy-return checked in all invalidation/persist sites (verified `_write_kv` returns a real bool)
- inflow loop now skips non-Safe-API chains (`test_inflow_detector_runs_only_for_safe_api_chains`)

Traced the new `fiat_balance`-as-total optimization end-to-end and it's sound: the on-chain backstop is append-only with no `fiat_balance` key, so Safe-API entries stay self-consistent and backstop entries fall back to recompute. One negligible consistency nit inline.

PR #336, opened 2026-06-04
feat(base): WS3 launch wiring — Aerodrome strategy, slippage check, address fixes

DIvyaNautiyal07:
## Workstream 3 for the Basius launch

Branched off `feat/ws2-remove-reversions` (PR #331), bundles every Base-specific code change WS3 calls for into one PR. PR 3.1 (address audit) and PR 3.4 (whitelist calibration) ran as research; their findings live in `basius_launch.md`.

## Changes

**3.1 — Aerodrome address audit on Base.** Verified each hard-coded Aerodrome address against the official `aerodrome-finance/contracts` README, the `velodrome-finance/slipstream` deployment script, and `velodrome-finance/sugar/deployments/base.env`, cross-checked against Base RPC `eth_getCode` and `factory()` reads.

- **Blocker fix** — `velodrome_non_fungible_position_manager_contract_addresses["base"]` in `optimus_abci/skill.yaml` was the vanity-hex placeholder `0x79e1bEf100000000000000000000000000000000`, which has no bytecode on Base. The wrapper skill overrides the inner skill at runtime, so this was the value production used; any Base CL pool entry would have crashed at the first contract call. Replaced with the live Aerodrome Slipstream V3 NFPM `0xe1f8cd9AC4e4A65F54f38a5CdAfCA44f6dD68b53` (verified: `factory()` returns the official Slipstream V3 PoolFactory `0xf8f2eB4940CFE7d13603DDDD87f123820Fc061Ef`).
- **Consistency fix** — the same param in `liquidity_trader_abci/skill.yaml` had the older Slipstream V1 NFPM. Aligned both skill yamls and the agent aea-config on the V3 address.
- **LpSugar bump** — `velodrome_pools_search.py` `SUGAR_CONTRACT_ADDRESSES["base"]` bumped from `0x27fc7453...` to `0x69dD9db6d8f8E7d83887A704f447b1a584b599A1`. The `all(uint256, uint256, uint256)` signature and the 32-field `Lp` struct field order in `LP_SUGAR_ABI` match the current Vyper source, so the bump is ABI-compatible.
- **Slipstream helper bump** — `velodrome_slipstream_helper_contract_addresses["base"]` bumped from `0x0AD09A66...` to `0x9c62ab10577fB3C20A22E231b7703Ed6D456CC7a`. Bytecode sizes 9350 → 9348 bytes (metadata-hash variance only); the math-helper interface is unchanged.
- **RewardsSugar deliberately not bumped** — V3 returns `LpEpoch (ts, lp, votes, emissions, bribes, fees)` instead of the `EpochData (timestamp, totalLiquidity, votes, emissions, emissionsToken, fees, volume)` shape the consumer decodes by index. Bumping without rewriting the consumer would mis-decode index 1 as the lp address and crash on `float()`. Comment in code records the migration as a follow-up.

**3.2 — register Aerodrome strategy.** `available_strategies["base"]` now includes `velodrome_pools_search` in `service.yaml` and `agents/optimus/aea-config.yaml`. The velodrome custom already branches on `chain == "base"` to use the Aerodrome DEX type and Sugar contracts, so the registration alone unlocks the path.

**3.3 — re-enable Aerodrome slippage protection.** `_enter_stable_volatile_pool` in `pools/velodrome.py` calls `_query_add_liquidity_velodrome` again instead of hardcoding `{amount_a: 0, amount_b: 0}`. TEMP marker dropped. Un-skipped `test_query_returns_none`, which was the canary for this re-enable.

**3.4 and 3.5 — no code change.**
- 3.4 LiFi calibration (USDC ↔ {axlUSDC, msUSD, frxUSD} at $100/$500/$1000/$5000) showed worst-case single-leg slippage of 0.42% (msUSD reverse), comfortably below any reasonable cap. All five Base stables kept in `WHITELISTED_ASSETS["base"]`.
- `slippage_tolerance` is a single global param, so Base inherits whatever Optimism's deployment env sets it to. Decision: match Optimism rather than splitting per-chain.

**3.6 — deferred** to the post-merge E2E (3.7) signal. Optimism-calibrated `max_fee_percentage` / `max_gas_percentage` are global; Base gas is cheap enough that they should pass trivially rather than block. If E2E surfaces a problem, do the chain-keyed split then.

## Test plan

- [x] Full test suite + 100% coverage gate (`tomte tox -e py3.10-darwin`)
- [x] Lint sweep clean: black, isort, flake8, mypy, pylint, darglint, bandit
- [x] `autonomy packages lock` — agent + service hashes updated
- [x] Aerodrome NFPM placeholder verified to have no bytecode on Base via `eth_getCode` against `base-rpc.publicnode.com`
- [x] Replacement Slipstream V3 NFPM verified via `factory()` matching the official Slipstream V3 PoolFactory
- [x] LpSugar `all()` ABI matched against `velodrome-finance/sugar/contracts/LpSugar.vy` field-by-field
- [x] RewardsSugar V3 `LpEpoch` shape diffed against the consumer's index-based decode; bump deferred because incompatible
- [ ] PR 3.7 E2E sign-off against a real funded Base safe before merging Basius launch

Stacked on top of PR #331. Should be reviewed last in the launch sequence.


OjusWiZard (approved):

packages/valory/agents/optimus/aea-config.yaml
```diff
@@ -321,7 +321,7 @@ models:
       selected_hyper_strategy: ${str:max_apr_selection}
       service_endpoint_base: ${str:https://optimism.autonolas.tech/}
       max_pools: ${int:1}
-      available_strategies: ${str:{"mode":["balancer_pools_search","asset_lending","velodrome_pools_search"],"optimism":["balancer_pools_search","uniswap_pools_search","velodrome_pools_search"]}}
+      available_strategies: ${str:{"mode":["balancer_pools_search","asset_lending","velodrome_pools_search"],"optimism":["balancer_pools_search","uniswap_pools_search","velodrome_pools_search"],"base":["balancer_pools_search","velodrome_pools_search"]}}
```
LOCKhart07: `file_hash_to_strategies` (line 318, just above) wasn't updated alongside this `available_strategies` change. Commit 3 synced these strategy hashes to the current per-file fingerprints in `service.yaml`, but the agent config here still carries the **old** CIDs — e.g. velodrome `bafybeicytgrvg2j25…` vs. the current `velodrome_pools_search.py` fingerprint `bafybeihfxuaheqjvy…` (all 5 entries are stale).

These are exactly the hashes commit 3 describes as "no longer matching any `.py` fingerprint → strategy never loads." At deployment `service.yaml` overrides this value, so the live service is fine — but a direct agent run / dev deployment that falls back to this config (without `FILE_HASH_TO_STRATEGIES` set) hits the broken strategy-loading path, so registering `base` strategies here has nothing to dispatch to.

Fix: apply the same 5-hash update to line 318. (Minor: both `skill.yaml` files still have the old `max_apr_selection` hash `bafybeie4sce…` in their 2-entry map — also overridden, optional to align.)

DhairyaPatel7 (approved):

PR #337, opened 2026-06-05
Base chain launch — Safe API, safety checks, Aerodrome portfolio, withdrawals

DIvyaNautiyal07:
## Summary

Launches Base support on the Optimus agent. Three workstreams plus the
Aerodrome-on-Base portfolio fixes that came out of the live launch test.

Setup deltas (README has the canonical version):
`BASE_LEDGER_RPC`, `SAFE_CONTRACT_ADDRESSES` now keys both chains,
`TARGET_INVESTMENT_CHAINS='["base"]'` to run against Base.

The agent runs against one chain at a time. With the default service
config (`["base","optimism","mode"]`), the first entry is what
dispatchers like withdrawal accounting read.

## WS1 — Safe API and funding accounting

- Route Safe API calls through the autonolas proxy, strengthen 429 retry
  with bounded backoff and floored negative `Retry-After`.
- TTL-based kv caches for safe balances, USDC withdrawals, and the
  initial-investment direct read; chain-scoped cache keys throughout so
  Optimism and Base never share state.
- Switch outgoing-USDC transfer dedup to `transferId` (fall back to
  txHash+logIndex) so a same-day withdrawal added after one already
  stored on the same date is not silently dropped.
- Proxy-safe pagination (`offset += len(transfers)`, not `+= limit`) and
  eligible-only early-stop.
- ETH-reversion logic removed; ROI now derives directly from
  `funding_events`.
- Balances cache with RPC change-detect and post-tx invalidation, Safe
  USD field used when available.

## WS2 — Safety checks

- Re-enabled the profitability gate and added a swap slippage cap.
- Stale CL-pool opportunity cache invalidated on period rollover.
- Coverage on the fee/gas threshold identity and period-0 / non-adjacent
  rollover paths.

## WS3 — Base chain launch

- Config, constants, and whitelist for Base. Tier-1 Aerodrome stablecoins
  whitelisted.
- `chain_to_rpc` plumbed into strategies; Base CL pool routing fixed; on-chain
  resolution of `is_stable` when the strategy omits it.
- Funding events on Base flow through the shared SafeGlobal fetcher with a
  `chain` arg.
- Customs test suite kept isolated by not caching `get_web3_connection`.

## Aerodrome portfolio fixes (from the live launch walk-through)

Five independent bugs caught against the funded Base safe
`0xC11c3B9Ed0DB7405394a4A26FF38d8065760B924` while walking the launch
checklist. None of them surface on Optimism because the agent only ever
runs one chain at a time and only against Velodrome positions, so the
chain-agnostic assumptions baked into the helpers below all happened to
hold there.

1. **Aerodrome routed through the Velodrome handler.**
   `calculate_user_share_values` built a `dex_handlers` dict keyed only
   on `DexType.VELODROME.value`, so Aerodrome positions fell through
   `Unsupported DEX type` and contributed $0 to `value_in_pools`. UI
   `dex_type_mapping` gets the matching entry.
2. **Staked Aerodrome LPs read from the gauge, not the pool.**
   `_update_velodrome_position` and `_get_user_share_value_velodrome_non_cl`
   used to call `pool.balanceOf(safe)`, which is 0 for a staked LP.
   `_resolve_velodrome_gauge_address` queries `Voter.gauges(pool)` and
   returns `(resolved, gauge_address)` so a transient Voter RPC failure
   is distinguishable from a genuine "no gauge" answer — callers skip the
   position rather than fall through to `pool.balanceOf` and write a
   known-wrong $0.
3. **Gauge reward symbol per chain.** `_get_velo_family_reward_symbol`
   returns `"AERO"` on Base and `"VELO"` on Optimism / Mode. Routed
   through the three sites that surface the symbol (UI label, the
   "Added X rewards" info log, the CoinGecko id lookup in
   `_calculate_safe_balances_value`).
4. **Base staking subgraph URL.** `staking_subgraph_endpoints["base"]`
   pointed at the Optimism deployment. Every `service(id: X)` query the
   agent made "for Base" was actually pulling Optimism's answer.
   Replaced with the real Base deployment; head block confirmed against
   Base mainnet.
5. **Funding filter classifies the from-address against the right chain.**
   `_should_include_transfer_optimism` did its `eth_getCode` and Safe
   lookup against Optimism regardless of the transfer's chain. A real
   Base transfer of 6.487584 USDC from `0xc84f7c63...` (a contract on
   Base, EOA on Optimism) was wrongly counted as new external capital.
   `chain` is now plumbed through the function, the RPC, the Safe API
   slug, and the per-chain cache key. The withdrawal-side classifier
   (`_is_not_other_contract_optimism`) gets the same treatment.

## Withdrawals on Base

`calculate_withdrawals_value` previously dispatched only on Mode and
Optimism, with everything else returning `Decimal(0)` silently. With the
default service config, `target_investment_chains[0]` is Base, so a
fresh run against Base silently returned 0 withdrawals and inflated
Total ROI by the amount of any outgoing USDC the safe had made.

`_track_erc20_transfers_optimism` and
`_track_and_calculate_withdrawal_value_optimism` now take a `chain` arg
(default `optimism`, so existing call sites stay green) and route the
persistence key, Safe API slug, token-metadata RPC, classifier, and
pricer by chain. The dispatcher now handles `(optimism, base)` via the
same SafeGlobal flow; the `else` branch logs at error so a future L2
added without updating the map is visible.

## E2E proof on the funded Base safe

Same funded Base safe, same period. With the fixes from the WS3 + Aerodrome
+ withdrawal work:

- Portfolio: $2.57 → $17.25 (correct LP value now counted)
- Partial ROI: -82.84% → +15.11%
- Total ROI: -78.95% → +18.99%, then accurate at +41.70% after the §4
  inflow test
- AERO label shows up in `portfolio_breakdown` instead of VELO
- Staking value reads $0.00 instead of the wrong-subgraph $0.58 noise
- Phantom $6.49 contract-funding entry filtered out of `funding_events`
- Inflow detector verified end to end:
  `Balance increase detected on base: token 0x833589f... 1263379 -> 9750963`

## Tests

- 549 tests in `test_fetch_strategies.py` alone, all green locally.
- New coverage for: gauge resolver four return shapes, both staked
  balance callers (gauge balance is read on success, no write on
  resolver failure), reward symbol helper, chain-routed from/to-side
  filters (Base routing, cache namespace separation, unknown-chain
  rejection), and the Base withdrawal dispatch + persistence-key
  isolation + classifier/pricer plumbing.

## Test plan

- [x] `pytest packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py` (549 / 549)
- [x] `autonomy packages lock --check`
- [x] Local lint suite: isort, black, flake8, darglint, mypy, pylint
- [x] Live agent walk-through against funded Base safe, sections §4 / §6
      / §10 / §12 / §13 / §22 of the launch checklist
- [ ] CI green
- [ ] Final ack from reviewers


packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3050,49 +3332,71 @@ def calculate_initial_investment_value_from_funding_events(
...
-        yield from self._save_chain_total_investment(chain, total_investment)
+        # Per-chain totals are written inside ``_calculate_chain_investment_value``;
+        # a single post-loop write would target only the last-iterated chain with
+        # the cross-chain sum and over-count it on the next cache-hit read.
 
         timestamp = int(self._get_current_timestamp())
-        yield from self._write_kv(
-            {"last_initial_value_calculated_timestamp": str(timestamp)}
-        )
+        yield from self._write_kv({initial_value_ts_kv_key(chain): str(timestamp)})
```
OjusWiZard: **Important — timestamp written for only the last chain.** This runs *after* the `for chain in target_investment_chains:` loop, so `chain` holds only the last-iterated value. With a multi-chain config (the `service.yaml` default is `["base","optimism","mode"]`), only the last chain's `initial_value_ts` key is refreshed; the others always look expired → full SafeGlobal history re-fetched and re-priced every cycle, burning API quota. The comment just above correctly avoided a post-loop *totals* write for this exact reason — the timestamp write has the same scoping bug.

Fix: write the per-chain timestamp inside the loop (after `_calculate_chain_investment_value`), or build one payload keyed per iterated chain.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2321,7 +2447,7 @@ def calculate_withdrawals_value(self) -> Generator[None, None, Decimal]:
         """Calculate the value of withdrawals."""
         chain = self.params.target_investment_chains[0]
```
OjusWiZard: **Important — only the first chain's withdrawals are counted.** Its sibling `calculate_initial_investment_value_from_funding_events` iterates *all* `target_investment_chains`, but this hard-indexes `[0]`. For any multi-chain config, withdrawals on the other chains are silently ignored, which **overstates Total ROI** for those chains.

Fix: iterate all chains and sum, mirroring the initial-investment path.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
```diff
@@ -132,18 +131,26 @@ def _check_price_movement(
...
+            # (token1/token0 with decimals embedded) from _get_current_pool_price,
+            # so their relative change is dimensionless. Express it as a percentage
+            # to compare against slippage_tolerance, which the rest of the codebase
+            # treats as a fraction of price.
+            if price_at_selection:
+                price_change_pct = (
+                    abs(current_price - price_at_selection) / price_at_selection
+                ) * 100
+            else:
+                price_change_pct = 0.0
```
OjusWiZard: **Important — falsy `price_at_selection` silently disables the price-movement check.** When `price_at_selection` is `0` (e.g. a prior RPC read returned 0), `price_change_pct = 0.0` → `price_moved = False` → the position is reported as "not moved" forever, suppressing all out-of-range rebalancing for its lifetime. The zero-tick branch a few lines up already returns `None` as the "can't check" sentinel, and the caller guards `is None`.

Fix:
```python
if not price_at_selection:
    self.context.logger.warning("[PRICE_MONITORING] price_at_selection is zero/None; cannot compute change")
    return None
```

packages/valory/customs/velodrome_pools_search/velodrome_pools_search.py
```diff
@@ -3293,6 +3319,17 @@ def run(
...
+    # Apply caller-provided per-chain RPC overrides so the strategy talks to
+    # the same endpoints as the agent's ledger connection. Falls back to the
+    # hardcoded defaults when a chain is not supplied.
+    chain_to_rpc = kwargs.pop("chain_to_rpc", None) or {}
+    if chain_to_rpc:
+        name_to_id = {cname: cid for cid, cname in CHAIN_NAMES.items()}
+        for chain_name, rpc_url in chain_to_rpc.items():
+            cid = name_to_id.get(chain_name.lower())
+            if cid is not None and rpc_url:
+                RPC_ENDPOINTS[cid] = rpc_url
```
OjusWiZard, line 3335: **Important — mutating a module-level global.** `RPC_ENDPOINTS` is a process-wide constant; this overwrites it permanently and never resets. Since the `lru_cache` on `get_web3_connection` was removed in this PR, this mutation is now the only RPC source of truth — a later `run()` without `chain_to_rpc` silently reuses the previous cycle's URLs (or the hardcoded `mainnet.base.org` default for Base), and it leaks across test invocations.

Fix: apply overrides to a local copy and thread it through:
```python
effective_rpc = {**RPC_ENDPOINTS}
effective_rpc[cid] = rpc_url
```

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -1488,11 +1531,15 @@ def _calculate_corrected_yield(
...
+                            reward_symbol, chain
+                        )
                         velo_price = yield from self._fetch_coin_price(velo_coin_id)
                         if velo_price is None:
                             self.context.logger.warning(
-                                "Could not fetch VELO price for yield calculation"
+                                f"Could not fetch {reward_symbol} price for "
+                                "yield calculation"
                             )
                             velo_price = Decimal(0)
```
OjusWiZard: **Advisory — missing VELO/AERO price is persisted as a known-wrong $0.** Falling back to `velo_price = Decimal(0)` undercounts reward value and writes that wrong number into the position's yield. A logged warning doesn't stop the bad value from being stored.

Fix: skip the yield write this cycle (or set `total_yield_usd = None` and guard the store) so the last-known yield is preserved instead of overwritten with an undercount.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -2342,32 +2468,89 @@ def calculate_withdrawals_value(self) -> Generator[None, None, Decimal]:
...
+
+            all_erc20_transfers = yield from self._track_erc20_transfers_optimism(
+                self.params.safe_contract_addresses.get(chain),
+                int(datetime.now().timestamp()),
+                chain=chain,
             )
-            if not all_erc20_transfers_optimism:
+            # ``None`` signals fetcher failure (exception or mid-pagination
+            # error). Skip the kv cache write so a transient blip doesn't
+            # pin ``withdrawal=0`` until the next TTL boundary.
```
OjusWiZard: **Advisory — fetch failure returns `Decimal(0)`, which flows into the ROI numerator as "zero withdrawals."** The comment correctly avoids *caching* the zero, but the returned `Decimal(0)` is still added in `_create_portfolio_data`, so a transient API failure inflates ROI by treating unaccounted withdrawals as if they never happened.

Fix: return `Optional[Decimal]` with `None` = "couldn't fetch", and have the caller skip the ROI calculation rather than assume zero.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -627,12 +662,15 @@ def calculate_user_share_values(self) -> Generator[None, None, None]:
         individual_shares = []
         portfolio_breakdown = []
 
-        # Map DEX types to their handler functions
+        # Map DEX types to their handler functions. Aerodrome on Base shares
+        # the Velodrome code path (same protocol fork, see
+        # VELODROME_FAMILY_DEX_TYPES), so it routes through the same handler.
         dex_handlers = {
```
OjusWiZard, line 668: **Test gap — the new `aerodrome` dispatch is never exercised.** Every position test hardcodes `dex_type="velodrome"`. CI is green at 100% line coverage, but a regression that drops `DexType.AERODROME.value` here (or from `VELODROME_FAMILY_DEX_TYPES`, or renames the `self.pools["aerodrome"]` key used by `_get_velodrome_pending_rewards`) would stay green while Base positions silently produce $0 portfolio value.

Add at least one `dex_type="aerodrome"` / `chain="base"` test asserting it routes through `_handle_velodrome_position` and reads the `"aerodrome"` pool.

OjusWiZard (commented):
Automated multi-perspective review (code-review, silent-failure, tests, comments, types, simplification). Inline comments below; confidence ≥ 80. No Critical issues. Most Important findings are multi-chain correctness bugs that are live in the default `service.yaml` (`["base","optimism","mode"]`) but dormant for a single-chain `aea-config` (`["optimism"]`).

Not posted inline because the lines aren't in this diff:
- **`calculate_user_share_values`** (~`fetch_strategies.py:727`): `except Exception … continue` contributes `Decimal(0)` for a failed position and logs only `str(e)`. A throwing Aerodrome/Base position silently shrinks total portfolio value → wrong rebalancing. Use `logger.exception(...)`.
- **`_create_portfolio_data`** (`fetch_strategies.py:1196`): blanket `except … return {}`; the caller tests `if portfolio_data:` so `{}` (falsy) leaves `self.portfolio_data` stale → ROI gate / sizing / withdrawal detection run on old data. Use `logger.exception(...)` and return `None` so the caller can tell "errored" from "empty".
- **`velodrome.py` error strings** (`:1023, :1209, :1245, :1281`): say `"No CLPoolManager contract address found"` but look up the **NonFungiblePositionManager** — misleads Base debugging.
- **Test gaps**: `_get_safe_balances_from_safe_api` Base TTL-key isolation, `calculate_initial_investment_value_from_funding_events` for `chain="base"`, and `_detect_and_invalidate_on_inflow("base", …)` are all untested.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in e0ccedee. The per-chain timestamp write now lives inside the loop, right after _calculate_chain_investment_value, so every iterated chain refreshes its own initial_value_ts_* key. New TestCalculateInitialInvestmentTimestampPerChain asserts both base and optimism keys land in the kv writes when both are configured.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed. calculate_withdrawals_value now iterates all target_investment_chains via a new _calculate_chain_withdrawal_value helper and sums per-chain results. Each chain still uses its own TTL kv keys (withdrawals_ts_* / total_withdrawals_*). The return type is Optional[Decimal] now (None on any chain's fetcher failure) so _create_portfolio_data can preserve last-known ROI instead of inflating it with a transient zero.

packages/valory/skills/liquidity_trader_abci/pools/velodrome.py
DIvyaNautiyal07: Fixed. A falsy price_at_selection now returns None — the existing 'can't check' sentinel that the caller already guards against — instead of falling through to a 0% change that would mark the position as in-range for the rest of its lifetime and suppress every rebalance.

packages/valory/customs/velodrome_pools_search/velodrome_pools_search.py
DIvyaNautiyal07: Fixed. run() now snapshots the pre-override values for each chain it touches, applies the overrides for one call only via a thin wrapper around _run_impl, and restores in finally. A later cycle without chain_to_rpc cannot reuse the previous cycle's URLs, and a test that crashes mid-run cannot leak overrides into the next test.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed. _calculate_corrected_yield now returns None on a missing VELO/AERO price; the caller preserves the position's last-known yield_usd and cost_recovered fields rather than overwriting them with an undercount that would misreport recovery progress. The existing test_velo_price_fetch_none_returns_none test pins the new contract.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed. calculate_withdrawals_value returns Optional[Decimal] now — None on any chain's fetcher failure. _create_portfolio_data skips the ROI math when withdrawals_value is None and logs that the recompute was skipped this cycle, so a transient API blip preserves the last-known ROI rather than flowing in as zero. value_in_withdrawals is serialized as None in that case too.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Added. TestCalculateUserShareValuesAerodromeDispatch captures the call args on a dex_type=aerodrome / chain=base position and asserts they reach _handle_velodrome_position. A regression that drops DexType.AERODROME.value from the dispatch (or breaks VELODROME_FAMILY_DEX_TYPES) would now fail this test rather than silently routing through 'unsupported dex type'.

DIvyaNautiyal07:
@OjusWiZard thanks for the multi-pass review. The four body items addressed in e0ccedee:

1. \`calculate_user_share_values\` per-position except now uses \`logger.exception(...)\` with the offending pool address included, so a throwing Aerodrome/Base position keeps the traceback rather than collapsing to \`str(e)\`.

2. \`_create_portfolio_data\` returns \`Optional[Dict]\` and uses \`logger.exception(...)\` on the blanket except. The caller's \`if portfolio_data:\` guard still rejects both \`None\` and \`{}\` (so the stale-data preservation behaviour is unchanged), but the intent is now explicit and the traceback is captured.

3. \`velodrome.py\` error strings that said \"No CLPoolManager contract address found\" now read \"No NonFungiblePositionManager contract address found\" so the contract that actually got looked up matches what gets logged. Also fixed the \"Call mint on the CLPoolManager contract\" comment for the same reason.

4. Base routing tests added: \`TestGetSafeBalancesFromSafeApiBaseTtlKeys\` pins \`safe_balances_base\` / \`last_safe_balances_calculated_timestamp_base\` reads and writes (with explicit asserts that no \`_optimism\` key leaks in on a Base call), \`TestDetectAndInvalidateOnInflowBase\` covers the Base-scoped snapshot read and write, and \`TestCalculateInitialInvestmentTimestampPerChain\` asserts both \`last_initial_value_calculated_timestamp_base\` and \`_optimism\` land when both chains are configured (which is the regression that was already silently in place before this commit).

974 tests across the two suites green, full lint + mypy + pylint pass, \`autonomy packages lock --check\` ok.

packages/valory/agents/optimus/aea-config.yaml
```diff
@@ -311,16 +311,17 @@ models:
...
       target_investment_chains: ${TARGET_INVESTMENT_CHAINS:list:["optimism"]}
       staking_chain: ${str:optimism}
-      file_hash_to_strategies: ${FILE_HASH_TO_STRATEGIES:str:{"bafybeie4sceuc3z52jqtnsn5knnsjeogu6gsnetvp325ao2tf4vv3dqcee":"max_apr_selection","bafybeihixbkln3gvgoqml2lboazujtdgzvopeqrgwkfsidhkdj2etrni7a":"balancer_pools_search","bafybeihida2j22vumxowkg5hsz7a7ntkiwv25r3oefwfxecxdrvmsfv6hy":"uniswap_pools_search","bafybeic74xaoypdx3npk6i5h7tadqcsxvdegtuebcic3kz5baqaddqt4vm":"asset_lending","bafybeicytgrvg2j25cqywf37p2yjw726brceutil3f52y2twp4x643axae":"velodrome_pools_search"}}
+      file_hash_to_strategies: ${FILE_HASH_TO_STRATEGIES:str:{"bafybeib2szveecjhwlu5aeww5umtso5sm4btogsrjuqg4tk5uormjmsppi":"max_apr_selection","bafybeiaktss4rvsyiqg43yauc3aj5mq3ztysgqopc2dvxkslswj25kdtcm":"balancer_pools_search","bafybeiglbwfrhxbgqyyg6oniwur2x6sswmi5ed55crzvrhr3o7p5x3kmne":"uniswap_pools_search","bafybeifiotxje5ndkdj673oo2uxfpj76fd32qjp7tpxbbdmidbzcqna23a":"asset_lending","bafybeig67ndkqry2u3mtbuow4tt7g53rngvqgjxy36jwz4wgixtvopjtfa":"velodrome_pools_search"}}
```
OjusWiZard: **Important (new) — stale `velodrome_pools_search` hash; the PR's velodrome/Aerodrome code won't actually run.** This commit bumped the `velodrome_pools_search` package to `bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq` in `packages.json` + `component.yaml`, but `file_hash_to_strategies` here still maps the **old** hash `bafybeig67ndkqry2u3mtbuow4tt7g53rngvqgjxy36jwz4wgixtvopjtfa` → `"velodrome_pools_search"`.

`StrategiesModel.setup()` builds `strategy_to_filehash = {name: hash}` from this map, and `evaluate_strategy.py` loads each strategy executable *by that hash*. So `"velodrome_pools_search"` resolves to the **pre-PR** package — the RPC snapshot fix, the new Sugar/Aerodrome addresses, and the whole Base support in this file are dead code at runtime for any deployment on the default config. (This branch already has prior "sync file_hash_to_strategies" / "stale strategy hashes" fixes — same failure mode.)

Fix: replace the old hash with `bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq` here **and** in `service.yaml:102`.

packages/valory/services/optimus/service.yaml
```diff
@@ -93,34 +93,37 @@ models:
...
       staking_chain: ${STAKING_CHAIN:str:""}
       initial_assets: ${INITIAL_ASSETS:str:{"mode":{"0x0000000000000000000000000000000000000000":"ETH","0xd988097fb8612cc24eeC14542bC03424c656005f":"USDC"}}}
-      file_hash_to_strategies: ${FILE_HASH_TO_STRATEGIES:str:{"bafybeie4sceuc3z52jqtnsn5knnsjeogu6gsnetvp325ao2tf4vv3dqcee":"max_apr_selection","bafybeihixbkln3gvgoqml2lboazujtdgzvopeqrgwkfsidhkdj2etrni7a":"balancer_pools_search","bafybeihida2j22vumxowkg5hsz7a7ntkiwv25r3oefwfxecxdrvmsfv6hy":"uniswap_pools_search","bafybeic74xaoypdx3npk6i5h7tadqcsxvdegtuebcic3kz5baqaddqt4vm":"asset_lending","bafybeicytgrvg2j25cqywf37p2yjw726brceutil3f52y2twp4x643axae":"velodrome_pools_search"}}
+      file_hash_to_strategies: ${FILE_HASH_TO_STRATEGIES:str:{"bafybeib2szveecjhwlu5aeww5umtso5sm4btogsrjuqg4tk5uormjmsppi":"max_apr_selection","bafybeiaktss4rvsyiqg43yauc3aj5mq3ztysgqopc2dvxkslswj25kdtcm":"balancer_pools_search","bafybeiglbwfrhxbgqyyg6oniwur2x6sswmi5ed55crzvrhr3o7p5x3kmne":"uniswap_pools_search","bafybeifiotxje5ndkdj673oo2uxfpj76fd32qjp7tpxbbdmidbzcqna23a":"asset_lending","bafybeig67ndkqry2u3mtbuow4tt7g53rngvqgjxy36jwz4wgixtvopjtfa":"velodrome_pools_search"}}
```
OjusWiZard: **Important (new) — same stale `velodrome_pools_search` hash as `aea-config.yaml:318`.** `file_hash_to_strategies` here still points `"velodrome_pools_search"` at the old hash `bafybeig67ndkqry2u3mtbuow4tt7g53rngvqgjxy36jwz4wgixtvopjtfa`, while `packages.json` was bumped to `bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq`. Update both files so the agent loads the PR's version of the strategy rather than the pre-PR one.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -1060,6 +1128,11 @@ def _create_portfolio_data(
                     self.context.logger.error(f"Error calculating ROI: {str(e)}")
                     total_roi = None
                     partial_roi = None
+            elif withdrawals_value is None:
```
OjusWiZard, line 1131: **Medium (new) — storing `total_roi=None`/`partial_roi=None` here trips `float(None)` downstream.** This branch (correctly) leaves ROI as `None` when withdrawals couldn't be fetched. But `_update_agent_performance_metrics` (~`:5398`) then does `float(self.portfolio_data.get("total_roi", 0.0))` — and `.get` returns `None` (the key is present with value `None`, so the `0.0` default doesn't apply), so `float(None)` raises `TypeError`. It's swallowed by that method's `try/except`, so the effect is: **on every cycle where the withdrawals fetch fails, agent performance metrics silently stop updating** and the dashboard shows stale ROI.

Fix: make the reads `None`-safe, e.g. `v = self.portfolio_data.get("total_roi"); total_roi_percentage = float(v) if v is not None else 0.0` (same for `partial_roi`).

OjusWiZard (commented):
Re-review of the fix commits (`e0ccedee`, `07c45c1a`). All 7 of my prior findings are addressed and resolved — nice work, the multi-chain timestamp/withdrawal fixes and the `Optional`-return propagation are clean and the new tests (aerodrome dispatch, per-chain timestamp) are well-targeted. The `calculate_withdrawals_value` → `_create_portfolio_data` `None`-skip path and the velodrome `try/finally` RPC restore both check out.

The fix introduced a few **new** issues, one of which is significant. Inline comments below; confidence ≥ 80.

**Also worth addressing (not anchored to diff lines):**
- **[Low] `_calculate_corrected_yield` return type is now wrong** (`fetch_strategies.py:1512`): it can `return None` *and* `return total_yield_usd`, but the annotation is still `Generator[Decimal, None, None]`. Should be `Generator[None, None, Optional[Decimal]]`. The runtime guard (`if yield_usd is None:`) is correct, but the annotation tells mypy the return is `None`, which could lead a future refactor to delete the guard as "dead."
- **[Test gaps]** CI enforces 100% line coverage, but a few new behaviors lack a discriminating test:
  - **RPC snapshot/restore** (`velodrome_pools_search.py`): no test asserts `RPC_ENDPOINTS` equals its pre-call snapshot after `run(chain_to_rpc=...)`. Since it's module-global state, this is the exact guarantee the fix adds — please pin it.
  - **Multi-chain withdrawal sum**: `TestCalculateWithdrawalsValue` only uses single-chain configs; add a 2-chain case (`["optimism","mode"]`) where both fetchers succeed and assert the sum. A single-chain test can't regress the original `[0]` bug.
  - **ROI-skip-when-`None`**: no test calls `_create_portfolio_data(initial_investment>0, withdrawals_value=None)` and asserts `total_roi is None`.
  - **Yield preservation**: no end-to-end test that a position's prior `yield_usd` survives a `_calculate_corrected_yield`→`None`.

packages/valory/agents/optimus/aea-config.yaml
DIvyaNautiyal07: Fixed in cd0c4c09. file_hash_to_strategies in aea-config.yaml:318 now points velodrome_pools_search at bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq, matching packages.json. Same update in service.yaml:102 (replied separately on the sibling comment).

packages/valory/services/optimus/service.yaml
DIvyaNautiyal07: Fixed in cd0c4c09. service.yaml:102's file_hash_to_strategies now points velodrome_pools_search at bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq, matching packages.json + aea-config.yaml. Default-config runs will now actually load the PR's strategy code.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed in cd0c4c09. _update_agent_performance_metrics now reads total_roi/partial_roi raw, defaults to 0.0 only when the value is None, and float-coerces otherwise. Good catch — get() with a default doesn't help when the key is present with a None value. Effect was exactly as you described: a withdrawal fetch failure cascaded into a frozen agent-performance dashboard until the next successful cycle. The fix is also covered by TestCreatePortfolioDataSkipsRoiWhenWithdrawalsNone which pins the None ROI flow at the producer end.

DIvyaNautiyal07:
@OjusWiZard thanks for the re-review. Body items addressed in cd0c4c09:

**[Low] Return annotation** — \`_calculate_corrected_yield\` is now \`Generator[None, None, Optional[Decimal]]\` with a docstring that spells out the None contract (caller preserves last-known yield). No mypy noise locally.

**[Test gaps]** — four new test classes, all green locally:

1. \`TestRunRpcOverrideSnapshotRestore\` in \`test_velodrome_pools_search.py\` — pins \`RPC_ENDPOINTS\` equals its pre-call snapshot after both a clean \`run()\` and a mid-\`run\` exception (via \`patch.object(vel_mod, \"_run_impl\", side_effect=RuntimeError)\` so the \`finally\` branch is exercised).

2. \`TestCalculateWithdrawalsValueMultiChainSum\` — Mode+Optimism config returns the sum of both per-chain values; capture asserts both chains were iterated. Plus a sibling test for the one-chain-fail-others-succeed → None contract.

3. \`TestCreatePortfolioDataSkipsRoiWhenWithdrawalsNone\` — calls \`_create_portfolio_data(initial_investment=200.0, withdrawals_value=None)\` and asserts \`total_roi\`/\`partial_roi\` are both None, \`value_in_withdrawals\` is None too.

4. \`TestUpdatePositionPreservesYieldOnPriceMiss\` — end-to-end: drives \`_update_position_with_current_value\` through a \`_calculate_corrected_yield\` that returns None and asserts the position's prior \`yield_usd\` + \`cost_recovered\` fields are untouched.

3048 tests across liquidity_trader_abci + velodrome_pools_search green; lint + mypy + pylint clean; \`autonomy packages lock --check\` ok.

packages/valory/agents/optimus/aea-config.yaml
OjusWiZard: Thanks — almost there, but the map and `packages.json` are out of sync again. This commit bumped the velodrome package hash to `bafybeielyh7i7qjm54h3tne4gcgiuqutqavmczoqc6bguupqnnjzwjtvm4` (it added `tests/test_velodrome_pools_search.py`), yet the map here now points at the *previous* hash `bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq`. The production `velodrome_pools_search.py` is identical between the two, so runtime behaviour is the same — **but** a clean `autonomy deploy` only bundles/pins the hashes in `packages.json` (`elyh7i7…`), so the agent still cannot fetch `gfi4hr35…` and the strategy load fails the same way as before. Please re-point this line **and** `service.yaml:102` at `bafybeielyh7i7qjm54h3tne4gcgiuqutqavmczoqc6bguupqnnjzwjtvm4`. (Longer term, regenerating this map from `packages.json` after every `autonomy packages lock` would stop this recurring drift.)

packages/valory/services/optimus/service.yaml
OjusWiZard: Same residual mismatch as `aea-config.yaml:318`: the map here points at `bafybeigfi4hr35qnlt7ic3gjtnzxqw6vhsdtwvzaduulysmt7jm42d6dgq`, but `packages.json` now publishes `bafybeielyh7i7qjm54h3tne4gcgiuqutqavmczoqc6bguupqnnjzwjtvm4` for the velodrome package (the test addition in this commit re-bumped the hash). Please update this line to `elyh7i7…` too so the bundled/pinned hash and the strategy map agree.

OjusWiZard (approved):
Approving — all review findings across both rounds are addressed and verified at head `e6433fab`.

**Round 1 (7 findings):** multi-chain initial-investment timestamp (now written per-chain inside the loop), multi-chain withdrawals aggregation (`Optional`, `None`-skips ROI), `price_at_selection==0` returns `None`, `RPC_ENDPOINTS` snapshot/restore, VELO/AERO price-miss preserves last-known yield, per-position `logger.exception`, and an Aerodrome-dispatch test.

**Round 2 (3 findings):** the stale `velodrome_pools_search` strategy hash is now synced to the bundled package hash (`bafybeielyh7i7…`) across `packages.json`, `aea-config.yaml`, and `service.yaml` (so the agent loads the PR's strategy, not a pre-PR one); `_update_agent_performance_metrics` is `None`-safe for ROI; and the `_calculate_corrected_yield` annotation is corrected. The four new test classes (RPC snapshot/restore, multi-chain withdrawal sum, ROI-skip-on-`None`, yield-preservation) are all discriminating.

Non-blocking nit for a future change: `TestUpdatePositionPreservesYieldOnPriceMiss` could add an `assert "current_value_usd" in position` canary to unambiguously distinguish the `None`-path from the `except` path. Not required for merge.

Nice work on the fixes. 🚀

LOCKhart07 (approved):

PR #338, opened 2026-06-05
fix(base): plug five gaps in Aerodrome portfolio valuation, rewards, and funding on Base

DIvyaNautiyal07:
## Summary

Five independent bugs that all silently mis-stated portfolio/ROI for Aerodrome on Base. Each was caught live against the funded Base safe (`0xC11c3B9Ed0DB7405394a4A26FF38d8065760B924`) while walking the launch checklist. None of them surface on Optimism because Optimus only ever runs one chain at a time and only against Velodrome positions, so the chain-agnostic assumptions baked into the helpers below all happened to hold there.

### 1. `c62c9b7b` — Aerodrome positions never entered share-value math

`calculate_user_share_values` built a `dex_handlers` dict keyed only on `DexType.VELODROME.value`. Aerodrome positions on Base fell through to the `Unsupported DEX type` branch, so `value_in_pools` was always 0 and Total ROI showed the safe balance as the entire portfolio. Fix routes Aerodrome through the same `_handle_velodrome_position`. UI `dex_type_mapping` gets the matching `aerodrome -> "aerodrome"` entry.

### 2. `e0b8bee1` — Staked Aerodrome LPs were read at the pool, not the gauge

Two sites (`_update_velodrome_position` and `_get_user_share_value_velodrome_non_cl`) called `pool.balanceOf(safe)` which returns 0 for staked positions because the LP tokens are at the gauge. The first site then tripped `check_and_update_zero_liquidity_positions` into marking the position CLOSED. The second collapsed the per-token contribution to zero. Added `_resolve_velodrome_gauge_address` that queries `Voter.gauges(pool)` and switched both sites to read from `VelodromeGaugeContract.balance_of` when `position.staked` is true. Also reverted the dead Tenderly virtual URL default for `ledger_apis.base.address` and made it env-overridable so `.env BASE_LEDGER_RPC` actually flows through.

### 3. `293ddcc9` — Gauge reward token mislabeled as VELO on Base

`_handle_velodrome_position`, the fallback CoinGecko price lookup in `_calculate_safe_balances_value`, and the yield-log line all hardcoded `"VELO"` even though `_get_velo_token_address` correctly returns the AERO address on Base. Added `_get_velo_family_reward_symbol(chain)` helper that returns `"AERO"` for Base and `"VELO"` elsewhere, and use it at the three sites.

### 4. `cf68564b` — Base staking subgraph URL pointed at Optimism

`staking_subgraph_endpoints["base"]` had the same subgraph ID as `["optimism"]`. Verified live: the configured deployment's head block matched Optimism mainnet, not Base mainnet. Every `service(id: X)` query the agent made "for Base" was actually pulling Optimism's answer for that id — fabricating rewards (when the id existed there) or zeroing them (when it didn't). Replaced with the real Base staking subgraph deployment `9etc5Ht8eQGghXrkbWJk2yMzNypCFTL46m1iLXqE2rnq`. Verified head block lines up with Base mainnet.

### 5. `f3adbd6c` — Funding filter wrongly counted Base contract transfers as EOA

`_should_include_transfer_optimism` did its EOA-vs-contract `eth_getCode` and its Gnosis-Safe lookup against Optimism regardless of the transfer's actual chain. A real Base transfer of 6.487584 USDC from `0xc84f7c63742EA1894EE04e5F49fbaE8C3a4a734d` (a contract on Base, EOA on Optimism) was wrongly counted as new external capital. Parametrized the helper with a `chain` arg (default `"optimism"` so existing call sites stay green), selected `self.params.optimism_ledger_rpc` or `self.params.base_ledger_rpc` per chain, routed the Safe lookup through `_build_safe_api_url(chain, ...)`, and scoped the contract-check cache key by chain.

## E2E proof of impact

Same funded Base safe, same period. With the fixes:

- Portfolio: $2.57 → $17.25 (correct LP value now counted)
- Partial ROI: -82.84% → +15.11%
- Total ROI: -78.95% → +18.99%, then accurate at +41.70% after the §4 inflow test
- AERO label shows up in `portfolio_breakdown` instead of VELO
- Staking value reads $0.00 instead of the wrong-subgraph $0.58 noise
- Phantom $6.49 contract-funding entry filtered out of `funding_events`
- Inflow detector verified end to end: `Balance increase detected on base: token 0x833589f... 1263379 -> 9750963`

## Test plan

- [x] `pytest packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_fetch_strategies.py` (509 / 509)
- [x] `autonomy packages lock --check`
- [x] Live agent run against funded Base safe, walked §4/§6/§10/§12/§13/§22 of the launch checklist
- [ ] CI green


LOCKhart07 (approved):
All five fixes check out. I verified the contract interfaces (`VelodromeVoterContract.gauges(pool_address)→{gauge}`, `VelodromeGaugeContract.balance_of(account)→{balance}`), confirmed `position` is in scope at both staked sites, that `DexType.AERODROME` exists on the base branch, that the funding-filter `chain` routing only ever sees `optimism`/`base` (Mode has its own `_should_include_transfer_mode` path), and that the optimism cache key stays back-compatible. CI is green.

Two non-blocking notes:

**1. New logic ships without tests.** ~130 lines added, no test files touched, and this branch's CI has no coverage gate. `_resolve_velodrome_gauge_address`, both `if position.get("staked")` gauge branches, `_get_velo_family_reward_symbol`, and the `chain="base"` path of `_should_include_transfer_optimism` have no dedicated assertions — the existing `"staked": True` tests mock the *old* `pool.get_gauge_address`, not the new Voter/Gauge path. Repo policy is 100% coverage (CLAUDE.md), so `tox -e coverage` would likely flag these. Worth a focused test asserting a staked position reads `gauge.balanceOf` (not the pool) and that a Base transfer is classified via the Base RPC + Base Safe API.

**2. `base_ledger_rpc` skill param ignores `BASE_LEDGER_RPC` (minor).** §2 made `ledger_apis.base.address` env-overridable (`${BASE_LEDGER_RPC:str:...}`), but the skill param at `aea-config.yaml:334` is `${str:https://mainnet.base.org}`. The §5 funding filter's `eth_getCode` runs through `self.params.base_ledger_rpc`, so setting `BASE_LEDGER_RPC` to a private endpoint applies it to `check_balance` but not the funding filter, which still hits public `mainnet.base.org`. Matches the pre-existing optimism param so it's not a regression, but inconsistent with §2's stated goal — consider `${BASE_LEDGER_RPC:str:...}` here too.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3038,6 +3091,39 @@ def _update_velodrome_position(
                     f"Failed to get balance for Velodrome pool position: {position}"
                 )
 
+    def _resolve_velodrome_gauge_address(
```
OjusWiZard, line 3146: **Important - silent $0 on transient gauge lookup failure + no test coverage.**

Voter addresses are configured for all three chains, so misconfig isn't the real risk - a transient `gauges()` RPC failure is. On failure this returns `None`, and both callers then fall back to `pool.balanceOf(safe)` which is `0` for a staked position, so the position is silently valued at $0 with only a `warning`. `contract_interact` collapses "RPC error" and "genuine ZERO_ADDRESS / no gauge" into the same `None`, so they can't be distinguished. Consider a sentinel so a *failed* gauge read on a staked position logs at `error` / skips rather than writing a known-wrong 0.

This method (and its branches) is also **completely untested** - add cases for: no voter configured, `ZERO_ADDRESS` gauge, and a valid gauge.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -3016,16 +3053,32 @@ def _update_velodrome_position(
...
-                performative=ContractApiMessage.Performative.GET_RAW_TRANSACTION,
-                contract_address=pool_address,
-                contract_public_id=VelodromePoolContract.contract_id,
-                contract_callable="get_balance",
-                data_key="balance",
-                account=safe_address,
-                chain_id=chain,
-            )
+            balance = None
+            if position.get("staked"):
```
OjusWiZard, line 3098: When `staked` is true but `_resolve_velodrome_gauge_address` returns `None`, this falls through to the pool `get_balance` below - which is `0` for a staked position - and writes `current_liquidity = 0`. See the note on `_resolve_velodrome_gauge_address`: a transient voter RPC blip silently zeroes the position. No test exercises this staked branch (existing `test_non_cl_*` cases omit the `staked` key).

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -1901,15 +1911,37 @@ def _get_user_share_value_velodrome_non_cl(
...
-            account=user_address,
-            chain_id=chain,
-        )
+        # Staked positions hold their LP tokens at the gauge, not at the safe.
+        # The pool's balanceOf(safe) returns 0 in that case and the share
+        # calculation collapses to zero, leaving the position out of
+        # value_in_pools. Resolve the gauge via the Voter and read the
+        # staked balance there when the position is marked staked.
+        user_balance = None
+        if position.get("staked"):
```
OjusWiZard, line 1921: Same staked-fallback caveat as in `_update_velodrome_position`: gauge-resolution `None` -> pool `check_balance` = 0 -> `user_share = 0` -> position contributes $0 to `value_in_pools`. Untested - every `TestGetUserShareValueVelodromeNonCl` case omits the `staked` key, so this whole branch is never entered.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4262,9 +4349,24 @@ def _fetch_optimism_transfers_safeglobal(
         return not fetch_failed
 
     def _should_include_transfer_optimism(
-        self, from_address: str
+        self, from_address: str, chain: str = "optimism"
```
OjusWiZard, line 4417: **Important - the symmetric withdrawal-side filter was left un-parametrized.**

This fixes the `from_address` (incoming) path, but the sibling `_is_not_other_contract_optimism(to_address)` - called from `_track_and_calculate_withdrawal_value_optimism` - still hardcodes `self.params.optimism_ledger_rpc` and an `optimism_` cache-key prefix. That's the *exact* bug class this commit fixes, left in place for outgoing/withdrawal destinations: a Base withdrawal target gets classified against Optimism state and shares Optimism's cache namespace. Either give it the same `chain` arg or document why withdrawals are Optimism-only.

Minor: the docstring's "defaults to optimism so the old call sites keep working unchanged" is misleading - the only caller was updated in this same PR to pass `chain=chain`; there are no pre-existing callers relying on the default.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4287,8 +4390,19 @@ def _should_include_transfer_optimism(
             except (json.JSONDecodeError, KeyError):
                 pass
 
+        chain_to_rpc = {
```
OjusWiZard, line 4457: Advisory (latent, not reachable today): for any chain not in `{optimism, base}`, `rpc_url` is `None` and the method returns `False`, silently *excluding* every transfer from that chain and understating initial investment. The dispatch guard upstream keeps this safe for now, but a future third Safe-backed L2 added without updating this map would silently drop its funding events. A cross-reference comment or an `error`-level log would make the gap visible.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -4904,6 +5023,19 @@ def _get_velo_token_address(self, chain: str) -> Optional[str]:
         velo_addresses = self.params.velo_token_contract_addresses
         return velo_addresses.get(chain)
 
+    def _get_velo_family_reward_symbol(self, chain: str) -> str:
```
OjusWiZard, line 5130: Advisory: the `chain == "base"` bare literal silently returns `"VELO"` for any other chain - fine today, but prefer `Chain.BASE.value` for consistency with the rest of the file and to avoid drift. Also untested: no case passes `chain="base"` to reach the `"AERO"` branch, so the wrong-symbol -> CoinGecko-lookup path on Base is unverified.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -1287,10 +1293,10 @@ def _handle_velodrome_position(
             position.get("token1"): position.get("token1_symbol"),
         }
 
-        # Add VELO to token_info if rewards exist
+        # Add the gauge reward token (VELO / AERO) to token_info if rewards exist
         velo_token_address = self._get_velo_token_address(chain)
         if velo_token_address and velo_token_address in user_balances:
-            token_info[velo_token_address] = "VELO"
+            token_info[velo_token_address] = self._get_velo_family_reward_symbol(chain)
```
OjusWiZard, line 1300: Nit: the routing here is correct, but the staked-rewards `logger.info("Added VELO rewards to position: ...")` a few lines above (~1287) still hardcodes "VELO" and will say VELO even on Base/Aerodrome, contradicting this AERO labeling. Route that log string through `_get_velo_family_reward_symbol(chain)` too.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -660,12 +662,15 @@ def calculate_user_share_values(self) -> Generator[None, None, None]:
...
-        # Map DEX types to their handler functions
+        # Map DEX types to their handler functions. Aerodrome on Base shares
+        # the Velodrome code path (same protocol fork, see
+        # VELODROME_FAMILY_DEX_TYPES), so it routes through the same handler.
         dex_handlers = {
             DexType.BALANCER.value: self._handle_balancer_position,
             DexType.UNISWAP_V3.value: self._handle_uniswap_position,
             DexType.STURDY.value: self._handle_sturdy_position,
             DexType.VELODROME.value: self._handle_velodrome_position,
+            DexType.AERODROME.value: self._handle_velodrome_position,
```
OjusWiZard, line 673: Routing Aerodrome through the Velodrome handler is correct (verified against `VELODROME_FAMILY_DEX_TYPES`). Overall the PR adds two new methods and four new branches with **no accompanying tests** - worth noting the repo's stated 100% coverage policy isn't gating this PR (the CI `test` jobs run `tox -e py3.x`; the coverage gate `tox -e coverage` isn't a required check). For a valuation/ROI fix, adding tests for the new paths is the highest-value follow-up.

OjusWiZard (commented):
Automated multi-pass review (correctness / silent-failure / tests / comments). The five fixes are correct and well-targeted — I verified the gauge & voter `data_key`s match the contract wrappers and the incoming-transfer chain parametrization is sound. No blocking bug; CI is green. Inline notes below are about **incompleteness and missing tests**, not regressions. The staked->pool fallback is *not* a regression: pre-PR, staked positions always read the pool and got 0, so this strictly improves that.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Done in e9d514e2. Changed the resolver to return (resolved, gauge_address) so a transient Voter RPC failure shows up as resolved=False, distinct from ZERO_ADDRESS or no-voter-configured which both come back as (True, None). Both staked callers now log at error and skip the position on any unresolved or empty gauge rather than falling through to pool.balanceOf and writing a wrong $0. Tests cover all four resolver paths plus the staked branches in _update_velodrome_position and _get_user_share_value_velodrome_non_cl.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Fixed. _update_velodrome_position now returns early on any unresolved or None gauge for a staked position rather than falling through to pool.balanceOf. Added TestUpdateVelodromePositionStaked covering the gauge-balance read, the transient resolver failure, and the no-voter case.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Same fix on this side. _get_user_share_value_velodrome_non_cl returns {} when the resolver fails or yields no gauge for a staked position. New TestGetUserShareValueVelodromeNonClStaked covers the gauge-read happy path and the resolver-failure-returns-empty case.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Parametrized _is_not_other_contract_optimism with a chain arg (default optimism), now routes the eth_getCode endpoint, the Safe API URL, and the cache key by chain. The misleading old-call-sites line is removed from the from-side docstring. Tests cover Base routing, the cache-namespace separation (a cached optimism verdict does not satisfy a base lookup), and unknown-chain rejection.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Escalated the unknown-chain branch on both filters from warning to error with a short comment about why, so a future Safe-backed L2 added without updating the RPC map is visible rather than silently excluded. Tests on both sides assert the path returns False and never issues an RPC call.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Swapped the bare base literal for Chain.BASE.value. Added TestGetVeloFamilyRewardSymbol covering Base AERO and Optimism/Mode VELO. The CoinGecko-id lookup site at the caller (2332-ish) reads the symbol from this helper, so pinning the mapping at the helper covers what feeds into that lookup.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Routed the staked-rewards info log through _get_velo_family_reward_symbol(chain) so it reads AERO on Base and VELO on Optimism/Mode, matching the routing two lines below.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Added tests for all the new paths in this PR: the gauge resolver (4 return shapes), both staked balance callers (gauge balance is read on success, no write on resolver failure), the reward-symbol helper, and the chain-routed from/to-side filters (Base routing, optimism cache does not satisfy a base lookup, unknown chain rejected without an RPC call). 20 new tests, 529 to 549 total, full file green locally.

DIvyaNautiyal07:
@LOCKhart07 thanks for the review. Both notes addressed in e9d514e2:

1. New logic is now covered. 20 new tests across 6 classes pin the gauge resolver's four return shapes, both staked balance callers (gauge balance is read on success, no write on resolver failure), the reward symbol helper, and the chain-routed from/to-side filters (Base routing, cache namespace separation, unknown-chain rejection). File goes from 529 to 549 tests, all green locally.

2. aea-config base_ledger_rpc is now \`${BASE_LEDGER_RPC:str:https://mainnet.base.org}\`, matching service.yaml and the §2 ledger_apis change, so the funding-filter eth_getCode honors BASE_LEDGER_RPC alongside check_balance.

OjusWiZard (approved):

LOCKhart07 (approved):
Re-reviewed `e9d514e2`. All of the second-round notes (the gauge-resolver sentinel, both staked-branch skips, the AERO staked-rewards log, `Chain.BASE.value`, the escalated unknown-chain logs, `base_ledger_rpc` env-override) are correctly addressed, and the 20 new tests are meaningful rather than mock-asserting — e.g. `test_uses_gauge_balance_not_pool_balance` forces the pool read to `None` so the assertion only passes via the gauge path, and `test_optimism_cache_does_not_satisfy_base_lookup` proves the base lookup still hits the RPC despite an optimism cache entry. CI is fully green.

One non-blocking scope question, not a regression: the to-side `_is_not_other_contract_optimism(chain=...)` parametrization is correct and unit-tested, but it's currently unreachable for Base in production — `calculate_withdrawals_value` only branches on `mode`/`optimism`, so a Base deployment hits `else: return Decimal(0)` and the helper is never called with `chain="base"` outside tests. Net effect is that Base Total-ROI ignores withdrawals entirely. That's probably deliberate for this launch (the E2E proof covers inflows, not outflows), but worth confirming it's intentional rather than a half-wired path. Approving.

PR #340, opened 2026-06-12
fix(staking): suppress vanity tx when EOA balance is below observed real-tx cost

DIvyaNautiyal07:
## Summary
- Staking-KPI behaviour currently pads the multisig nonce with a zero-value vanity tx whenever the agent falls behind on activity. Vanity txs cost orders of magnitude less than real swaps, so an EOA that cannot fund actual on-chain work can still keep the KPI green. The funding alert fires but staking rewards keep flowing, so the symptom is masked. Pearl support thread caught this when a user's agent had been frozen on an out-of-range Velodrome CL position for ~40 days while reporting normal KPI.
- Before `_prepare_vanity_tx`, the behaviour now reads the agent EOA balance from the funds_manager shared-state hook (same source as `/funds-status`) and the recent real-tx cost from the `GasCostTracker` that `post_tx_settlement` already populates from settled non-vanity tx receipts. If balance is below the median of the last five recorded real-tx costs on the staking chain, the vanity tx is skipped and a WARNING with concrete numbers is logged.
- Either signal missing (no tracker history yet, hook not registered, multicall failed, agent EOA absent from response) falls open to current behaviour, so a transient lookup error never silently kills staking activity.

## Grounding
The cost threshold is anchored on the actual `gasUsed * effectiveGasPrice` from settled txs in the on-disk tracker, so the gate self-tunes across gas regimes without a config knob. Verified the chokepoint at `post_tx_settlement.py:72` is the only place real-tx costs are recorded (liquidity_trader swaps route through `ContractApiMessage.GET_RAW_TRANSACTION` → Safe settlement, no Python-level `eth.estimate_gas`). Verified the affected user's funds-status snapshot (`balance=6.92e13`) is below their observed real-tx cost (`1.55e14` across 4 attempts) so the gate would have fired on his agent.

Median-of-five rather than latest-only so a single gas-price spike does not strand the agent. Tracker keeps 20 records so the window is well-bounded.

## Test plan
- [x] `pytest packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py` — 4 new cases (suppression fires on balance < cost, vanity tx runs on balance ≥ cost, vanity tx runs when tracker empty, vanity tx runs when funds-status hook missing) plus the existing 12 still pass
- [x] `autonomy packages lock --check` passes
- [x] Packages pushed to IPFS so CI sync resolves them
- [ ] Spot-check on a service: confirm a healthy EOA produces no WARNING and vanity tx still lands
- [ ] Spot-check on a starved EOA (or a service forced into that state): confirm the WARNING fires with balance/cost numbers and KPI starts failing

## Notes for reviewers
- No new config params. The operator's existing `FUND_REQUIREMENTS` knobs are untouched; this PR layers an additional, dynamic gate on top.
- Adds `valory/funds_manager` to `liquidity_trader_abci/skill.yaml` skills list because the behaviour now imports `GET_FUNDS_STATUS_METHOD_NAME` from it. funds_manager is already loaded by `agents/optimus/aea-config.yaml` so this is a packaging declaration, not a runtime wiring change.


packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +156,59 @@ def async_act(self) -> Generator:  # type: ignore[override]
             yield from self.wait_until_round_end()
             self.set_done()
 
+    def _real_tx_cost_vs_balance(
+        self, chain: str
+    ) -> Optional[Tuple[int, int]]:
+        """Return (eoa_balance, recent_real_tx_cost) on `chain`, or None.
```
OjusWiZard: **Critical (CI is failing on this):** darglint flags `DAR101: - chain` and `DAR201: - return` here. Add the directives:

```rst
:param chain: chain name as used in ``chain_to_chain_id_mapping``.
:return: ``(eoa_balance_wei, recent_real_tx_cost_wei)``, or None if either signal is unavailable.
```

While editing: "Returns None when **either** signal cannot be read" undersells it — there are ~8 distinct `return None` paths. And the hook returns a `FundRequirements` instance whose `.get_response_body()` is then called; the current wording implies the hook returns the body directly.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +156,59 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+        if not records:
+            return None
+        try:
+            recent_costs = [
+                int(r["gas_used"]) * int(r["gas_price"])
+                for r in records[-_RECENT_GAS_RECORDS_TO_CONSIDER:]
+            ]
+        except (KeyError, TypeError, ValueError):
+            return None
+        if not recent_costs:
```
OjusWiZard: **Critical:** unreachable dead code — and it makes the 100%-coverage gate unpassable. `records` is already checked non-empty above, so the slice is non-empty, and the comprehension either succeeds with ≥1 element or raises into the `except` that already returned. `recent_costs` can never be empty here. Since this line can never execute, no test can ever cover it; delete the guard.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -109,11 +116,30 @@ def async_act(self) -> Generator:  # type: ignore[override]
                             self.context.logger.info(
                                 f"Number of tx left to meet KPI: {num_of_tx_left_to_meet_kpi}"
                             )
-                            self.context.logger.info("Preparing vanity tx..")
-                            vanity_tx_hex = yield from self._prepare_vanity_tx(
+                            signal = self._real_tx_cost_vs_balance(
                                 chain=self.params.staking_chain  # type: ignore[arg-type]
                             )
-                            self.context.logger.info(f"tx hash: {vanity_tx_hex}")
+                            if signal is not None and signal[0] < signal[1]:
```
OjusWiZard: **Important:** `signal[0] < signal[1]` compares two anonymous same-typed wei values whose order lives only in a docstring — swapping the return order or the comparison silently inverts the gate, and `tox.ini` exempts this whole package from mypy (`ignore_errors = True`), so there is no tooling backstop. Minimal fix — unpack once, before the comparison (also removes the duplicate unpack on the next line):

```python
if signal is not None:
    balance, cost = signal
    if balance < cost:
        ...
```

A small `NamedTuple` (`signal.eoa_balance < signal.recent_real_tx_cost`) would be sturdier still.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +156,59 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+        cost = int(median(recent_costs))
+
+        try:
+            hook = self.context.shared_state.get(GET_FUNDS_STATUS_METHOD_NAME)
+        except (AttributeError, TypeError):
+            return None
+        if hook is None:
+            return None
+        try:
+            response = hook().get_response_body()
```
OjusWiZard, line 226: **Important (design):** this is a live, blocking RPC inside the ABCI generator. The shared-state hook is the bound method `FundsManagerBehaviour.get_funds_status`, which constructs `Web3(HTTPProvider(...))` and runs a synchronous W3Multicall **per configured chain** on every call — no caching (`funds_manager/behaviours.py:57-65,158+`). `async_act` is a cooperative generator, so a slow/hanging RPC stalls the whole agent loop until the round timer fires; there is also no explicit HTTP timeout bounding it. And it queries all chains when only `staking_chain` is needed.

Suggestion: have `funds_manager` cache its last poll result in shared state and read the cached scalar here — or at minimum bound the call with a timeout and document the latency assumption. (The `/funds-status` handler shares this behavior, but it is not on the consensus-critical path.)

OjusWiZard (commented):
Multi-perspective review (correctness, silent failures, type design, tests, comments, simplification). Findings are inline; cross-cutting items that have no diff line to attach to:

**CI is red on this PR** — `linter_checks` fails on three tools (run 27435494542):
- darglint: `DAR101`/`DAR201` on `_real_tx_cost_vs_balance`, `DAR201` on test `_make_behaviour` (inline comments below)
- black: would reformat both touched py files
- isort: both touched py files

`make format` + the docstring directives should clear it; remember to `autonomy packages lock` afterwards since fingerprints change. The `test` job was skipped because lint failed, so the 100%-coverage gate has not run yet — see the inline comment on `TestVanityTxFundingGate` for the branches it will flag.

**Design question (the only non-mechanical item):** the funds-status hook is a live, synchronous W3Multicall over *every* configured chain — see inline comment at the `hook()` call site. Worth deciding whether to read a cached value instead before merge.

Also noted: `tox.ini` sets `ignore_errors = True` for all of `liquidity_trader_abci.behaviours`, so mypy checks none of the new code — the tuple-ordering comment below has no type-check backstop.

**Verified non-issues** while reviewing: `str(chain_id)` tracker keys match what `post_tx_settlement.py` writes (incl. after JSON round-trip); `ZERO_ADDRESS` matches the production config key; funds-status `balance` is always a base-10 string or `None`, both handled.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +156,59 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+        except Exception:  # pylint: disable=broad-except
```
OjusWiZard: **Important:** every failure path in this method is fully silent, so "fail open" becomes "fail open forever, undetectably". A permanently broken hook (API drift lands here in the bare `except Exception`), corrupt gas records (line 182), or a misconfigured `chain_to_chain_id_mapping` (line 172) all look identical to a healthy fresh boot — the gate is silently dead and the exact bug this PR fixes returns with no operator signal.

Transient/expected paths (no records yet, hook not registered) can stay quiet, but the unexpected ones deserve a log:

```python
except Exception as exc:  # pylint: disable=broad-except
    self.context.logger.warning(
        f"funds-status hook raised {type(exc).__name__}({exc}); "
        "failing open (vanity tx allowed)"
    )
    return None
```

Same treatment for the malformed-gas-record `except` above.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +156,59 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+                int(r["gas_used"]) * int(r["gas_price"])
+                for r in records[-_RECENT_GAS_RECORDS_TO_CONSIDER:]
+            ]
+        except (KeyError, TypeError, ValueError):
+            return None
+        if not recent_costs:
+            return None
+        cost = int(median(recent_costs))
+
+        try:
```
OjusWiZard: Advisory: this `try/except` is unreachable — `context.shared_state` is a framework-guaranteed dict (`optimus_abci/handlers.py` subscripts it bare), and `dict.get(str_key)` raises neither `AttributeError` nor `TypeError`. The phantom safety net just obscures which call is actually risky (the `hook()` invocation below). Replace with a plain `.get()` + the existing `None` check.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -109,11 +116,30 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
                                 chain=self.params.staking_chain  # type: ignore[arg-type]
                             )
-                            self.context.logger.info(f"tx hash: {vanity_tx_hex}")
+                            if signal is not None and signal[0] < signal[1]:
+                                balance, cost = signal
+                                # Padding the multisig nonce with vanity txs
+                                # while the EOA cannot fund a real on-chain
+                                # action hides the funding alert behind a
+                                # green staking KPI.
+                                self.context.logger.warning(
```
OjusWiZard, line 141: Advisory: this is the only `%`-style logger call in the `behaviours/` directory — this file and its siblings use f-strings exclusively (including the `info` calls just below). The lazy-formatting benefit is nil for a WARNING that always emits; match the codebase style.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +156,59 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+        """
+        chain_id = self.params.chain_to_chain_id_mapping.get(chain)
+        if not chain_id:
+            return None
+        records = self.gas_cost_tracker.data.get(str(chain_id), [])
+        if not records:
+            return None
+        try:
+            recent_costs = [
+                int(r["gas_used"]) * int(r["gas_price"])
```
OjusWiZard, line 211: Advisory: if a web3 stack ever stored `gasUsed`/`effectiveGasPrice` as hex strings, `int("0x…")` raises `ValueError` and the gate silently never fires again. Today they are ints in practice (e.g. `decision_making.py` multiplies them raw), so this is hardening, not a bug — cheapest fix is normalizing with `int()` at write time in `post_tx_settlement.py` so the tracker only ever contains ints.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
```diff
@@ -23,23 +23,71 @@
...
     """Generator function that yields once and returns False."""
     yield
     return False
 
 
 def _make_behaviour():
-    """Create a CheckStakingKPIMetBehaviour without __init__."""
+    """Create a CheckStakingKPIMetBehaviour without __init__.
+
+    Mirrors production state where the base behaviour __init__ always
```
OjusWiZard, line 80: **Critical (CI is failing on this):** darglint flags `DAR201: - return` on this docstring — the multi-line description now requires a `:return:` directive:

```rst
:return: a ``CheckStakingKPIMetBehaviour`` instance with a mocked context.
```

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
```diff
@@ -415,3 +463,166 @@ def bad_hash(**kwargs):
                 assert result is None
         finally:
             mod.hash_payload_to_hex = original_hash
+
+
+class TestVanityTxFundingGate:
```
OjusWiZard, line 496: **Important (will trip the 100%-coverage gate once lint is green):** these four tests cover the right primary scenarios, but seven `return None` branches in `_real_tx_cost_vs_balance` are never executed: missing chain_id mapping (173), malformed gas record (182-183), `shared_state` except (190-191), hook raising inside `hook().get_response_body()` (196-197), `balance_str is None` (204-205), non-numeric balance (208-209) — plus the unreachable guard at 184-185 that no test can ever cover (see comment there).

Cheapest fix: direct unit tests of `_real_tx_cost_vs_balance` rather than full `async_act` runs. Also worth one boundary test: `balance == cost` → strict `<` means the vanity tx runs (fail-open) — currently the fence-post policy is implied, never asserted.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
```diff
@@ -415,3 +463,166 @@ def bad_hash(**kwargs):
...
+        synced_mock = MagicMock()
+        synced_mock.period_count = 20
+        synced_mock.period_number_at_last_cp = 0
+        synced_mock.min_num_of_safe_tx_required = 5
+        return synced_mock
+
+    def test_vanity_tx_suppressed_when_balance_below_recent_real_tx_cost(
+        self,
+    ) -> None:
+        """David's edge case: EOA below real-tx cost but vanity tx still cheap.
```
OjusWiZard: Advisory: "David's edge case" / "his Pearl logs" are person-specific references in a public repo — untraceable for any other contributor and rotted from day one. Keep the numbers, drop the person:

```
"""Suppress the vanity tx when the EOA balance is below the recent real-tx cost.

Representative values from a production incident: balance ~6.92e13 wei,
real-tx cost ~1.55e14 wei. Guard must skip _prepare_vanity_tx and emit a WARNING.
"""
```

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
```diff
@@ -415,3 +463,166 @@ def bad_hash(**kwargs):
...
+        vanity_called = {"flag": False}
+
+        def fake_prepare_vanity_tx(chain):
+            vanity_called["flag"] = True
+            yield
+            return "0xvanity"
+
+        obj._prepare_vanity_tx = fake_prepare_vanity_tx
+
+        TestCheckStakingKPIMetBehaviour()._run_async_act(
```
OjusWiZard: Advisory: instantiating the *other* test class just to borrow its helper (`TestCheckStakingKPIMetBehaviour()._run_async_act(...)`, repeated in all four tests) is fragile coupling — any future fixture/`__init__` change on that class silently breaks this one. Extract `_run_async_act` to a module-level function (or a shared base) and call it from both classes.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
```diff
@@ -415,3 +463,166 @@ def bad_hash(**kwargs):
...
+            obj, self._params_with_optimism_mapping(), self._synced_with_kpi_unmet()
+        )
+
+        assert vanity_called["flag"] is False
+        obj.context.logger.warning.assert_called()
```
OjusWiZard: **Important:** `assert_called()` only proves *a* warning was emitted — a bug that logged zeros or swapped balance/cost would still pass. Assert on the arguments:

```python
args = obj.context.logger.warning.call_args.args
assert 69_207_718_314_629 in args
assert 155_000_000_000_000 in args
```

DhairyaPatel7:
Thanks for the thorough review — addressed in `edd2f39f` (then merged `main` to clear conflicts in `06e87867`). Mapping each finding to its fix:

**CI red (lint) — fixed**
- darglint / black / isort now pass; re-ran `autonomy packages lock` so fingerprints match. Full `black-check · isort-check · flake8 · mypy · pylint · darglint` is green locally and `linter_checks` is now green on CI.

**`check_staking_kpi_met.py`**
- **Tuple ordering (122):** replaced `Tuple[int, int]` with a `_FundingSignal` NamedTuple; the gate now reads `signal.eoa_balance < signal.recent_real_tx_cost`, so swapping order can't silently invert it.
- **Docstring (162):** added `:param:` / `:return:`; reworded "either signal" to reflect the several distinct fail-open paths, and clarified the hook returns a `FundRequirements` whose `get_response_body()` is read.
- **Dead code (184):** deleted the unreachable `if not recent_costs` guard.
- **Phantom try/except (188):** `shared_state` is a framework-guaranteed dict — now a plain `.get()`.
- **Silent failures (196):** the *unexpected* fail-open paths (hook raising, malformed gas record) now emit a WARNING; the routine empty paths (no mapping, no records yet, hook not registered, balance absent) stay quiet.
- **Logger style (128):** suppression WARNING is now an f-string.

**`post_tx_settlement.py`**
- **Hex hardening (179):** `gas_used` / `effective_gas_price` are normalized with `int()` at the single write chokepoint, so the tracker only ever holds ints.

**Tests**
- **darglint (81):** added `:return:`.
- **Coverage (468):** new `TestRealTxCostVsBalance` directly unit-tests every `return None` branch plus the `balance == cost` fence-post → module is back at **100% branch coverage**.
- **Fragile coupling (545):** `_run_async_act` is now a module-level helper used by both classes.
- **Weak assert (550):** the suppression test now asserts the WARNING carries the real balance/cost numbers (`69207718314629`, `155000000000000`).
- **Person-specific docstring (517):** dropped the name/Pearl-logs reference, kept the representative numbers.

**Design question (195) — live, blocking funds-status RPC.** I kept the live read and documented the latency assumption rather than adding a caching layer, for two reasons:
1. The call *is* bounded — `funds_manager._get_web3` sets `request_kwargs={"timeout": rpc_timeout_seconds}` plus a retry config, so each RPC has an explicit HTTP timeout (worst case `chains × retries × timeout`).
2. The gate only runs on the rare KPI-behind path (KPI unmet **and** threshold period exceeded **and** tx-left > 0), not every round.

If you'd still prefer the cached-scalar approach (funds_manager writes its last poll into shared state and the gate reads that, also letting us query only `staking_chain`), I'm happy to do it as a follow-up — just say the word and I'll wire it up.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -130,6 +168,78 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+        error never silently kills the staking KPI. Routine empty paths (no
+        chain mapping, no records yet, hook not registered, balance absent) stay
+        quiet; the unexpected ones (malformed gas record, hook raising) emit a
+        WARNING so a permanently dead gate is not mistaken for a healthy boot.
+
+        Latency note: the hook runs synchronous Multicall RPCs (one per
+        configured chain). Each RPC is bounded by funds_manager's
+        ``rpc_timeout_seconds`` and retry config, and this gate only runs on the
+        rare KPI-behind path, so the cooperative ``async_act`` loop is not
+        stalled unboundedly.
```
OjusWiZard: **Important — this latency note is not true, and it's the justification for the still-open blocking-RPC thread.** `rpc_timeout_seconds` exists nowhere in the repo except this line (grepped). `funds_manager/behaviours.py:59` is a bare `Web3(Web3.HTTPProvider(rpc_url))` — no `request_kwargs={'timeout': ...}`, no retry wrapper — and web3's HTTPProvider passes no default timeout to `requests`, so the call can block `async_act` indefinitely. The note papers over the concern with a config key that doesn't exist rather than mitigating it.

Either wire a real timeout/retry into `funds_manager._perform_w3_multicall`, or reword to state plainly that the RPC is currently unbounded and the risk is accepted because the gate only runs on the rare KPI-behind path. (Also: it still queries *all* configured chains when only `staking_chain` is needed.)

packages/valory/skills/liquidity_trader_abci/behaviours/post_tx_settlement.py
```diff
@@ -87,6 +87,11 @@ def fetch_and_log_gas_details(self) -> Generator[Any, Any, Any]:
         effective_gas_price = response.get("effectiveGasPrice")
         gas_used = response.get("gasUsed")
         if gas_used and effective_gas_price:
+            # Normalize to int at the single write chokepoint so the tracker
+            # only ever holds ints; downstream readers (e.g. the vanity-tx
+            # funding gate) then never hit a hex-string -> ValueError path.
```
OjusWiZard: **Important — the rationale here is backwards.** `int("0xff")` *raises* `ValueError` (only `int(x, 16)` parses hex). So if receipts ever delivered hex strings, lines 93-94 would crash at **write** time — this normalization does not protect readers from a hex-string path. In practice the values are already ints, so the `int()` calls really guard against `float`/decimal-string, not hex. The code is harmless; just fix the comment, e.g. "values may arrive as float or decimal-string from some middleware paths, so normalize to int here so the tracker only ever holds ints."

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
```diff
@@ -415,3 +491,284 @@ def bad_hash(**kwargs):
...
+    def test_balance_equal_cost_is_not_suppressed(self) -> None:
+        """Fence-post: balance == cost means strict ``<`` is False -> vanity runs."""
+        obj = self._obj_with(
+            records=_gas_records(200),
+            shared_state=self._healthy_shared_state(200),
+        )
+        signal = _call_signal(obj, self._params())
+        assert signal is not None
+        assert signal.eoa_balance == signal.recent_real_tx_cost
+        assert not (signal.eoa_balance < signal.recent_real_tx_cost)
```
OjusWiZard: **Important — this fence-post test is tautological.** Line 774 asserts `not (signal.eoa_balance < signal.recent_real_tx_cost)` on two equal values — that's `not (200 < 200)`, always true regardless of production code. It tests Python's `<` operator, not the gate. It never runs `async_act` / checks that `_prepare_vanity_tx` fires, so the exact `<` vs `<=` boundary at `check_staking_kpi_met.py:135` is pinned by **no** test (the two integration tests only cover strictly-below and strictly-above). A `<`→`<=` regression would ship green.

Fix: add a `TestVanityTxFundingGate` case with `balance == cost` asserting `vanity_called["flag"] is True` (mirrors `test_vanity_tx_runs_when_balance_above_recent_real_tx_cost`).

OjusWiZard (commented):
Re-review of the fix commit (edd2f39f). CI is fully green and all 12 prior code findings are addressed cleanly — thanks. Three items remain, two of which were *introduced* by the fix commit itself. The blocking-RPC item (#1) bears directly on the one review thread I left open; I've also replied there.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
OjusWiZard: Following up on this in the re-review of edd2f39f: the fix commit answered this thread with a docstring latency note (check_staking_kpi_met.py:188-192) claiming the RPC is "bounded by funds_manager's `rpc_timeout_seconds` and retry config." That config does not exist — `rpc_timeout_seconds` appears nowhere in the repo except that line, and `funds_manager/behaviours.py:59` builds a bare `Web3(Web3.HTTPProvider(rpc_url))` with no timeout kwarg and no retry. So the blocking call is still unbounded; the note documents a mitigation that isn't there. Leaving this thread open — see the new inline comment at the docstring for the suggested resolution (wire a real timeout, or document the risk honestly).

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
DhairyaPatel7: You're right — `funds_manager` is gitignored in this repo (`.gitignore:27`) and referenced only by package hash, and the shipped version builds a bare `Web3(HTTPProvider(rpc_url))` with no timeout, so `rpc_timeout_seconds` was a phantom (it existed only in my local untracked copy). Fixed in 0bf618b4: the note now states plainly that the RPC is currently unbounded and queries all configured chains, and that the risk is accepted because the gate only runs on the rare KPI-behind path — with the timeout/cache mitigation called out as a `funds_manager` follow-up (it can't be wired here since that package isn't sourced in this repo).

packages/valory/skills/liquidity_trader_abci/behaviours/post_tx_settlement.py
DhairyaPatel7: Correct — `int("0xff")` raises, so this never protected against a hex-string path; in practice it guards float / decimal-string. Reworded the comment accordingly in 0bf618b4.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_check_staking_kpi_met.py
DhairyaPatel7: Agreed — it was testing Python's `<`, not the gate. 0bf618b4 refocuses the unit assertion onto the method's returned values and adds `TestVanityTxFundingGate.test_vanity_tx_runs_when_balance_equals_cost`, which drives `async_act` with balance == cost and asserts the vanity tx fires (and no WARNING is emitted). A `<`→`<=` regression now fails this test. Module stays at 100% coverage.

LOCKhart07 (approved):
Re-reviewed against head `0bf618b4`. All prior review threads are genuinely resolved (NamedTuple gate, fail-open WARNINGs, dead-guard removal, darglint/coverage), and I independently verified the load-bearing assumptions: gas-record keys (`gas_used`/`gas_price`) match the reader, vanity txs are excluded from the tracker at `post_tx_settlement.py:44`, native balance keys under `ZERO_ADDRESS` in the default config, and `agent_address` casing matches between the gate and funds_manager. Tests cover every `return None` branch with falsifiable assertions and pin the `<` fence-post end-to-end. CI is green. Approving.

One non-blocking accuracy note + two minor latent notes below.

**1. The latency docstring (`check_staking_kpi_met.py:188-193`) misdescribes the pinned funds_manager.** It states funds_manager "builds its `Web3` provider with no explicit HTTP timeout, so a slow/hanging endpoint can block the cooperative `async_act` loop until the round timer fires." The version this PR actually pins (`bafybeibne4…`, from the open-autonomy 0.21.23 bump, confirmed ancestor of the head) builds the provider in `_get_web3` with `request_kwargs={"timeout": rpc_timeout_seconds}` (default 10s) **and** `ExceptionRetryConfiguration(retries=rpc_max_retries)` (default 2), and caches the client. So the RPC is **bounded** (~10s × 3 attempts, serial per chain), not unbounded — the opposite of what the note claims. This is the second time this note has been off (first it over-claimed the mitigation; now it over-claims the risk). The gate behaves correctly regardless; only the documented justification is wrong. Suggest rewording to reflect the real bound (e.g. "bounded by funds_manager's `rpc_timeout_seconds` (default 10s) and `rpc_max_retries` (default 2), serial across all configured chains").

**2. (minor, fail-open) Hardcoded `ZERO_ADDRESS` native key.** `_real_tx_cost_vs_balance` reads native balance via `.get(ZERO_ADDRESS)`, but funds_manager also treats `0xeeee…eeee` as native. An operator configuring the native token with that sentinel in `FUND_REQUIREMENTS` would silently fail the gate open. The default optimus config uses the zero address, so this is safe today — just a latent footgun worth a comment.

**3. (minor, fail-open) Chain must be in `FUND_REQUIREMENTS`.** `response.get(chain)` requires `staking_chain` to be present in the operator's `FUND_REQUIREMENTS`, else the gate fails open. Safe, but undocumented.

OjusWiZard (approved):
Approving. Three review rounds; all code findings addressed and verified against the latest commit (0bf618b4), CI fully green (lint, 100% coverage, full test matrix). The one remaining item — the synchronous, currently-unbounded funds-status RPC inside `async_act` — is accepted as a documented risk (it only runs on the rare KPI-behind path) with a timeout/caching fix deferred to a `funds_manager` follow-up.

PR #341, opened 2026-06-18
feat(whitelist): restrict Optimism and Base whitelists to approved stablecoins

DhairyaPatel7:
## What

Restricts the **investable** asset whitelist for the **optimus** (Optimism) and **basius** (Base) agents to an approved stablecoin set, while keeping already-held but de-whitelisted assets detectable and convertible.

### Investable set — `WHITELISTED_ASSETS` (gates *new* positions; consumed by `evaluate_strategy`)

**Optimism (5):**
| Symbol | Address |
| --- | --- |
| USDC (native) | `0x0b2c639c533813f4aa9d7837caf62653d097ff85` |
| sDAI | `0x2218a117083f5b482b0bb821d27056ba9c04b1d3` |
| USDT0 | `0x01bff41798a0bcf287b996046ca68b395dbc1071` |
| USDC.e | `0x7f5c764cbc14f9669b88837ca1490cca17c31607` |
| USDT (bridged) | `0x94b008aa00579c1307b0ef2c499ad98a8ce58e58` |

**Base (6):**
| Symbol | Address |
| --- | --- |
| USDC (native) | `0x833589fcd6edb6e08f4c7c32d4f71b54bda02913` |
| BOLD | `0x03569cc076654f82679c4ba2124d64774781b01d` |
| msUSD | `0x526728dbc96689597f85ae4cd716d4f7fccbae9d` |
| frxUSD | `0xe5020a6d073a794b6e7f05678707de47986fb0b6` |
| eUSD | `0xcfa3ef56d303ae4faaba0592388f19d7c3399fb4` |
| axlUSDC | `0xeb466342c4d449bc9f53a865d5cb90586f405215` |

Mode (legacy) is left unchanged.

## Investable vs. held/sweep split (review fix)

`WHITELISTED_ASSETS` is a single dict that drove **both** investability and on-chain detection/sweep. Removing oUSDT from it would also drop it from balance detection and the withdraw sweep — stranding oUSDT that existing optimus agents still hold (no loss; invisible during SafeApi gaps and not auto-converted). To avoid that, the concern is now split in `liquidity_trader_abci/behaviours/base.py`:

- **`WHITELISTED_ASSETS`** — the investable set above. Gates *new* positions only (`evaluate_strategy`).
- **`LEGACY_HELD_ASSETS`** — assets removed from investment but possibly still held (currently **oUSDT** on Optimism).
- **`HELD_ASSETS`** — superset (investable + legacy held), consumed by the three detection/sweep sites: the last-known-balance tracker, the on-chain balance backstop (`_supplement_with_onchain_whitelisted_balances`), and the withdraw sweep-to-USDC fallback (`_prepare_swap_to_usdc_actions_standard`).

Net: no new oUSDT investment, but held oUSDT stays **visible, convertible, and priceable**. `COIN_ID_MAPPING` is intentionally left intact so held-but-de-whitelisted tokens can still be priced via CoinGecko. Drop the `LEGACY_HELD_ASSETS` entry in a follow-up once on-chain oUSDT reaches zero.

## Runtime mechanism (corrected)

The real consumers (investability, detection, sweep) read the **module constants** (`WHITELISTED_ASSETS` / `HELD_ASSETS`) directly, so the trim takes effect immediately for all agents. The JSON-backed `self.whitelisted_assets` is **not** reconciled from the constant — `fetch_strategies.py:315-319` only seeds it from `WHITELISTED_ASSETS` when the persisted file is empty; otherwise existing agents keep reading their (now stale) `whitelisted_assets.json`. That stale JSON is only the price-drop tracker's state and does not gate the behaviour this PR changes. (Corrects an earlier, inaccurate claim that the JSON is reconciled from the constant.)

## Tests
- Backstop and sweep tests now reference `HELD_ASSETS`, so they cover the legacy-held **oUSDT** path the trim actually changes.
- New `TestWhitelistInvariants` pins the exact investable membership per chain, asserts oUSDT is **not** investable but **remains** in `HELD_ASSETS`/`LEGACY_HELD_ASSETS`, and checks the `HELD_ASSETS ⊇ WHITELISTED_ASSETS` invariant.
- `test_base.py`, `test_withdraw_funds.py`, `test_evaluate_strategy.py`, `test_fetch_strategies.py` — **1455 passed**.
- Re-locked package hashes (`autonomy packages lock`).

## Base de-whitelist asymmetry (intentional)

This PR also removes 6 tokens from the Base investable set (USDT `0xfde4…`, USDe, DOLA, MUSD, USR, USDz) without adding them to `LEGACY_HELD_ASSETS`. That is deliberate: **basius (Base) is not launched yet**, so no agent holds any of them — there are no Base balances to strand, so the oUSDT-style held/sweep handling is unnecessary. If basius launches and any held token is later de-whitelisted, it should be added to `LEGACY_HELD_ASSETS["base"]` then. The Base investable set is pinned by `TestWhitelistInvariants` to prevent drift.

### DIFF
--- a/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
+++ b/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
@@ -280,7 +280,11 @@ POOL_FILENAME = "current_pool.json"
 READ_MODE = "r"
 WRITE_MODE = "w"
 
-# Whitelist of allowed token addresses for each chain with their symbols
+# Investable whitelist: token addresses (with symbols) the agent is allowed to
+# ENTER new positions with, per chain. This is the gate consumed by the
+# strategy / investability path (evaluate_strategy). To stop investing in an
+# asset, remove it here -- but if existing agents may still HOLD it, also add it
+# to LEGACY_HELD_ASSETS below so it stays detectable, convertible and priceable.
 WHITELISTED_ASSETS = {
     "mode": {
         # MODE tokens - stablecoins
@@ -316,6 +320,30 @@ WHITELISTED_ASSETS = {
         "0xcfa3ef56d303ae4faaba0592388f19d7c3399fb4": "eUSD",
         "0xeb466342c4d449bc9f53a865d5cb90586f405215": "axlUSDC",
     },
+}
+
+# Assets that are no longer investable (intentionally absent from
+# WHITELISTED_ASSETS) but may still sit in an existing agent's Safe. They are
+# excluded from new investment yet must remain (a) detectable by the on-chain
+# balance backstop and last-known-balance tracker, (b) convertible by the
+# withdraw sweep-to-USDC fallback, and (c) priceable via COIN_ID_MAPPING. Drop
+# an entry here once on-chain balances for it have reached zero.
+LEGACY_HELD_ASSETS: Dict[str, Dict[str, str]] = {
+    # oUSDT was removed from the Optimism investable whitelist, but existing
+    # optimus agents may still hold it.
+    "optimism": {
+        "0x1217bfe6c773eec6cc4a38b5dc45b92292b6e189": "oUSDT",
+    },
+}
+
+# Detection / sweep set: every investable asset PLUS any legacy held asset for
+# the chain. Consumed by the on-chain balance backstop
+# (_supplement_with_onchain_whitelisted_balances), the last-known-balance
+# tracker, and the withdraw sweep-to-USDC fallback. Always a superset of
+# WHITELISTED_ASSETS.
+HELD_ASSETS: Dict[str, Dict[str, str]] = {
+    chain: {**assets, **LEGACY_HELD_ASSETS.get(chain, {})}
+    for chain, assets in WHITELISTED_ASSETS.items()
 }
 
 COIN_ID_MAPPING = {
@@ -587,7 +615,7 @@ class LiquidityTraderBaseBehaviour(
         if native_balance is not None:
             current[ZERO_ADDRESS.lower()] = int(native_balance)
 
-        for token_addr, _symbol in WHITELISTED_ASSETS.get(chain, {}).items():
+        for token_addr, _symbol in HELD_ASSETS.get(chain, {}).items():
             token_balance = yield from self._get_token_balance(
                 chain, safe_address, token_addr
             )
@@ -859,11 +887,12 @@ class LiquidityTraderBaseBehaviour(
                         }
                     )
 
-        # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
-        # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
-        # returns an incomplete trusted-token list. Covers every entry in
-        # WHITELISTED_ASSETS[chain]. Reward tokens (OLAS, VELO) are disjoint
-        # from the whitelist and handled by _fetch_reward_balances below.
+        # On-chain backstop for held ERC20s and native ETH. Catches the case
+        # where SafeApi is unavailable (e.g. monthly quota exceeded) or returns
+        # an incomplete trusted-token list. Covers every entry in
+        # HELD_ASSETS[chain] (the investable whitelist plus legacy held assets),
+        # so already-held but de-whitelisted tokens stay visible. Reward tokens
+        # (OLAS, VELO) are disjoint and handled by _fetch_reward_balances below.
         yield from self._supplement_with_onchain_whitelisted_balances(
             chain, safe_address, balances
         )
@@ -3402,9 +3431,9 @@ class LiquidityTraderBaseBehaviour(
         safe_address: str,
         balances: List[Dict[str, Any]],
     ) -> Generator[None, None, None]:
-        """Add on-chain balances for whitelisted assets not present in ``balances``.
+        """Add on-chain balances for held assets not present in ``balances``.
 
-        Mutates ``balances`` in place. For each entry in ``WHITELISTED_ASSETS[chain]``
+        Mutates ``balances`` in place. For each entry in ``HELD_ASSETS[chain]``
         plus native ETH, appends an entry sourced from on-chain reads when:
         - the address is not already in ``balances`` (case-insensitive match), and
         - the on-chain balance is strictly positive.
@@ -3448,7 +3477,7 @@ class LiquidityTraderBaseBehaviour(
                     f"Error reading native balance on {chain}: {str(e)}"
                 )
 
-        for whitelisted_addr, symbol in WHITELISTED_ASSETS.get(chain, {}).items():
+        for whitelisted_addr, symbol in HELD_ASSETS.get(chain, {}).items():
             key = whitelisted_addr.lower()
             if key in already_have:
                 continue
--- a/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
+++ b/packages/valory/skills/liquidity_trader_abci/behaviours/base.py
@@ -302,32 +302,19 @@ WHITELISTED_ASSETS = {
     "optimism": {
         # Optimism tokens - stablecoins
         "0x0b2c639c533813f4aa9d7837caf62653d097ff85": "USDC",
-        "0x01bff41798a0bcf287b996046ca68b395dbc1071": "USDT0",
-        "0x94b008aa00579c1307b0ef2c499ad98a8ce58e58": "USDT",
-        "0x7f5c764cbc14f9669b88837ca1490cca17c31607": "USDC.e",
         "0x2218a117083f5b482b0bb821d27056ba9c04b1d3": "sDAI",
-        "0x1217bfe6c773eec6cc4a38b5dc45b92292b6e189": "oUSDT",
+        "0x01bff41798a0bcf287b996046ca68b395dbc1071": "USDT0",
+        "0x7f5c764cbc14f9669b88837ca1490cca17c31607": "USDC.e",
+        "0x94b008aa00579c1307b0ef2c499ad98a8ce58e58": "USDT",
     },
     "base": {
         # Base tokens - stablecoins
         "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913": "USDC",
-        "0x1217bfe6c773eec6cc4a38b5dc45b92292b6e189": "oUSDT",
-        "0xeb466342c4d449bc9f53a865d5cb90586f405215": "axlUSDC",
+        "0x03569cc076654f82679c4ba2124d64774781b01d": "BOLD",
         "0x526728dbc96689597f85ae4cd716d4f7fccbae9d": "msUSD",
         "0xe5020a6d073a794b6e7f05678707de47986fb0b6": "frxUSD",
-        # Tier-1 stablecoin additions from the Base/Aerodrome LpSugar whitelist
-        # sweep: each is $1-pegged, has a CoinGecko coin id for pricing, and pairs
-        # with a whitelisted anchor (USDC; eUSD also pairs frxUSD). A pool is only
-        # investable when every token in it is whitelisted, so each unlocks real
-        # Aerodrome depth. Ordered by safety x liquidity.
-        "0xfde4c96c8593536e31f229ea8f37b2ada2699bb2": "USDT",
-        "0x5d3a1ff2b6bab83b63cd9ad0787074081a52ef34": "USDe",
         "0xcfa3ef56d303ae4faaba0592388f19d7c3399fb4": "eUSD",
-        "0x4621b7a9c75199271f773ebd9a499dbd165c3191": "DOLA",
-        "0xdd468a1ddc392dcdbef6db6e34e89aa338f9f186": "MUSD",
-        "0x35e5db674d8e93a03d814fa0ada70731efe8a4b9": "USR",
-        "0x04d5ddf5f3a8939889f11e97f8c4bb48317f1938": "USDz",
-        "0x03569cc076654f82679c4ba2124d64774781b01d": "BOLD",
+        "0xeb466342c4d449bc9f53a865d5cb90586f405215": "axlUSDC",
     },
 }
 
@@ -875,9 +862,8 @@ class LiquidityTraderBaseBehaviour(
         # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
         # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
         # returns an incomplete trusted-token list. Covers every entry in
-        # WHITELISTED_ASSETS[chain] including oUSDT, which used to be fetched
-        # separately. Reward tokens (OLAS, VELO) are disjoint from the
-        # whitelist and handled by _fetch_reward_balances below.
+        # WHITELISTED_ASSETS[chain]. Reward tokens (OLAS, VELO) are disjoint
+        # from the whitelist and handled by _fetch_reward_balances below.
         yield from self._supplement_with_onchain_whitelisted_balances(
             chain, safe_address, balances
         )
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -22,10 +22,10 @@
         "contract/valory/velodrome_voter/0.1.0": "bafybeieawpdivhzrbjk6ivsrncxah5y5gbb32qxmrcb2dxmw55ev7gowda",
         "contract/valory/velodrome_cl_gauge/0.1.0": "bafybeicqjs5styv3qubfd7askc6nnd2hrwqlawd6b7rvzyvp2ycbenensm",
         "contract/valory/balancer_queries/0.1.0": "bafybeieopocjnprwfazmacrgr4n6up2p2warpq7seiwhocb5f2uy4mrmrq",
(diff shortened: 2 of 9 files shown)


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -302,32 +302,19 @@ class Chain(Enum):
     "optimism": {
         # Optimism tokens - stablecoins
         "0x0b2c639c533813f4aa9d7837caf62653d097ff85": "USDC",
+        "0x2218a117083f5b482b0bb821d27056ba9c04b1d3": "sDAI",
         "0x01bff41798a0bcf287b996046ca68b395dbc1071": "USDT0",
-        "0x94b008aa00579c1307b0ef2c499ad98a8ce58e58": "USDT",
         "0x7f5c764cbc14f9669b88837ca1490cca17c31607": "USDC.e",
-        "0x2218a117083f5b482b0bb821d27056ba9c04b1d3": "sDAI",
-        "0x1217bfe6c773eec6cc4a38b5dc45b92292b6e189": "oUSDT",
+        "0x94b008aa00579c1307b0ef2c499ad98a8ce58e58": "USDT",
```
OjusWiZard, line 312: Advisory, non-blocking: `COIN_ID_MAPPING["optimism"]` still carries `"ousdt": "openusdt"` (~line 356) for the now de-whitelisted oUSDT. It's unreachable dead config (the symbol no longer surfaces on this chain), but worth pruning while you're curating Optimism.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -302,32 +302,19 @@ class Chain(Enum):
...
     },
     "base": {
         # Base tokens - stablecoins
         "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913": "USDC",
-        "0x1217bfe6c773eec6cc4a38b5dc45b92292b6e189": "oUSDT",
-        "0xeb466342c4d449bc9f53a865d5cb90586f405215": "axlUSDC",
+        "0x03569cc076654f82679c4ba2124d64774781b01d": "BOLD",
```
OjusWiZard, line 317: Advisory, non-blocking: `COIN_ID_MAPPING["base"]` still has entries for tokens removed here -- `ousdt`/`usde`/`dola`/`musd`/`usr`/`usdz` (~lines 365-371). All dead config now. Safe to leave, but a good moment to prune for consistency with the curated Base whitelist.

OjusWiZard (dismissed):
Approving. The core change is low-risk and correct: no new addresses are introduced (Optimism nets to removing oUSDT only; sDAI/USDT are reordered), no dangling references to the removed addresses exist, the backstop comment is accurate, and the oUSDT->sDAI test pivot is valid (sDAI is still whitelisted for Optimism).

One non-blocking follow-up inlined below: the COIN_ID_MAPPING tables still carry entries for tokens de-whitelisted here.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -875,9 +862,8 @@ def _get_safe_balances_from_safe_api(
         # On-chain backstop for whitelisted ERC20s and native ETH. Catches the
         # case where SafeApi is unavailable (e.g. monthly quota exceeded) or
         # returns an incomplete trusted-token list. Covers every entry in
-        # WHITELISTED_ASSETS[chain] including oUSDT, which used to be fetched
-        # separately. Reward tokens (OLAS, VELO) are disjoint from the
-        # whitelist and handled by _fetch_reward_balances below.
+        # WHITELISTED_ASSETS[chain]. Reward tokens (OLAS, VELO) are disjoint
```
bennyjo: **Blocking for the Optimism (optimus) side — basius/Base is fine, no holdings exist there.**

This comment now asserts the backstop "Covers every entry in `WHITELISTED_ASSETS[chain]`" — but optimus agents currently hold oUSDT (`0x1217bfe6…`), and this PR removes oUSDT from `WHITELISTED_ASSETS["optimism"]`. Because the single dict drives both *investability* and *detection/sweep*, removing oUSDT also drops it from:

- `base.py:3451` `_supplement_with_onchain_whitelisted_balances` — the on-chain balance backstop this comment describes (the one for when SafeApi is unavailable/incomplete), and
- `withdraw_funds.py:422` `_prepare_swap_to_usdc_actions_standard` — the withdraw swap-to-USDC fallback for idle tokens not already in `portfolio_breakdown`.

Net effect: held oUSDT can drop out of the agent's balance view during a SafeApi gap and isn't auto-converted on withdrawal unless it happens to be in `portfolio_breakdown`. Funds stay Safe-bound (no theft risk), but they can be stranded from the automated flows.

Would you consider splitting the Optimism entry into an *investable* set (the trimmed 5, consumed by `evaluate_strategy.py:1375,2499`) and a broader *held/sweep* set (the 5 + oUSDT, consumed by the two backstops above plus `base.py:590`)? That stops new oUSDT investment while keeping existing oUSDT detectable and convertible; once on-chain oUSDT reaches zero, drop the sweep entry in a follow-up.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -302,32 +302,19 @@ class Chain(Enum):
     "optimism": {
         # Optimism tokens - stablecoins
         "0x0b2c639c533813f4aa9d7837caf62653d097ff85": "USDC",
+        "0x2218a117083f5b482b0bb821d27056ba9c04b1d3": "sDAI",
```
bennyjo, line 309: Two smaller follow-ups on this trim:

1. **Test coverage** — the suite iterates `WHITELISTED_ASSETS["optimism"]` dynamically (e.g. `test_base.py:1729/1749/1751`), so it adapts to any contents and wouldn't catch a mistyped address or an accidental re-add. Consider a membership assertion pinning each chain's exact set (Optimism == these 5, Base == these 6), plus a negative test that a removed token (oUSDT) is *not* picked up by the backstop — the repoints move the existing assertions onto sDAI (still whitelisted), so the path this PR actually changes stays untested.
2. **PR description accuracy** (not code-blocking) — the description says "at runtime `self.whitelisted_assets = WHITELISTED_ASSETS` and the JSON is reconciled from it." `fetch_strategies.py:315–319` only assigns the constant when the persisted JSON is empty; existing agents read the stale JSON, which is never reconciled from the constant. The trim still works because the real consumers read the module constant directly — but the JSON reconciliation the description describes doesn't run, and removed tokens linger in each agent's `whitelisted_assets.json`. Worth correcting so a future maintainer doesn't rely on it.

bennyjo (commented):
Whitelist trim looks correct for Base; flagging one blocking item for the Optimism oUSDT removal (held by optimus agents) plus two follow-ups.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
bennyjo: Non-blocking

LOCKhart07 (dismissed):
Reviewed — looks good. Diff matches the description exactly (Optimism nets to 5 tokens, removing only oUSDT; Base nets to 6, removing oUSDT/USDT/USDe/DOLA/MUSD/USR/USDz), hashes propagate consistently, and the test repoints (oUSDT → sDAI) keep the same backstop/dedup/dust logic under test with a still-whitelisted token.

Verified the restriction is actually effective: every investment-decision hot path keys off the `WHITELISTED_ASSETS` constant directly (`evaluate_strategy.py:1375/2499`, `withdraw_funds.py:422`, `base.py:610/3586`), not the persisted `data/whitelisted_assets.json`, so the trim takes effect immediately regardless of any stale on-disk file.

No blocking issues. Two non-blocking notes:

1. **Stranded-funds risk for existing holders of removed tokens.** Both the on-chain backstop (`base.py:610/3586`) and the withdrawal swap-to-USDC sweep (`withdraw_funds.py:417`) iterate `WHITELISTED_ASSETS`. After this trim, an *idle* balance of a now-removed token (oUSDT on OP; USDT/USDe/DOLA/MUSD/USR/USDz on Base) becomes invisible — it won't appear in the portfolio and won't be swept to USDC on withdrawal. Existing pool *positions* can still be exited (exit is driven by `current_positions`), but the resulting non-whitelisted tokens won't be swept. Worth confirming no live agent currently holds these, or sweeping them first.

2. *(Minor/optional)* `COIN_ID_MAPPING` still lists the removed symbols (`ousdt`, `usde`, `dola`, `musd`, `usr`, `usdz`), and `velodrome_pools_search.py` still has oUSDT in its decimals/coingecko maps. Harmless — these are pricing dictionaries (a superset is fine) and investment is gated by the whitelist — but they're now orphaned for these chains if you want the "whitelist is source of truth" cleanliness to extend there.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
bennyjo: Open question from my point of view if this is relevant.

DhairyaPatel7: Fixed in ff30aba by splitting the single dict into two roles:

- `WHITELISTED_ASSETS` is now **investable-only** (gates `evaluate_strategy`).
- `LEGACY_HELD_ASSETS` holds de-whitelisted-but-possibly-still-held assets (currently `oUSDT` on Optimism).
- `HELD_ASSETS` = investable ∪ legacy, and drives the three detection/sweep sites: the last-known-balance tracker (`base.py:618`), `_supplement_with_onchain_whitelisted_balances` (`base.py:3480`), and `_prepare_swap_to_usdc_actions_standard` (`withdraw_funds.py:424`).

Net: no new oUSDT investment, but held oUSDT stays visible during SafeApi gaps and still auto-converts on withdrawal. `COIN_ID_MAPPING` kept intact so it's also priceable. `LEGACY_HELD_ASSETS` is the single place to drop oUSDT in a follow-up once on-chain balance reaches zero. Base/basius unchanged in behaviour (no holdings, `HELD_ASSETS["base"] == WHITELISTED_ASSETS["base"]`).

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DhairyaPatel7: Both addressed in ff30aba:

1. **Tests** — added `TestWhitelistInvariants` pinning the exact investable membership (Optimism == the 5, Base == the 6), a negative assertion that oUSDT is *not* investable, that it *remains* in `HELD_ASSETS`/`LEGACY_HELD_ASSETS`, and the `HELD_ASSETS ⊇ WHITELISTED_ASSETS` superset invariant. The backstop/sweep tests now reference `HELD_ASSETS` and run on **oUSDT**, so the path the trim actually changed is covered (not parked on sDAI).
2. **PR description** — corrected. The JSON is *not* reconciled from the constant: `fetch_strategies.py:315-319` only seeds `self.whitelisted_assets` from `WHITELISTED_ASSETS` when the persisted file is empty; existing agents keep reading the stale JSON, which is only the price-drop tracker's state. The trim works because the real consumers read the module constants directly. Description now says this.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DhairyaPatel7: Deliberately keeping this one. In ff30aba, oUSDT is retained in `HELD_ASSETS`/`LEGACY_HELD_ASSETS` so existing optimus agents' held oUSDT stays detectable and convertible — which means it also still needs its coin id to be **priced** via CoinGecko (`get_coin_id_from_symbol` → `openusdt`, `base.py:2478`). Pruning `COIN_ID_MAPPING["optimism"]["ousdt"]` would make held oUSDT unpriceable (coin id `None` → no value/APR). So it's load-bearing on Optimism now, not dead. I'll drop the `LEGACY_HELD_ASSETS` entry and this mapping entry together in a follow-up once on-chain oUSDT reaches zero.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DhairyaPatel7: Agreed these are genuinely dead on Base (no holdings there). Leaving them for now to avoid extra churn and to keep the mapping symmetric with the Optimism held-asset handling — they're simply never reached, so harmless. Happy to prune the Base-only entries (`usde`/`dola`/`musd`/`usr`/`usdz`) in this PR if you'd prefer; just say the word and I'll add it.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -302,35 +306,46 @@ class Chain(Enum):
...
+    },
+}
+
+# Assets that are no longer investable (intentionally absent from
+# WHITELISTED_ASSETS) but may still sit in an existing agent's Safe. They are
+# excluded from new investment yet must remain (a) detectable by the on-chain
+# balance backstop and last-known-balance tracker, (b) convertible by the
+# withdraw sweep-to-USDC fallback, and (c) priceable via COIN_ID_MAPPING. Drop
+# an entry here once on-chain balances for it have reached zero.
+LEGACY_HELD_ASSETS: Dict[str, Dict[str, str]] = {
```
bennyjo, line 331: This split cleanly fixes the oUSDT stranding on Optimism — but the same PR de-whitelists 6 Base tokens (USDT `0xfde4…`, USDe, DOLA, MUSD, USR, USDz) and none of them get the `LEGACY_HELD_ASSETS` treatment, so they're absent from `HELD_ASSETS["base"]`.

Those 6 shipped in v0.11.0 (`e1d413ec`, 2026-06-05) and were genuinely investable on Base for ~13 days, so a basius agent could be holding them. If so, on pool exit the token lands idle in the Safe and — during a SafeApi gap / incomplete trusted-token list — the on-chain backstop won't surface it (not in `HELD_ASSETS`) and the withdraw sweep won't convert it. That's the exact failure mode this PR fixes for oUSDT, left open on Base. (Pricing is fine — they're still in `COIN_ID_MAPPING["base"]` — it's detection during the SafeApi gap that's missing.)

Is it confirmed no basius agent holds these (e.g. basius not live on v0.11.0, or never entered those pools)? If yes, worth a line in the PR body so the asymmetry doesn't read as an oversight. If not, they belong in `LEGACY_HELD_ASSETS["base"]` for symmetry with oUSDT. Either way, a `TestWhitelistInvariants` assertion pinning the Base set's intended disposition would prevent future drift — the Base side is currently only covered by dynamic iteration, which won't catch an accidental re-add or a missing legacy entry.

bennyjo (commented):
Re-review: Optimism oUSDT split looks correct — one question on the Base de-whitelisted set.

OjusWiZard (dismissed):
Re-reviewed after ff30aba. The HELD_ASSETS / LEGACY_HELD_ASSETS split cleanly resolves the backstop/sweep stranding concern: all three detection/sweep sites (last-known-balance tracker, on-chain balance backstop, withdraw sweep-to-USDC) now read HELD_ASSETS, while every investability site correctly stays on WHITELISTED_ASSETS. HELD_ASSETS is derived as the union so the superset invariant cannot drift. Comments are accurate, and TestWhitelistInvariants pins each chain's exact investable set, asserts oUSDT is non-investable but still held, and verifies the superset invariant — with the backstop/sweep tests now exercising the legacy-held oUSDT path the trim actually changed. No new addresses introduced; no dangling references. LGTM.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DhairyaPatel7: Confirmed no holdings: **basius is not launched yet**, so no basius agent has ever invested in those 6 Base tokens — there's nothing on Base to strand. That's why they don't need `LEGACY_HELD_ASSETS` treatment: the symmetry argument with oUSDT only applies where holdings can actually exist, and on Base they can't (yet).

`oUSDT` is the one real case because it's an Optimism token and optimus *is* live with agents holding it. If/when basius launches and any of those Base tokens are later de-whitelisted while held, that's the moment to add them to `LEGACY_HELD_ASSETS["base"]` — same pattern.

On drift: the Base side is pinned, not just dynamically iterated — `TestWhitelistInvariants.test_base_investable_set_is_exact` asserts `WHITELISTED_ASSETS["base"]` is exactly the curated 6, and `test_held_assets_is_superset_of_investable` covers Base too. So an accidental re-add/removal on Base fails the suite. I've also added a note to the PR description so the asymmetry doesn't read as an oversight.

LOCKhart07 (approved):

bennyjo (approved):
Re-review (head bbc6c40): the WHITELISTED_ASSETS / LEGACY_HELD_ASSETS / HELD_ASSETS split correctly fixes the oUSDT stranding on Optimism — all three detection/sweep sites read HELD_ASSETS, the investability gate stays on WHITELISTED_ASSETS, oUSDT stays detectable/convertible/priceable, and TestWhitelistInvariants pins the membership per chain (Base included). The Base de-whitelisting is a no-op since basius is not launched — no holdings to strand — so LEGACY_HELD_ASSETS["base"] isn't needed yet; the right pattern is to add a Base token there only if it's de-whitelisted while held once basius is live. Latest commit is a cosmetic D403 docstring/lock-hash change. LGTM.

PR #342, opened 2026-06-18
feat(staking): replace vanity Safe txs with Mech marketplace requests (new staking mechanism)

LOCKhart07:
## Objective

On the new staking contracts the on-chain activity counter counts **Mech marketplace requests** (`MechMarketplace.mapRequestCounts[safe]`, read by `RequesterActivityCheckerV2`) instead of Safe nonces. This ports trader [#993](https://github.com/valory-xyz/trader/pull/993)'s mechanism into Optimus/Basius: on the new regime the agent satisfies the on-chain activity check by sending a **real `MechMarketplace.request(...)`** instead of an empty **vanity Safe tx**.

Everything about *when* we fire (the period-count trigger) and *how* we settle (the existing tx-settlement FSM) is unchanged. Old-regime services are completely unaffected — the vanity-tx path is retained and used whenever the activity checker is not `RequesterActivityCheckerV2`.

One codebase, **both deployments** (Optimus/Optimism + Basius/Base), selected per-deployment via env vars keyed off `staking_chain`. **USDC** payment on both chains (per product decision).

> Draft: the core mechanism, FSM wiring, config, tests (100% on new dev code) and lint are complete. A few values need a final confirmation before merge — see **Open items** below.

## How it works

1. Reach `CheckStakingKPIMetRound` as today.
2. Detect the regime from the activity-checker `VERSION()` (`== "0.2.0"` ⇒ new), read **on the staking chain**. Cached on shared state; a transient read returns `None` and retries rather than latching "old".
3. KPI is measured in **mech requests since last checkpoint** (`mapRequestCounts` delta) on the new regime.
4. If KPI not met **and** the period-count trigger fires **and** a request is owed **and** the EOA can fund the tx → the producer injects one `MechMetadata` (static prompt + fixed tool) and hands off to the composed `mech_interact_abci` legs (`MechVersionDetectionRound → MechRequestRound`), which build the USDC `approve` + `MechMarketplace.request` multisend and settle it through the normal FSM.
5. `PostTxSettlementRound` routes back to the staking loop (same shape as the vanity tx today).
6. `is_activity_target_met` (default target **1**) is surfaced on `/healthcheck` under `agent_health` for Pearl auto-run rotation.

## Key changes

- **Vendor** the third_party `mech_interact_abci` skill + mech contracts + `valory/acn_data_share` protocol (canonical registry hashes; gitignored, pulled via `autonomy packages sync`).
- **New dev contract** `valory/mech_activity` for the `VERSION()`/`activityChecker()` reads — the third_party `staking_activity_checker` can't be modified (it'd be reverted by sync). Mirrors trader's contract, with tests.
- **`base.py`**: `_is_new_staking_regime()` (shared-state cache, fail-open on transient errors); **regime-dependent nonce index at BOTH read sites** — live `getMultisigNonces` *and* the `getServiceInfo()[2]` checkpoint snapshot move to index `1` (`mapRequestCounts`) on the new regime, `0` (Safe nonce) on the old. (This was review finding #1 — changing only one site silently breaks the KPI delta.)
- **`CheckStakingKPIMetBehaviour`**: branch on regime; on new regime inject one `MechMetadata` and emit `MECH_REQUEST_NEEDED`; funding guard + period-count trigger retained. Also computes the `agent_health` activity-target signal when staked.
- **FSM**: new `MECH_REQUEST_NEEDED` / `MECH_REQUEST_TX_EXECUTED` events + edges; `MechRequestRound` (and `MechPurchaseSubscriptionRound`, defensively) added to the post-settlement `tx_submitter → event` map so the new-regime loop doesn't die on `UNRECOGNIZED`.
- **`optimus_abci`**: chain `MechInteractAbciApp`; mix `MechParams`/mech `SharedState` into the composition `Params`/`SharedState` so the composed mech behaviours read their config off the single live `self.context`; register mech behaviours + API-spec models.
- **healthcheck** (`_handle_get_health`): adds `agent_health { is_staking_kpi_met, is_activity_target_met, activity_target, activity_completed }`; `null` on old regime / unstaked.
- **Config** (`skill.yaml` / `aea-config.yaml` / `service.yaml`): `activity_target` (1), `mech_tool`, `mech_request_prompt`, and the `mech_interact` marketplace params — all env-overridable per deployment (Optimism/USDC committed as defaults; Base overrides via env).

### A note on the response leg
The spec called for composing the Request leg only. The FSM validity check requires every composed round to be reachable, so the mech-request settlement is routed through `MechResponseRound` (poll-then-**discard** — the response *content* is never acted on, only the on-chain liveness tick matters; no answer-handling round is composed) before returning to the staking loop. This is the only deviation from "Request-only".

### A note on `multisend_address`
The shared settlement `multisend_address` was previously `0x000…0` with **no env override**. It now defaults to the Optimism multisend (`0xbE5b…`) and is overridable via the conventional **`MULTISEND_ADDRESS`** env var (review round 2). No previously-set operator var is shadowed (none existed); Base deployments override it via env.

## Review round 2 (addressed)

Following review by @OjusWiZard and @DhairyaPatel7 (commit `a58d3c61`):

- 🔴 **`VERSION()` read on the wrong chain** — the regime probe defaulted to the ethereum-mainnet ledger, so the new regime was never detected and the whole mechanism stayed inert. Now passes `chain_id=staking_chain`; regression test asserts the dispatched chain.
- 🟠 **Duplicate read / snapshot skew** — `_is_staking_kpi_met` now returns `(kpi_met, nonces_since_last_cp)` from one snapshot, reused for both the KPI verdict and the healthcheck signal (was two separate reads that could disagree).
- 🟠 **Subscription submitter** — `MechPurchaseSubscriptionRound` mapped to `MECH_REQUEST_TX_EXECUTED` defensively (unreachable on USDC, but can't drop the loop to `FailedMultiplexerRound`).
- 🟠 **Tests** — added `agent_health` handler tests, a new-regime funding-guard suppression test, mech-submitter post-settlement tests, and `liveness_ratio` guard tests.
- 🟡 **Polish** — unstaked KPI path logs INFO not ERROR; `mech_activity` class docstring + `liveness_ratio` `_NO_SUCH_FUNCTION` guard; corrected "Response leg not composed" docstrings.
- ⚙️ **Config** — `MECH_MULTISEND_ADDRESS` → `MULTISEND_ADDRESS`; `valid_mechs` default aligned to the checksummed `priority_mech_address` (comparison is case-insensitive either way).

## Validation

- ✅ `make fix-abci-app-specs` (both FSMs regenerated, composition valid — 27 rounds, no unreachable states)
- ✅ `autonomy packages lock` + `--check` verification successful
- ✅ Tests: 3161 passing across the touched suites; **100% coverage on the new dev code** (regime detection, nonce-index, producer, healthcheck data, new SynchronizedData/payload fields)
- ✅ Lint: black / isort / flake8 / darglint / pylint / mypy all clean

## Open items (need a call before merge)

1. **`mech_tool` name** — defaulted to `openai-gpt-4o-2024-08-06`; confirm the tool the configured priority mech serves (Dhairya). The response is discarded, but a v2 request must pass a valid tool string.
2. **Base USDC mech address** — USDC chosen for both chains, but at probe time Base had no live `FixedPriceTokenUSDC` mech (all 52 live Base mechs were Native). The committed defaults are Optimism's USDC mech (svc 194 `0x650C…D0d`); set `MECH_*` env for Base once its USDC mech is deployed/known.
3. **`agent_registry_address`** — placeholder `0x0`. Confirmed **inert on our path**: only read when `use_acn_for_delivers=True` (response.py), which our config disables. Still worth pointing at the real Optimism/Base agent registry before merge in case ACN delivers are ever enabled.
4. **`agent_health` nesting** — four keys nested under `agent_health` (trader's shape). Pearl reads the rotation signal on-chain (`mapRequestCounts − serviceInfo[2][1]`), so the healthcheck field is informational; confirm the shape still matches what Pearl/FE consume.
5. **Re-verify mech/marketplace addresses** on both chains right before merge (mechs were freshly deployed at spec time).


packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1551,6 +1558,98 @@ def _get_liveness_period(self, chain: str) -> Generator[None, None, Optional[int
...
+        ``None`` when the ``VERSION`` getter is genuinely absent (an old
+        checker). The contract layer swallows only "no such function" and
+        re-raises transient errors, which surface here as a non-RAW_TRANSACTION
+        performative; we must NOT treat those as "old".
+
+        :param checker_address: the activity-checker contract address.
+        :yield: contract API messages while awaiting the version read.
+        :return: a ``(version_or_none, read_ok)`` tuple.
+        """
+        response_msg = yield from self.get_contract_api_response(
```
OjusWiZard, line 1593: **🔴 Critical — VERSION() is read on the wrong chain, so the new regime is never detected.**

This `get_contract_api_response` call passes **no `chain_id`**. Every other chain-targeted read in this file passes `chain_id=chain`. The framework defaults to `default_ledger_id`, which is `ethereum` (`aea-config.yaml:85`) pointing at an **Ethereum-mainnet** RPC (`chain_id: 1`, `aea-config.yaml:139-140`). The activity checker `0x7Fd1…bf68` lives on **Optimism** (`staking_chain=optimism`).

Result: the read hits mainnet, finds no contract, `MechActivityContract.version()` swallows the `BadFunctionCallOutput` → `data=None` → `is_new = (None == "0.2.0") == False` → `staking_regime_is_new` is cached as OLD for the whole process. On the new `RequesterActivityCheckerV2` contract the agent keeps sending vanity Safe txs, never ticks `mapRequestCounts`, and drifts toward eviction — the entire mechanism this PR adds never activates. `TestStakingRegimeDetection` can't catch this because it stubs the contract response.

**Fix:** add `chain_id=self.params.staking_chain,` to the call, and add a regression test asserting the dispatch chain (assert on the kwargs, not a stubbed return).

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -92,7 +96,37 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
             is_staking_kpi_met = yield from self._is_staking_kpi_met()
+
+            # ``_is_staking_kpi_met`` returns ``None`` only when the service is
+            # not STAKED. When staked, compute the activity-target signal for the
+            # /healthcheck endpoint; it is populated on the new staking regime
+            # only (``None`` on the old regime / unstaked — see §5.4).
+            multisig_nonces_since_last_cp = None
+            is_new_regime = None
+            if is_staking_kpi_met is not None:
+                multisig_nonces_since_last_cp = (
```
OjusWiZard: **🟠 Important — duplicate on-chain read + snapshot skew.**

`_is_staking_kpi_met()` (line 107) already calls `_get_multisig_nonces_since_last_cp` internally; this calls it again. That doubles `getMultisigNonces` + `getServiceInfo` reads every staked period, and because the two reads are separate await points the KPI verdict and the `activity_completed` / `is_activity_target_met` healthcheck values can be computed from **different on-chain snapshots**.

**Fix:** have `_is_staking_kpi_met` return `(Optional[bool], Optional[int])` (verdict + nonce delta) and reuse the single read for both the KPI decision and the healthcheck signal.

packages/valory/skills/optimus_abci/handlers.py
```diff
@@ -1600,6 +1600,27 @@ def _handle_get_health(
...
             period = None
 
+        # Staking activity signal for Pearl auto-run rotation (new staking
+        # regime). ``is_activity_target_met`` is the rotation/stop signal Pearl
+        # reads; Pearl (not the agent) enacts the stop. On the old regime /
+        # unstaked these are ``None`` (the off-chain target concept does not
+        # apply) — see §5.4. The four keys mirror trader PR #993 so Pearl/FE read
+        # the same contract.
+        try:
+            agent_health = {
```
OjusWiZard, line 1610: **🟠 Important — missing tests for `agent_health`.**

This new block has zero assertions in `tests/test_handlers.py`. Happy-path tests use an unconfigured `MagicMock` for `synchronized_data` (the four keys return truthy mocks, never inspected); the error-path tests only assert `period is None`, not the all-`None` `agent_health` fallback. Add tests for: (a) `is_activity_target_met` propagating into the JSON (the Pearl rotation signal), (b) the all-`None` shape when `synchronized_data` raises, (c) the new-regime-not-yet-met case.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -134,19 +160,39 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
-                                # while the EOA cannot fund a real on-chain
-                                # action hides the funding alert behind a
-                                # green staking KPI.
+                                # Padding the activity counter while the EOA
+                                # cannot fund a real on-chain action hides the
+                                # funding alert behind a green staking KPI. The
+                                # mech request also costs real gas (+ USDC), so
+                                # suppressing it here is even more correct than
+                                # for the old vanity tx.
                                 self.context.logger.warning(
```
OjusWiZard: **🟠 Important — funding guard untested on the new-regime path.**

This suppression branch fires before any regime check, so it suppresses the **mech request** on the new regime just as it suppresses the vanity tx on the old one. The only suppression test (`test_vanity_tx_suppressed_when_balance_below_recent_real_tx_cost`) forces the old regime. Add a test with `_is_new_staking_regime=True` **and** `balance < recent_real_tx_cost`, asserting `payload.mech_requests is None` and that no tx is built.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -158,16 +204,40 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
             self.set_done()
 
+    def _build_mech_request_metadata(self) -> str:
+        """Build the JSON payload of one ``MechMetadata`` for the activity request.
+
+        On the new staking regime the agent ticks the on-chain activity counter
+        by sending a real ``MechMarketplace.request(...)`` instead of an empty
+        vanity Safe tx. We inject a single static request (fixed tool + static
+        prompt) for the composed ``mech_interact_abci`` legs to build and settle;
+        the response is discarded (the Response leg is not composed), so the
```
OjusWiZard: **🟠 Important — docstring contradicts the FSM.** “the response is discarded (the Response leg is not composed)” is wrong: `composition.py:71-73` wires `MechResponseRound` as a live, reachable round (the PR body calls this “poll-then-discard … the only deviation from Request-only”). Reword to: “the response **content** is discarded after polling — only liveness is needed, not the mech answer.” Same wording is repeated at `models.py:476`.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -92,7 +96,37 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+            is_activity_target_met = None
+            activity_target = None
+            activity_completed = None
+
+            multisig = self.params.safe_contract_addresses.get(
+                self.params.staking_chain
+            )
             is_staking_kpi_met = yield from self._is_staking_kpi_met()
+
+            # ``_is_staking_kpi_met`` returns ``None`` only when the service is
```
OjusWiZard: **🟡 Advisory — comment inaccurate.** `_is_staking_kpi_met` does **not** return `None` only when unstaked; it also returns `None` when `min_num_of_safe_tx_required is None` and when the nonce read fails. Reword to cover the transient-failure cases.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -92,7 +96,37 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+                is_new_regime = yield from self._is_new_staking_regime()
+                if is_new_regime and multisig_nonces_since_last_cp is not None:
+                    activity_target = self.params.activity_target
+                    activity_completed = multisig_nonces_since_last_cp
+                    is_activity_target_met = (
+                        multisig_nonces_since_last_cp >= activity_target
+                    )
+
             if is_staking_kpi_met is None:
                 self.context.logger.error("Error checking if staking KPI is met.")
```
OjusWiZard: **🟡 Advisory — noisy ERROR log.** When the service is simply not staked, `_is_staking_kpi_met` returns `None` and this logs ERROR every period for a normal state. Downgrade to INFO/DEBUG for the unstaked case and reserve ERROR for staked-but-read-failed.

packages/valory/contracts/mech_activity/contract.py
```diff
@@ -0,0 +1,101 @@
...
+# do NOT catch ``ValueError``: web3 raises a ``ValueError`` subclass for
+# transient RPC blips (e.g. "Skipping the filtering operation as the RPC is
+# misbehaving"), so swallowing it would mis-classify a passing failure as "old
+# checker" and cache that for the whole process lifetime. Connection / RPC /
+# timeout errors (including ``ValueError``) must propagate so the caller retries.
+_NO_SUCH_FUNCTION = (BadFunctionCallOutput, ContractLogicError)
+
+
+class MechActivityContract(Contract):
+    """The Service Staking contract."""
```
OjusWiZard: **🟡 Advisory — wrong class docstring.** Says “The Service Staking contract” (copy-paste); this is the `MechActivity` activity-checker wrapper. Also: `liveness_ratio` (line 52) lacks the `_NO_SUCH_FUNCTION` guard its siblings have — any RPC/ABI error propagates uncaught.

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -176,6 +176,15 @@ def __init__(self, *args: Any, skill_context: SkillContext, **kwargs: Any) -> No
...
         self._endpoint_breakers_lock = threading.Lock()
+        # Staking-regime detection cache (new == RequesterActivityCheckerV2).
+        # ``None`` means "not yet detected". Cached on the shared state (not on
+        # the behaviour) so it survives behaviour re-instantiation within a
+        # period; the activity-checker contract a service runs against does not
+        # change without a Pearl restart, which resets this process. This field
+        # MUST live on the live shared-state class — in the composed
+        # ``optimus_abci`` app the instantiated ``SharedState`` subclasses this
+        # one, so placing it here keeps it reachable via ``context.state``.
+        self.staking_regime_is_new: Optional[bool] = None
```
OjusWiZard, line 187: **🟡 Advisory — tri-state encoded as `Optional[bool]`.** `staking_regime_is_new` conflates “never detected” and “transient failure” (both `None`) with definitive old/new. A small `StakingRegime` enum (`UNKNOWN/OLD/NEW`) makes the cache-write invariant (“never cache a transient `None`”) explicit and unmistakable.

OjusWiZard (commented):
Automated multi-perspective review (6 specialized agents; high-severity findings verified against source). One Critical defeats the PR's core objective on the configured Optimism deployment and is masked by mocked tests. Four Important items follow. Verdict: please address C1 + I1–I4 before merge. Inline comments below.

packages/valory/services/optimus/service.yaml
```diff
@@ -55,7 +55,7 @@ models:
       keeper_timeout: 30.0
       max_attempts: 10
       max_healthcheck: 120
-      multisend_address: ${MULTISEND_ADDRESS:str:0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761}
+      multisend_address: ${MECH_MULTISEND_ADDRESS:str:0xbE5b0013D2712DC4faF07726041C27ecFdBC35AD}
```
DhairyaPatel7: Breaking change: env var renamed silently. `multisend_address: ${MULTISEND_ADDRESS:str:0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761}` was changed to `${MECH_MULTISEND_ADDRESS:str:0xbE5b0013D2712DC4faF07726041C27ecFdBC35AD}`. Any existing deployment that sets `MULTISEND_ADDRESS` (and Pearl/operator scripts likely do) will silently fall back to the new default `0xbE5b…` instead of their configured value, and the default address itself also changed. The same rename is in `packages/valory/agents/optimus/aea-config.yaml:254`. This isn't in the PR description's Key Changes. Either keep the env var name as `MULTISEND_ADDRESS` (the mech multisend is the same `multisend_address` param anyway — both skills read the same key), or document the rename + add an explicit migration note in the PR description so deployers update their env.

packages/valory/skills/optimus_abci/handlers.py
DhairyaPatel7, line 1610: Missing test coverage on the new `agent_health` block. The PR description claims '100% coverage on the new dev code', but `tests/test_handlers.py` is unchanged and grep for `agent_health` there is empty. Add a `_handle_get_health` test that asserts the four keys (`is_staking_kpi_met`, `is_activity_target_met`, `activity_target`, `activity_completed`) appear in the response and that the `(AttributeError, ValueError)` fallback returns all-None — Pearl's auto-run rotation depends on this contract.

packages/valory/skills/liquidity_trader_abci/states/post_tx_settlement.py
```diff
@@ -73,6 +74,10 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
             submitter_to_event: Dict[str, Event] = {
                 CallCheckpointRound.auto_round_id(): Event.CHECKPOINT_TX_EXECUTED,
                 CheckStakingKPIMetRound.auto_round_id(): Event.VANITY_TX_EXECUTED,
+                # New regime: the activity tx is submitted by mech_interact_abci's
+                # MechRequestRound, so it must map back to the staking loop or the
+                # multiplexer returns UNRECOGNIZED and the loop dies.
+                MechRequestRound.auto_round_id(): Event.MECH_REQUEST_TX_EXECUTED,
```
DhairyaPatel7, line 83: Incomplete submitter→event map: `MechPurchaseSubscriptionRound` is reachable as a tx submitter via `(MechRequestRound, BUY_SUBSCRIPTION): MechPurchaseSubscriptionRound → RandomnessTransactionSubmissionRound`, but only `MechRequestRound.auto_round_id()` is added here. If the configured priority mech ever switches to a subscription-type mech, the post-settlement lookup returns `UNRECOGNIZED` → `FailedMultiplexerRound` → ResetAndPause, and the new-regime staking loop silently falls out — exactly the failure mode this PR's review-#2 fix was meant to prevent. Unreachable with the current USDC mech config, but worth adding `MechPurchaseSubscriptionRound.auto_round_id(): Event.MECH_REQUEST_TX_EXECUTED` defensively (or assert it's blocked at the regime gate).

packages/valory/agents/optimus/aea-config.yaml
```diff
@@ -301,6 +319,22 @@ models:
...
+      multisend_batch_size: ${MECH_MULTISEND_BATCH_SIZE:int:1}
+      mech_contract_address: ${MECH_CONTRACT_ADDRESS:str:0x650C77F49cFa69e2AC1990A6B4585C322B012D0d}
+      ipfs_address: ${MECH_IPFS_ADDRESS:str:https://gateway.autonolas.tech/ipfs/}
+      mech_chain_id: ${MECH_CHAIN_ID:str:optimism}
+      mech_wrapped_native_token_address: ${MECH_WRAPPED_NATIVE_TOKEN_ADDRESS:str:0x4200000000000000000000000000000000000006}
+      mech_interaction_sleep_time: ${MECH_INTERACTION_SLEEP_TIME:int:10}
+      use_mech_marketplace: ${USE_MECH_MARKETPLACE:bool:true}
+      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x46C0D07F55d4F9B5Eed2Fc9680B5953e5fd7b461","priority_mech_address":"0x650C77F49cFa69e2AC1990A6B4585C322B012D0d","priority_mech_service_id":194,"response_timeout":300,"use_dynamic_mech_selection":false}}
+      agent_registry_address: ${MECH_AGENT_REGISTRY_ADDRESS:str:0x0000000000000000000000000000000000000000}
+      use_acn_for_delivers: ${USE_ACN_FOR_DELIVERS:bool:false}
```
DhairyaPatel7, line 334: `valid_mechs` default is lowercase (`["0x650c77f49cfa69e2ac1990a6b4585c322b012d0d"]`) but `priority_mech_address` is mixed-case (`0x650C77F49cFa69e2AC1990A6B4585C322B012D0d`). If `mech_interact_abci` does a case-sensitive membership check, the configured priority mech will be rejected as invalid. Same mismatch in `service.yaml`. Verify the comparison normalizes case (lowercase or checksum), or align both defaults to the same form.

packages/valory/agents/optimus/aea-config.yaml
```diff
@@ -301,6 +319,22 @@ models:
       staking_token_contract_address: ${str:0x88996bbdE7f982D93214881756840cE2c77C4992}
       staking_activity_checker_contract_address: ${str:0x7Fd1F4b764fA41d19fe3f63C85d12bf64d2bbf68}
       staking_threshold_period: ${int:5}
+      activity_target: ${ACTIVITY_TARGET:int:1}
+      mech_tool: ${MECH_TOOL:str:openai-gpt-4o-2024-08-06}
+      mech_request_prompt: ${MECH_REQUEST_PROMPT:str:Optimus staking activity request.}
+      multisend_batch_size: ${MECH_MULTISEND_BATCH_SIZE:int:1}
```
DhairyaPatel7, line 325: `agent_registry_address` defaults to `0x0000000000000000000000000000000000000000`. Already called out as an open item in the PR description, but flagging here so it isn't lost: a deployment that forgets `MECH_AGENT_REGISTRY_ADDRESS` ships with the zero address; if `mech_interact_abci` ever calls into it (even on a non-USDC path), this fails opaquely. Either point at the real Optimism/Base agent registry before merge, or hard-fail at param init when it's `0x0` and the V2 path is selected.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
LOCKhart07: Fixed in a58d3c61. `_get_checker_version` now passes `chain_id=self.params.staking_chain`. Confirmed the ledger connection's contract dispatcher consumes `chain_id` to select the RPC (same mechanism `contract_interact` relies on; `ledger_id` only selects the plugin family). Added `test_get_checker_version_dispatches_to_staking_chain` which asserts the dispatched `chain_id` kwarg rather than a stubbed return.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
LOCKhart07: Fixed in a58d3c61. `_is_staking_kpi_met` now returns `(kpi_met, nonces_since_last_cp)` from a single snapshot; the behaviour reuses that delta for both the KPI verdict and the healthcheck signal, so the second `getMultisigNonces`+`getServiceInfo` read (and the snapshot skew) are gone. The now-provably-dead `is not None` guards on the delta were removed (the tuple couples verdict+delta) to keep 100% branch coverage; `TestIsStakingKpiMet` updated for the tuple.

packages/valory/skills/optimus_abci/handlers.py
LOCKhart07: Fixed in a58d3c61. Added `agent_health` tests in test_handlers.py: keys propagate (incl. `is_activity_target_met`), a new-regime-not-yet-met case (`is_activity_target_met=False`), and the all-`None` fallback when `synchronized_data` raises (extended the two pre-FSM error tests to assert the dict shape).

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
LOCKhart07: Fixed in a58d3c61. Added `test_new_regime_funding_guard_suppresses_mech_request`: `_is_new_staking_regime=True` + balance < recent real-tx cost asserts `payload.mech_requests is None`, no vanity tx, and the funding WARNING.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
LOCKhart07: Fixed in a58d3c61. Reworded both here and at models.py: the Response leg IS composed (poll-then-discard) — the response *content* is discarded after polling, only the on-chain liveness tick matters.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
LOCKhart07: Fixed in a58d3c61. Reworded; the comment now covers all undetermined cases (not staked, `min_num_of_safe_tx_required` unknown, transient nonce-read failure) — folded into the tuple refactor.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
LOCKhart07: Fixed in a58d3c61. The unstaked/undetermined path now logs INFO; the genuine staked-but-read-failed case still logs ERROR inside `_is_staking_kpi_met` / the nonce helpers.

packages/valory/contracts/mech_activity/contract.py
LOCKhart07: Fixed in a58d3c61. Corrected the class docstring (it's the Mech activity-checker wrapper) and wrapped `liveness_ratio` in the same `_NO_SUCH_FUNCTION` guard as its siblings, with absent-returns-None and transient-error-propagates tests.

packages/valory/skills/liquidity_trader_abci/models.py
LOCKhart07: Acknowledged — declining for now. The 'never cache a transient None' invariant is already enforced and documented at every write site (`_is_new_staking_regime` returns None without caching on a transient read). A `StakingRegime` enum would touch the cache field + 3 base.py sites + tests for a cosmetic gain on a draft; happy to do it as a follow-up if you'd prefer it in this PR.

packages/valory/services/optimus/service.yaml
LOCKhart07: Fixed in a58d3c61 — renamed the env var to `MULTISEND_ADDRESS` in both aea-config.yaml and service.yaml, keeping the real Optimism default. One correction on the history: git shows the pre-PR value was `${str:0x0000000000000000000000000000000000000000}` — a hardcoded zero with **no env var at all**, so nothing was silently renamed and no existing `MULTISEND_ADDRESS` was shadowed. But your underlying point stands: it's a shared settlement param, so `MULTISEND_ADDRESS` is the correct conventional name (both skills read the same `multisend_address` key).

packages/valory/skills/optimus_abci/handlers.py
LOCKhart07: Fixed in a58d3c61. test_handlers.py now asserts the four keys propagate into the response, the all-`None` fallback on the `(AttributeError, ValueError)` path, and a new-regime-not-yet-met case (`is_activity_target_met=False`) — the Pearl rotation contract.

packages/valory/skills/liquidity_trader_abci/states/post_tx_settlement.py
LOCKhart07: Fixed in a58d3c61 — added `MechPurchaseSubscriptionRound.auto_round_id(): Event.MECH_REQUEST_TX_EXECUTED` defensively, with tests for both mech submitters (request and subscription). Unreachable with the current USDC config, but it can no longer drop the staking loop to FailedMultiplexerRound if a subscription-type mech is ever configured.

packages/valory/agents/optimus/aea-config.yaml
LOCKhart07: Not a bug — the comparison is case-insensitive: `valid_mechs` is lowercased on load (mech_interact_abci/models.py:300) and the membership test uses `static_priority.lower()` (behaviours/request.py:602). Aligned the `valid_mechs` default to the checksummed `priority_mech_address` form anyway in a58d3c61 so the two read identically.

packages/valory/agents/optimus/aea-config.yaml
LOCKhart07: Inert on our configured path: `agent_registry_address` is only read via `agent_registry_contract_interact` from response.py:371, which runs only when `use_acn_for_delivers=True` — and our config sets `use_acn_for_delivers: false`. So the zero default cannot be hit on the USDC marketplace-v2 request path. Leaving the real Optimism/Base registry address as a pre-merge item (needs the team); flagging that a hard-fail-at-init guard would have to live in the vendored mech_interact model, which I'd rather not patch here. Happy to add it if you want belt-and-suspenders.

packages/valory/agents/optimus/aea-config.yaml
```diff
@@ -233,7 +251,7 @@ models:
       keeper_timeout: 30.0
       max_attempts: 10
       max_healthcheck: 120
-      multisend_address: ${str:0x0000000000000000000000000000000000000000}
+      multisend_address: ${MULTISEND_ADDRESS:str:0xbE5b0013D2712DC4faF07726041C27ecFdBC35AD}
```
OjusWiZard, line 254: **🟠 Verify before merge — shared settlement multisend default changed.**

This `multisend_address` sits in the tx-settlement/termination block, so it's the multisend used by **every** settled tx — the existing Balancer/Uniswap/Velodrome trading path, not just the new mech path. On `main` the default was `0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761` (canonical Safe MultiSend v1.3.0); this PR changed it to `0xbE5b0013D2712DC4faF07726041C27ecFdBC35AD`, which isn't a standard canonical Safe MultiSend address. The fix commit reverted the env-var name (good) but kept the new default, and the address change isn't in the PR's Key changes.

If `0xbE5b0013…` isn't a valid MultiSend(CallOnly) on Optimism/Base, existing trading settlement regresses. Please confirm it's correct on both chains, or restore `0xA238…`. Same default in `service.yaml:58`. (Non-blocking for approval, but worth confirming.)

packages/valory/agents/optimus/aea-config.yaml
```diff
@@ -301,6 +319,22 @@ models:
...
+      mech_request_prompt: ${MECH_REQUEST_PROMPT:str:Optimus staking activity request.}
+      multisend_batch_size: ${MECH_MULTISEND_BATCH_SIZE:int:1}
+      mech_contract_address: ${MECH_CONTRACT_ADDRESS:str:0x650C77F49cFa69e2AC1990A6B4585C322B012D0d}
+      ipfs_address: ${MECH_IPFS_ADDRESS:str:https://gateway.autonolas.tech/ipfs/}
+      mech_chain_id: ${MECH_CHAIN_ID:str:optimism}
+      mech_wrapped_native_token_address: ${MECH_WRAPPED_NATIVE_TOKEN_ADDRESS:str:0x4200000000000000000000000000000000000006}
+      mech_interaction_sleep_time: ${MECH_INTERACTION_SLEEP_TIME:int:10}
+      use_mech_marketplace: ${USE_MECH_MARKETPLACE:bool:true}
+      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x46C0D07F55d4F9B5Eed2Fc9680B5953e5fd7b461","priority_mech_address":"0x650C77F49cFa69e2AC1990A6B4585C322B012D0d","priority_mech_service_id":194,"response_timeout":300,"use_dynamic_mech_selection":false}}
+      agent_registry_address: ${MECH_AGENT_REGISTRY_ADDRESS:str:0x0000000000000000000000000000000000000000}
```
OjusWiZard, line 333: **🟡 Open item (tracked) — `agent_registry_address` defaults to `0x0`.** A deployment that forgets `MECH_AGENT_REGISTRY_ADDRESS` ships with the zero address; if `mech_interact_abci` ever calls into it (even off the USDC path) it fails opaquely. Suggest hard-failing at param init when it's `0x0` while the V2 path is selected, rather than deferring to a runtime revert. Same default in `service.yaml:95`.

OjusWiZard (dismissed):
Approving. All findings from the earlier review round are addressed (chain-scoped VERSION read, single-snapshot KPI read, docstring/comment/log fixes, `agent_health` + new-regime funding-guard tests, `post_tx_settlement` submitter map, `valid_mechs` casing). One non-blocking item to verify before merge — the shared `multisend_address` default change — plus the already-tracked `agent_registry_address` open item; both inline below.

packages/valory/agents/optimus/aea-config.yaml
DhairyaPatel7: Thanks @OjusWiZard — dug into this, and it shouldn't regress the trading path:

- **Consumer:** the single `multisend_address` (set under the `optimus_abci` params block) feeds `mech_interact_abci` — `models.py:261-263` `enforce(...)` and `behaviours/base.py:139/338/377`, i.e. the mech-request path. The Balancer/Uniswap/Velodrome trading settlement instead reads the per-chain map `multisend_contract_addresses[chain]` (`pools/uniswap.py:273`, `pools/velodrome.py`), so it isn't affected by this default.
- **Address:** `0xbE5b0013D2712DC4faF07726041C27ecFdBC35AD` is the Optimism Safe MultiSend — it's exactly the `optimism` entry in `multisend_contract_addresses` (Base/Mode use `0x998739BFdAAdde7C933B942a68053933098f9EDa`) and what production Optimus (Optimism) deployments already run.
- **vs. main:** on `main`/merge-base this line was `${str:0x0000…0000}` (zero address, no env binding) rather than `0xA238…`. This PR adds the `MULTISEND_ADDRESS` env binding **and** the Optimism default, so it stays overridable per chain.

Net: correct for the Optimism target and overridable via `MULTISEND_ADDRESS`; the trading path keys off the per-chain map regardless. Carried through unchanged in the `main` merge.

packages/valory/agents/optimus/aea-config.yaml
DhairyaPatel7: Agreed this is worth hardening. Default is `0x0` with the `MECH_AGENT_REGISTRY_ADDRESS` override, consumed by `mech_interact_abci` (`models.py:290-295`, `behaviours/base.py:272`). A hard-fail at param init — when the V2/mech path is selected and the address is still `0x0` — is a reasonable guard so a misconfigured deployment fails fast instead of reverting opaquely at runtime. Tracking it as a follow-up; @LOCKhart07 can fold it into this PR or a separate one. Value preserved as-is through the `main` merge.

DhairyaPatel7 (approved):

OjusWiZard (approved):
Re-approving on the current HEAD (post `main` merge `7452cc3e`). Verified the merge cleanly preserves both the mech-staking code and main's whitelist work in `base.py`, the chain-scoped VERSION read survives, package hashes are consistent, and the key modules parse. All review threads (mine and @DhairyaPatel7's) are resolved or answered. Remaining open items are intentional follow-ups (the `agent_registry_address` 0x0 hard-fail guard) and the PR's own pre-merge product confirmations (mech_tool name, Base USDC mech address, Pearl agent_health nesting, address re-verification).

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -105,69 +148,104 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+                        self.context.logger.info(
+                            f"Number of tx left to meet KPI: {num_of_tx_left_to_meet_kpi}"
                         )
-                        if num_of_tx_left_to_meet_kpi > 0:
+                        signal = self._real_tx_cost_vs_balance(
+                            chain=self.params.staking_chain  # type: ignore[arg-type]
+                        )
+                        if (
+                            signal is not None
+                            and signal.eoa_balance < signal.recent_real_tx_cost
```
bennyjo, line 169: **New-regime activity tx checks only native gas, not USDC**

This affordability guard reads the EOA's native balance (`get_funds_status` → `ZERO_ADDRESS` key) vs recent native tx cost. On the old regime that was complete — the vanity tx only cost gas. On the new regime the action it gates is a `MechMarketplace.request` that also spends **USDC** (`approve` + `request`), and the comment here even acknowledges "costs real gas (+ USDC)" — but USDC is never checked.

Failure case: a Safe with gas but little/no USDC passes this guard, builds the request, and settlement reverts on the payment leg → `mapRequestCounts` never ticks → KPI stays unmet → the request is rebuilt next period, burning gas in a loop with no distinct signal (the gas-based "activity tx suppressed" warning never fires). Prolonged KPI failure is what leads to eviction, so this can silently degrade a staked service.

`mech_price` isn't known at this point (it's read inside `mech_interact_abci` after version detection), so a fix needs an explicit Safe-USDC balance read here before firing on the new regime — symmetric to the native-gas guard, with its own suppression warning. Alternatively, if operator/Pearl funding tooling is expected to guarantee USDC, could we document that assumption rather than leave it implicit?

bennyjo (commented):
One funding-affordability gap on the new staking regime — see inline.

PR #343, opened 2026-06-19
fix(optimus_abci): order MechParams before TerminationParams so multisend_address resolves

DhairyaPatel7:
## Problem

The release-only `make check-agent-runner` step fails on `v0.12.0-rc2` (the first release to include the new mech path from #342) when the packaged agent loads:

```
File "vendor/valory/skills/mech_interact_abci/models.py", line 262, in __init__
    enforce(multisend_address is not None, "Multisend address not specified!")
aea.exceptions.AEAEnforceError: Multisend address not specified!
[FAIL] Did not find 'Starting AEA' within 80s
```

This isn't caught by the normal PR CI (the binary runner check only runs in the release flow) or by unit tests (they construct params from mocks).

## Root cause

`optimus_abci`'s composed `Params` mixes in both `TerminationParams` and `MechParams`, and **both consume `multisend_address`**:

- `TerminationParams.__init__` reads it via `self._ensure("multisend_address", kwargs, str)` — and `_ensure` **pops** the key off `kwargs`.
- `MechParams.__init__` reads it non-destructively via `kwargs.get("multisend_address")`.

The MRO was `Params(TerminationParams, LiquidityTraderParams, MechParams)`, so `TerminationParams` ran first and popped the key; by the time `MechParams` (last) ran, `kwargs.get` returned `None` and its `enforce` raised. (`liquidity_trader` is unaffected — it uses the per-chain `multisend_contract_addresses` map, not this single param.)

## Fix

Reorder the MRO so `MechParams` precedes `TerminationParams`:

```python
class Params(MechParams, TerminationParams, LiquidityTraderParams):
```

`MechParams` now reads `multisend_address` (non-destructively) before `TerminationParams` pops it. Hashes relocked accordingly (`optimus_abci` / agent / service).

## Verification (local)

- Rebuilt the agent from the fixed packages and ran it with only `STORE_PATH` set (as `check-agent-runner` does) → reaches **`Starting AEA … in 'async' mode`**, no `Multisend address not specified!`.
- `optimus_abci` + `liquidity_trader` tests: **pass**.
- `tomte tox`: `check-hash`, `check-packages`, `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`, `black-check`, `isort-check`, `flake8`, `darglint`, `mypy`, `pylint` — **all green**.

### DIFF
--- a/packages/valory/skills/optimus_abci/models.py
+++ b/packages/valory/skills/optimus_abci/models.py
@@ -95,15 +95,24 @@ MULTIPLIER = 40
 
 
 class Params(  # pylint: disable=too-many-ancestors
+    MechParams,
     TerminationParams,
     LiquidityTraderParams,
-    MechParams,
 ):
     """A model to represent params for multiple abci apps.
 
     Also mixes in ``MechParams`` so the composed ``mech_interact_abci``
     behaviours can read their marketplace/request config off the single shared
     ``self.context.params`` in this composition skill.
+
+    ``MechParams`` is listed first on purpose. Both it and ``TerminationParams``
+    consume ``multisend_address``, but ``TerminationParams`` reads it via
+    ``_ensure`` (which *pops* the key off ``kwargs``) while ``MechParams`` reads
+    it non-destructively via ``kwargs.get``. With ``TerminationParams`` ahead of
+    ``MechParams`` in the MRO the pop runs first, so ``MechParams`` sees ``None``
+    and ``enforce(multisend_address is not None)`` raises "Multisend address not
+    specified!" at agent load. Running ``MechParams`` first lets it read the
+    value before ``TerminationParams`` pops it.
     """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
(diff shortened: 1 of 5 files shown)


OjusWiZard (approved):

packages/valory/skills/optimus_abci/models.py
```diff
@@ -95,15 +95,24 @@
```
bennyjo, line 96: This lands without a fast-feedback regression test — per the PR description the bug is invisible to unit tests (params built from mocks) and to normal PR CI (the binary `check-agent-runner` runs only in the release flow), which is why it surfaced as late as `v0.12.0-rc2`. Would you consider adding a unit test that instantiates this composed `Params` with a representative full kwargs dict (including `multisend_address`) and asserts it constructs without raising and that `params.multisend_address` is set? That locks the MRO/kwarg-pop contract so a future reorder of these bases fails in PR CI rather than only at release-time agent boot — which is also the only thing currently defending the upstream asymmetry the docstring documents (mech-interact's `kwargs.get` vs termination's popping `_ensure`).

bennyjo (approved):
Approving — correct, minimal MRO fix; root-cause analysis matches the code. One non-blocking suggestion on regression-test coverage inline.

PR #344, opened 2026-06-22
feat(basius): add valory/basius agent and service packages

DIvyaNautiyal07:
## Summary
- Olas AgentRegistry token 115 (Basius) was minted with on-chain publicId `valory/basius`. Pearl's deploy path runs `autonomy/cli/helpers/chain.py:240 verify_service_dependencies`, which compares the subgraph's `publicId` for the agent token against the local service package's `agent:` field, both stripped to `author/name`. The existing optimus service declares `agent: valory/optimus`, so Basius deploys fail with `Public ID \`valory/basius:any\` for token 115 does not match with the one defined in the service package \`valory/optimus:any\``.
- Adds a new `valory/basius` agent package that is functionally identical to `valory/optimus` (same skills, connections, contracts, protocols, customs, handlers, env-var overrides) with only the agent identity changed. The shared skills (`optimus_abci`, `liquidity_trader_abci`, `funds_manager`, etc.) are unchanged.
- Adds a new `valory/basius` service package that references the new agent and bakes Base-chain defaults for the operational params so a non-Pearl deploy works without overrides. Pearl will still override every relevant env var via its `env_variables` block.
- `autonomy packages lock` registers both new packages as dev. Packages pushed to IPFS so CI sync resolves.
- Whitelists the two new directories in `.gitignore` (the existing `packages/valory/{agents,services}/*` ignore had explicit allow-lines only for `optimus` / `optimus_pearl`).

## Why a new agent package, not just a new service.yaml
The trader pattern (one agent package + many service packages all pointing to the same `agent:`) works because trader has a single on-chain agent identity. Basius and Optimus are two distinct on-chain identities (tokens 115 and 40), and `verify_service_dependencies` enforces a strict `author/name` match with no aliasing. So the local agent package has to mirror the on-chain identity. Skill code stays shared via the existing optimus_abci/liquidity_trader_abci skills.

## Base-chain defaults that are baked in
- `target_investment_chains` → `["base"]`
- `staking_chain` → `base`
- `initial_assets` → base USDC default
- `mech_chain_id` → `base`
- `network_selector` (coingecko) → `base`
- `service_endpoint_base` → `https://base.autonolas.tech/`
- `tendermint_p2p_url` host → `basius_tm_0`
- `service_id` → `basius`
- `funds_manager` block: `fund_requirements`, `rpc_urls`, `safe_contract_addresses` all use the `base` chain key. This is the line that the `funds_manager._validate_chain_keys` check actually reads.
- `genai_x402_server_base_url` → `…gemini/base/v1beta`

## Out of scope and untouched
- Multi-chain reference maps (`allowed_chains`, `balancer_graphql_endpoints`, `chain_to_platform_id_mapping`, `available_strategies`, etc.) keep all chains. The agent code performs per-chain lookups against these maps regardless of which chain is currently active, so stripping them would risk runtime KeyErrors.
- `optimus_abci` skill reference (`public_id: valory/optimus_abci:0.1.0`) is intentional — the skill code is shared between both agents.
- Pearl follow-up needed in `olas-operate-app`: point `BASIUS_SERVICE_TEMPLATE.hash` to the new basius service hash `bafybeiehwn2lc773hlwybuo6hwce556pnozk7z7c3cn25reegaefzcaoe4` and update `BASIUS_TEMPLATE_RELEASE.agent_release.name` from `optimus` to `basius`. My earlier `FUND_REQUIREMENTS` env-var PR (operate-app PR 2041) stays valid layered on top.

## Test plan
- [x] `autonomy packages lock --check` passes
- [x] Both new packages pushed to IPFS
- [ ] Pearl rc with the updated template hash boots Basius past `verify_service_dependencies` (the specific failure from rc14)
- [ ] Optimus deploys unchanged (no behavioural change — only a new sibling package was added)

### DIFF
--- a/packages/valory/services/basius/service.yaml
+++ b/packages/valory/services/basius/service.yaml
@@ -1,0 +1,239 @@
+name: basius
+author: valory
+version: 0.1.0
+description: A base liquidity trader service.
+aea_version: '>=2.0.0, <3.0.0'
+license: Apache-2.0
+fingerprint: {}
+fingerprint_ignore_patterns: []
+agent: valory/basius:0.1.0:bafybeicvsp5zraysyjg4lhq7hg2igcwax7mmuogmi4ertu3swsnovxiyey
+number_of_agents: 1
+deployment:
+  agent:
+    ports:
+      0:
+        8716: 8716
+    volumes:
+      ./data: /data
+dependencies:
+  requests_toolbelt:
+    version: ==1.0.0
+---
+public_id: valory/optimus_abci:0.1.0
+type: skill
+models:
+  benchmark_tool:
+    args:
+      log_dir: ${LOG_DIR:str:/logs}
+  params:
+    args:
+      setup:
+        all_participants: ${ALL_PARTICIPANTS:list:[]}
+        consensus_threshold: null
+        safe_contract_address: ${SAFE_CONTRACT_ADDRESS:str:0x0000000000000000000000000000000000000000}
+      genesis_config:
+        genesis_time: '2022-09-26T00:00:00.000000000Z'
+        chain_id: chain-c4daS1
+        consensus_params:
+          block:
+            max_bytes: '22020096'
+            max_gas: '-1'
+            time_iota_ms: '1000'
+          evidence:
+            max_age_num_blocks: '100000'
+            max_age_duration: '172800000000000'
+            max_bytes: '1048576'
+          validator:
+            pub_key_types:
+            - ed25519
+          version: {}
+        voting_power: '10'
+      cleanup_history_depth: 1
+      cleanup_history_depth_current: null
+      drand_public_key: 868f005eb8e6e4ca0a47c8a77ceaa5309a47978a7c71bc5cce96366b5d7a569937c529eeda66c7293784a9402801af31
+      earliest_block_to_monitor: ${EARLIEST_BLOCK_TO_MONITOR:int:8053690}
+      keeper_timeout: 30.0
+      max_attempts: 10
+      max_healthcheck: 120
+      multisend_address: ${MULTISEND_ADDRESS:str:0xbE5b0013D2712DC4faF07726041C27ecFdBC35AD}
+      termination_sleep: ${TERMINATION_SLEEP:int:900}
+      reset_pause_duration: ${RESET_PAUSE_DURATION:int:30}
+      on_chain_service_id: ${ON_CHAIN_SERVICE_ID:int:null}
+      reset_tendermint_after: ${RESET_TENDERMINT_AFTER:int:30}
+      retry_attempts: 400
+      retry_timeout: 3
+      request_retry_delay: 1.0
+      request_timeout: 10.0
+      round_timeout_seconds: 30.0
+      service_id: basius
+      service_registry_address: ${SERVICE_REGISTRY_ADDRESS:str:0x48b6af7B12C71f09e2fC8aF4855De4Ff54e775cA}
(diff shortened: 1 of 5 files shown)


OjusWiZard (approved):

iasonrovis (approved):

packages/valory/services/basius/service.yaml
```diff
@@ -0,0 +1,239 @@
...
+      tls_verify: ${TLS_VERIFY:bool:true}
+      mode_conduit_explorer_url: ${MODE_CONDUIT_EXPLORER_URL:str:https://explorer-mode-mainnet-0.t.conduit.xyz/api/v2}
+      safe_api_url: ${SAFE_API_URL:str:https://rpc-gate.autonolas.tech/safe/tx-service}
+      safe_api_chain_slugs: ${SAFE_API_CHAIN_SLUGS:str:{"optimism":"oeth","base":"base"}}
+      mode_native_explorer_url: ${MODE_NATIVE_EXPLORER_URL:str:https://explorer.mode.network/api}
+  coingecko:
+    args:
+      token_price_endpoint: ${COINGECKO_TOKEN_PRICE_ENDPOINT:str:/api/v3/simple/token_price/{asset_platform_id}?contract_addresses={token_address}&vs_currencies=usd}
+      coin_price_endpoint: ${COINGECKO_COIN_PRICE_ENDPOINT:str:/api/v3/simple/price?ids={coin_id}&vs_currencies=usd}
+      api_key: ${COINGECKO_API_KEY:str:CG-mf5xZnGELpSXeSqmHDLY2nNU}
```
bennyjo, line 148: Hardcoded CoinGecko key as the default. optimus uses `${COINGECKO_API_KEY:str:null}` in both its service and agent config, and basius's own agent config already uses `null` — this one file is the only place a literal key ships to repo/IPFS. Suggest reverting to `null` and rotating the key.
```suggestion
      api_key: ${COINGECKO_API_KEY:str:null}
```

packages/valory/services/basius/service.yaml
```diff
@@ -0,0 +1,239 @@
...
+type: connection
+config:
+  ledger_apis:
+    ethereum:
+      address: ${ETHEREUM_LEDGER_RPC:str:https://virtual.mainnet.rpc.tenderly.co/85a9fd10-356e-4526-b1f6-7148366bf227}
+      chain_id: ${ETHEREUM_LEDGER_CHAIN_ID:int:1}
+      poa_chain: ${ETHEREUM_LEDGER_IS_POA_CHAIN:bool:false}
+      default_gas_price_strategy: ${ETHEREUM_LEDGER_PRICING:str:eip1559}
+    base:
+      address: ${BASE_LEDGER_RPC:str:https://virtual.arbitrum.rpc.tenderly.co/8973f254-1594-4a82-8e26-25a10a01bf46}
```
bennyjo, line 171: base/optimism/mode default to private Tenderly virtual forks, and base + optimism point at the **same Arbitrum fork** while `chain_id` stays `8453`/`10` — `eth_chainId` won't match, so a non-Pearl deploy breaks. optimus uses `mainnet.base.org` / `mainnet.optimism.io` / `mainnet.mode.network` here. Suggestion fixes base inline; please apply the same to optimism (`:176`) and mode (`:181`). (ethereum is unchanged from optimus — no action needed there.)
```suggestion
      address: ${BASE_LEDGER_RPC:str:https://mainnet.base.org}
```

bennyjo (commented):
Two follow-up items — both are divergences from the optimus baseline that look like leaked local-test config.

PR #349, opened 2026-06-23
fix(basius): KPI mech-request wiring + Base launch fixes

DIvyaNautiyal07:
## Summary

- Fixes the staking KPI activity-tick on the new `RequesterActivityCheckerV2` regime. `CheckStakingKPIMetPayload` declared `event` mid-list, ahead of `mech_requests` and the activity-target fields. `CollectSameUntilThresholdRound.end_block` projects consensus via `dict(zip(selection_key, payload.values))` and `values` walks dataclass fields in declaration order, so `event`'s value was overwriting `mech_requests` in `synced_data` and the round emitted `STAKING_KPI_NOT_MET` instead of `MECH_REQUEST_NEEDED`. The composed mech leg never fired and `mapRequestCounts` stayed at zero. Moves `event` to the last declared field (matching `CallCheckpointPayload`'s shape) so trailing zip-truncation drops it cleanly.
- Drops the EOA-funded gate from the new-regime mech branch in `CheckStakingKPIMetBehaviour`. The gate guarded the old vanity-tx branch (padding the activity counter while the EOA cannot fund a real on-chain action would hide a funding alert behind a green KPI). Mech-marketplace requests settle via the Safe and are paid in USDC via x402, so EOA balance is the wrong gate for the new regime. Old-regime vanity tx still gates on EOA balance.
- Adds the five `mech_interact_abci` rounds composed into the FSM to `rounds_info` so the FSM display is complete.
- Points Basius Base defaults at the right deployments: mech-marketplace address, service registry address, multisend address, plus a Basius-branded mech request prompt.

Side fixes from the same Basius launch log:
- Casts `entry_cost` to `Decimal` before the recovery-percentage divide in `_update_position_with_current_value`. Production stores `entry_cost` as `float`, which raised `TypeError: unsupported operand type(s) for /: 'decimal.Decimal' and 'float'` and zeroed every yield report through the `except` branch.
- Adds three missing Base stablecoin CoinGecko IDs (`msusd → metronome-synth-usd`, `frxusd → frax-usd`, `axlusdc → axlusdc`). Each verified against `api.coingecko.com/api/v3/coins/base/contract/<addr>`. Without these, IL risk scoring on Base dropped every USDC pair where the other side was one of the three stables.

## Test plan
- [ ] Full skill suite passes locally (2830 tests in `liquidity_trader_abci`).
- [ ] New regression tests pass and fail under mutation (verified by reverting the payload field order — both `test_payload_values_align_with_selection_key` and `test_end_block_emits_mech_request_needed_via_real_payload_projection` fail with the exact production shape: ``assert None == '[{"prompt": ...}]'``).
- [ ] Re-run the Basius Base launch agent past one liveness window; expect `MechVersionDetectionRound` entered after `CheckStakingKPIMetRound` and `mapRequestCounts` incrementing on-chain.


iasonrovis (approved):

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -198,12 +178,30 @@ def async_act(self) -> Generator:  # type: ignore[override]
                             )
                             mech_requests_json = self._build_mech_request_metadata()
```
OjusWiZard, line 179: Advisory (observability): decoupling the mech request from EOA balance is the right call (Safe-settled + USDC via x402). But the Safe exec still consumes some native gas at settlement, and the new regime now emits no funding signal at all. Consider a non-gating `logger.warning` here when `_real_tx_cost_vs_balance` reports the EOA below recent cost, purely for operator visibility.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -198,12 +178,30 @@ def async_act(self) -> Generator:  # type: ignore[override]
...
+                            # hide a funding alert behind a green staking KPI.
+                            signal = self._real_tx_cost_vs_balance(
                                 chain=self.params.staking_chain  # type: ignore[arg-type]
                             )
-                            self.context.logger.info(f"tx hash: {vanity_tx_hex}")
+                            if (
+                                signal is not None
+                                and signal.eoa_balance < signal.recent_real_tx_cost
+                            ):
+                                self.context.logger.warning(
```
OjusWiZard, line 191: Advisory (pre-existing): when the under-funded EOA suppresses the vanity tx, the only trace is one WARNING per period; `is_activity_target_met` is populated only on the new regime, so `/healthcheck` can't distinguish "suppressed for funding" from "no action owed." Consider setting `is_activity_target_met=False` on suppression so the shortfall is observable. Non-blocking follow-up.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
```diff
@@ -1480,10 +1480,13 @@ def _update_position_with_current_value(
                     else:
                         position["cost_recovered"] = False
                         position["yield_usd"] = float(yield_usd)
+                        entry_cost_dec = Decimal(str(entry_cost))
```
OjusWiZard, line 1483: The cast is correct. Minor: `_calculate_corrected_yield` still returns `Decimal(0)` on a token-decimals fetch failure, which the caller treats as a real zero-yield measurement and persists. Worth a separate follow-up to return `None` there so it preserves the last-known value like the other failure path added in this skill.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -402,6 +402,9 @@ class Chain(Enum):
         "eusd": "electronic-usd",
         "dola": "dola-usd",
         "musd": "mezo-usd",
+        "msusd": "metronome-synth-usd",
```
OjusWiZard, line 405: Advisory (nice-to-have): the three new Base CoinGecko IDs aren't exercised by any test. A chain-scoped `get_coin_id_from_symbol` round-trip test would guard against a stale/misspelled ID silently pricing a token at zero — and against chain mix-ups (e.g. `bold` -> `liquity-bold-2` on Base vs `liquity-bold` elsewhere). PR notes each ID was verified against the live API, so low risk.

OjusWiZard (approved):
Approving — the three core fixes are correct and well-tested.

- **Payload reorder**: moving `event` last is the right fix for the `dict(zip(selection_key, payload.values))` truncation; verified against the round's `selection_key`. The new `test_end_block_emits_mech_request_needed_via_real_payload_projection` reproduces the exact production projection path, so future field-order drift fails loudly.
- **EOA gate scoped to old regime**: sound; fail-opens on `signal is None`.
- **`Decimal(str(entry_cost))` cast**: correctly fixes the float/Decimal `TypeError` on `/` and `-`; `str()` avoids float-repr noise. (Note: the `>=` comparison was never affected — `Decimal >= float` is valid in Python 3.)

All inline comments below are advisory/follow-up, none blocking.

Separate follow-up (out of this diff): `_calculate_corrected_yield` returns `Decimal(0)` when token-decimals fetch fails, which the caller treats as a real measurement and overwrites `yield_usd` — the same silent-zero class fixed elsewhere here. Returning `None` would route into the existing last-known-value preservation.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
DIvyaNautiyal07: Good call. Will address in a follow-up PR — non-gating logger.warning on the new-regime path when _real_tx_cost_vs_balance reports the EOA below recent cost, so we keep the operator visibility without re-introducing the gate.

packages/valory/skills/liquidity_trader_abci/behaviours/fetch_strategies.py
DIvyaNautiyal07: Agreed, this is the same silent-zero class. Will address in a follow-up PR — return None from _calculate_corrected_yield on the decimals-fetch failure path so the caller takes the existing preserve-last-known route, with a regression test that forces the failure.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Will add in a follow-up PR — a parametrized chain-scoped get_coin_id_from_symbol round-trip test for the three new Base IDs (and bold, which would differ on other chains).

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
DIvyaNautiyal07: Skipping. The signal would help, but it's on the old vanity-tx regime which is being deprecated in favor of the new mech-request path. Not worth the diff at this point.

bennyjo (commented):
Two non-blocking notes — the field-order fix and tests look correct. One question on the gate removal, one note on the field-order fragility.

packages/valory/skills/liquidity_trader_abci/behaviours/check_staking_kpi_met.py
```diff
@@ -198,12 +178,30 @@ def async_act(self) -> Generator:  # type: ignore[override]
                             )
                             mech_requests_json = self._build_mech_request_metadata()
                         else:
-                            # Old regime: keep the existing vanity Safe tx.
-                            self.context.logger.info("Preparing vanity tx..")
-                            vanity_tx_hex = yield from self._prepare_vanity_tx(
+                            # Old regime: keep the existing vanity Safe tx, gated
+                            # by the EOA-funded check so vanity activity does not
+                            # hide a funding alert behind a green staking KPI.
+                            signal = self._real_tx_cost_vs_balance(
```
bennyjo, line 184: Question (non-blocking): the EOA-vs-recent-real-tx-cost gate now applies only to the old-regime vanity tx, and the new-regime mech path fires unconditionally. The rationale (mech-marketplace `request()` settles via the Safe and is paid in USDC/x402, so EOA native balance is the wrong gate) makes sense — but does *submitting* that Safe tx still require any native gas from the agent EOA? If it does, an under-funded EOA on the new regime will now attempt and fail the activity tick rather than emit the funding warning. Happy to leave as-is if a relayer/x402 covers submission; flagging because the failure mode would be a silent stuck tick instead of the named funding alert.

packages/valory/skills/liquidity_trader_abci/payloads.py
```diff
@@ -59,6 +58,13 @@ class CheckStakingKPIMetPayload(MultisigTxPayload):
     is_activity_target_met: Optional[bool] = None
     activity_target: Optional[int] = None
     activity_completed: Optional[int] = None
+    # Field order matters: ``event`` must remain the last declared field so
+    # ``zip(selection_key, payload.values)`` (see ``CollectSameUntilThresholdRound.end_block``)
+    # aligns each selection_key entry with its same-named payload position. The
+    # round's selection_key intentionally omits ``event``; with ``event`` last,
+    # ``zip`` truncates the trailing value cleanly. Reordering any earlier
+    # field would silently shift db keys and break the MECH_REQUEST_NEEDED branch.
+    event: Optional[str] = None
```
bennyjo, line 67: Fix is correct. Note (non-blocking): correctness here depends on declaration order matching `selection_key` order positionally via `zip(selection_key, payload.values)` truncation — a genuinely fragile contract. The comment + the new `test_payload_values_align_with_selection_key` / `test_end_block_emits_mech_request_needed_via_real_payload_projection` regression tests adequately guard it, so this is fine to merge; just flagging that any future field added before `event` (or any new field not in `selection_key`) reintroduces the same silent db-key shift.

bennyjo (approved):
LGTM. Core payload field-order fix is correct and properly regression-tested (real payload→selection_key projection, fails under field-order mutation); Decimal cast and CoinGecko side-fixes are low-risk. No blocking issues. Two non-blocking notes left inline. Worth confirming the rounds_info mech-rounds and Basius mech-marketplace defaults landed in main, since they're not in this diff.

PR #350, opened 2026-06-23
fix(basius): point mechs_subgraph at marketplace-base on the autonolas proxy

DIvyaNautiyal07:
## Summary

Basius runs on Base with the Base mech marketplace `0xf24eE42edA0fc9b33B7D41B06Ee8ccD2Ef7C5020`, but `optimus_abci/skill.yaml:481` hardcodes the `mechs_subgraph.url` at `https://api.subgraph.autonolas.tech/api/proxy/marketplace-optimism`. The Optimism proxy doesn't index Base marketplace mechs, so every `MechInformationBehaviour.fetch_mechs_info()` call returned `[]`. `mech_info.py:156` then set `last_failure_reason = "valid_mech_list_empty"`, the round emitted `NONE`, and the composed Optimus FSM routed `(MechInformationRound, NONE) → MechVersionDetectionRound` forever — no mech request was ever submitted, so `mapRequestCounts` stayed at 0 and the staking KPI never got ticked.

Adds a `mechs_subgraph` model override in `basius/service.yaml` and `basius/aea-config.yaml` pointing at the Base proxy, made env-overridable via `MECHS_SUBGRAPH_URL` for operator flexibility. Optimus is untouched.

```yaml
mechs_subgraph:
  args:
    url: ${MECHS_SUBGRAPH_URL:str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-base}
```

## Verified before the change

- `POST {marketplace-base}/_meta` → `HTTP 200`, current Base block `47714957`.
- The priority mech `0xe535D7AcDEeD905dddcb5443f41980436833cA2B` is indexed at id `112` with `10090` total deliveries and a non-null metadata CID — so the exact query `mech_info.py` issues (`address_in: [...]`, `service_: {totalDeliveries_gt: 0}`, recent `blockTimestamp_gt`) will return a non-empty result on the first call.
- Pattern matches `trader`'s approach: per-chain marketplace subgraph URLs, env-overridable. Optimus already uses `marketplace-optimism`; Basius now uses `marketplace-base`.

## Test plan
- [ ] Boot a Basius Base agent past `staking_threshold_period` periods.
- [ ] Observe `Event.MECH_REQUEST_NEEDED → MechVersionDetectionRound → MechInformationRound` resolves with `Updated mechs' information: ...` (non-empty) instead of the empty-list infinite loop.
- [ ] After mech request settles, observe `mapRequestCounts` increment on-chain at the activity checker.

## Out of scope
- Making the URL env-overridable in `optimus_abci/skill.yaml` itself (Optimus continues to default to `marketplace-optimism` via the unchanged skill default).
- Any fallback path when the subgraph is empty — kept as-is for now; we can revisit if the Base subgraph turns out to be flaky.

### DIFF
--- a/packages/valory/services/basius/service.yaml
+++ b/packages/valory/services/basius/service.yaml
@@ -6,7 +6,7 @@ aea_version: '>=2.0.0, <3.0.0'
 license: Apache-2.0
 fingerprint: {}
 fingerprint_ignore_patterns: []
-agent: valory/basius:0.1.0:bafybeigki6o6zw7nm53xm7vfxd2npy5qkxukfnmjfgrexkddsra3rbgnxi
+agent: valory/basius:0.1.0:bafybeierqtcvug7p4g4w6hdzl3ja2avvsdsbnj2zskyzztmbzwuvaanwbq
 number_of_agents: 1
 deployment:
   agent:
@@ -157,6 +157,9 @@ models:
       coingecko_server_base_url: ${COINGECKO_SERVER_BASE_URL:str:https://api.coingecko.com}
       coingecko_x402_server_base_url: ${COINGECKO_X402_SERVER_BASE_URL:str:https://x402.olas.autonolas.tech/v1/coingecko/{chain}}
       network_selector: ${GENAI_NETWORK_SELECTOR:str:base}
+  mechs_subgraph:
+    args:
+      url: ${MECHS_SUBGRAPH_URL:str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-base}
 ---
 public_id: valory/ledger:0.19.0
 type: connection
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -26,9 +26,9 @@
         "skill/valory/liquidity_trader_abci/0.1.0": "bafybeigoe5msyzv2ompaxvhyijqf54666nltucnblqzuezysfesnzsjip4",
         "skill/valory/optimus_abci/0.1.0": "bafybeihoneiqnskjd7eskcngaauypulhoqh5apvklrkbz6irpnt2556qnm",
         "agent/valory/optimus/0.1.0": "bafybeifcj4v3lke5l7ovizcqglcplfdngiszgil2cplq2l4sq652b47jaq",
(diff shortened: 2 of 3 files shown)


OjusWiZard (approved):

iasonrovis (approved):

packages/valory/services/basius/service.yaml
```diff
@@ -157,6 +157,9 @@ models:
       coingecko_server_base_url: ${COINGECKO_SERVER_BASE_URL:str:https://api.coingecko.com}
       coingecko_x402_server_base_url: ${COINGECKO_X402_SERVER_BASE_URL:str:https://x402.olas.autonolas.tech/v1/coingecko/{chain}}
       network_selector: ${GENAI_NETWORK_SELECTOR:str:base}
+  mechs_subgraph:
+    args:
+      url: ${MECHS_SUBGRAPH_URL:str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-base}
```
bennyjo, line 162: This correctly removes the *chain-mismatch* cause of the `valid_mech_list_empty -> NONE -> MechVersionDetectionRound` loop. The "Out of scope" note already flags it, but raising as a question since it's the failure class most likely to resurface this exact ticket: if `marketplace-base` ever returns `[]` (proxy flakiness, the priority mech `0xe535...cA2B` rotating out, or the `blockTimestamp_gt`/`totalDeliveries_gt` filter excluding everything), the agent silently re-enters the same infinite loop with no on-chain progress and no distinct log signal.

Is deferring the empty-result fallback deliberate for this PR? If so, fine to leave — flagging because the only signal operators would have is the absence of `mapRequestCounts` increments, which is easy to miss. A bounded-retry-then-warn path would make the degradation observable rather than silent.

bennyjo (commented):
One residual-risk question on the empty-subgraph path — happy to leave as-is if deliberate.

bennyjo (approved):
LGTM. Correct, minimal config fix — repoints basius mechs_subgraph from marketplace-optimism to marketplace-base, overriding only `url` so it merges cleanly with the inherited args, placed under the right optimus_abci skill block in both aea-config.yaml and service.yaml. packages.json bumps only the basius agent+service hashes; optimus untouched. Approving.

PR #355, opened 2026-06-29
feat(offchain): wire mech-interact off-chain branch through Optimus FSM

DIvyaNautiyal07:
## Summary

Wires the off-chain mech-interact path (valory-xyz/mech-interact#95) through the Optimus FSM so the agent can run on the new HTTP-based mech, including the full auto-deposit on 402 retry without manual BalanceTracker funding.

## What changed

- **3 new contract packages** vendored from mech-interact: `mech_marketplace`, `balance_tracker_fixed_price_native`, `balance_tracker_fixed_price_token`.
- **liquidity_trader_abci**:
  - New `Event.OFFCHAIN_MECH_DEPOSIT_SETTLED`.
  - New `FinishedWithOffchainMechDepositSettledRound` degenerate state, wired into the transition function and final_states.
  - `PostTxSettlementRound.end_block` dispatches the new `OFFCHAIN_DEPOSIT_TX_SUBMITTER` sentinel (from mech_interact_abci) to that event; the existing `MechRequestRound.auto_round_id()` -> `MECH_REQUEST_TX_EXECUTED` entry is unchanged.
  - One new test pinning the sentinel dispatch.
- **optimus_abci composition**: 4 new transitions for the off-chain final states.
- **optimus_abci skill.yaml**: off-chain knobs under `mech_marketplace_config` (defaults keep `use_offchain=false` so on-chain Optimus deployments are unaffected).
- `autonomy packages lock --check` passes; FSM specs regenerated.

## Off-chain flow

Empty BalanceTracker on first request -> mech returns 402 -> executor builds approve+depositFor multisend -> sentinel `tx_submitter` routes the settled tx back into `MechRequestRound` -> `_retry_pending` re-POSTs -> mech 200 -> response polled over HTTP. No human in the loop.

## Test plan

- [x] 3650 tests pass (`pytest packages/valory/skills/{liquidity_trader_abci,optimus_abci,mech_interact_abci}/tests/`).
- [x] `autonomy packages lock --check` passes.
- [x] `autonomy analyse fsm-specs` passes.
- [ ] Local e2e with a mech running in HTTP mode (Gnosis service 2692 or Optimism mech 195): first-request 402 -> deposit -> retry -> response, all hands-off.

This targets `feat/offchain-epic` rather than `main` because it depends on valory-xyz/mech-interact#95 and the new IPFS hash for `mech_interact_abci` in `packages.json` isn't on the registry until #95 lands.


bennyjo (approved):
Approving. This is the Optimus-side wiring for the off-chain mech path; the bulk of the behaviour lives in mech-interact #95 (now merged into this base). The FSM wiring is clean and the three integration invariants it depends on are satisfied in #95:

1. **402 destination is validated before signing** — #95's `offchain_request.py` runs `_validate_402_destination` (payTo vs `mapPaymentTypeBalanceTrackers(paymentType)`, and asset vs tracker `token()` on the token path) before `_build_deposit_tx`. Mismatch → `BAD_RESPONSE` failover, no deposit signed; and it fails closed if the on-chain read fails.
2. **Deposit tx stamps the exact sentinel** — `tx_submitter = OFFCHAIN_DEPOSIT_TX_SUBMITTER` (`"mech_request_round_offchain_deposit"`) only on the deposit branch, matching the dispatch added here in `PostTxSettlementRound`. The success branch deliberately omits it, so a first-try 200 won't misroute into the retry leg.
3. **`RESPONSE_ROUND_TIMEOUT` migration is clean** — #95 removes the `(MechResponseRound, ROUND_TIMEOUT)` transition and replaces it with `RESPONSE_ROUND_TIMEOUT`, so no orphaned 30s timeout hangs the poll loop.

Framework FSM audit ran clean: new events all have transitions, the four inner finals are mapped in the composition (asserted by an exact key-set test), and db pre/post conditions match the sibling mech legs. Security: no fund movement or signing in this diff — the vendored wrappers are read-only `.call()`s / calldata encoders; `use_offchain` defaults `false` so on-chain deployments are unaffected.

Note it targets `feat/offchain-epic`, not `main`, per the dependency on #95's registry hashes.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
```diff
@@ -0,0 +1,89 @@
...
+        contract_address = ledger_api.api.to_checksum_address(contract_address)
+        account = ledger_api.api.to_checksum_address(account)
+        contract_instance = cls.get_instance(ledger_api, contract_address)
+        balance = contract_instance.functions.mapRequesterBalances(account).call()
+        return dict(balance=balance)
+
+    @classmethod
+    def build_deposit_for_data(
+        cls,
+        ledger_api: LedgerApi,
```
OjusWiZard, line 71: **[Important] Type error.** `build_deposit_for_data` annotates `ledger_api: LedgerApi`, but line 87 calls `ledger_api.api.to_checksum_address(...)` — `.api` lives on `EthereumApi`, not the `LedgerApi` ABC. The sibling `get_balance` in this same file types it as `EthereumApi`, and `mech_marketplace/contract.py` casts explicitly, so this is the odd one out and will trip mypy strict.

Fix: `ledger_api: EthereumApi` (matches the sibling method; simpler than a `cast`).

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
```diff
@@ -0,0 +1,89 @@
...
+
+        :param ledger_api: the ledger API object.
+        :param contract_address: the BalanceTracker contract address.
+        :param account: the requester address being credited.
+        :return: ``{"data": bytes}`` calldata for the multisend batch.
+        """
+        contract_instance = cls.get_instance(ledger_api, contract_address)
+        account = ledger_api.api.to_checksum_address(account)
+        data = contract_instance.encode_abi("depositFor", args=(account,))
+        return {"data": bytes.fromhex(data[2:])}
```
OjusWiZard, line 89: **[Advisory] `bytes.fromhex(data[2:])` assumes the `0x` prefix.** `encode_abi` returns a `0x`-prefixed `HexStr` today, so this is correct and matches the repo convention (`velodrome_cl_gauge`, `sturdy_yearn_v3_vault`). Optional hardening: guard the prefix and raise, so a future web3 change surfaces as a loud error rather than silently mis-sliced calldata that reverts on-chain.

packages/valory/contracts/balance_tracker_fixed_price_token/contract.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        contract_address = ledger_api.api.to_checksum_address(contract_address)
+        contract_instance = cls.get_instance(ledger_api, contract_address)
+        token = contract_instance.functions.token().call()
+        return dict(token=str(token))
+
+    @classmethod
+    def build_deposit_for_data(
+        cls,
+        ledger_api: LedgerApi,
```
OjusWiZard, line 96: **[Important] Type error (same as the native tracker).** `build_deposit_for_data` annotates `ledger_api: LedgerApi` but line 116 accesses `.api` (an `EthereumApi` attribute). The other methods here type it as `EthereumApi`. Fix: change the annotation to `EthereumApi`.

packages/valory/contracts/balance_tracker_fixed_price_token/contract.py
```diff
@@ -0,0 +1,118 @@
...
+        :param ledger_api: the ledger API object.
+        :param contract_address: the BalanceTracker contract address.
+        :param account: the requester address being credited.
+        :param amount: the deposit amount in the token's smallest unit.
+        :return: ``{"data": bytes}`` calldata for the multisend batch.
+        """
+        contract_instance = cls.get_instance(ledger_api, contract_address)
+        account = ledger_api.api.to_checksum_address(account)
+        data = contract_instance.encode_abi("depositFor", args=(account, amount))
+        return {"data": bytes.fromhex(data[2:])}
```
OjusWiZard, line 118: **[Advisory]** Same `bytes.fromhex(data[2:])` prefix assumption as the native tracker (consistent with repo convention; optional prefix guard).

packages/valory/contracts/mech_marketplace/contract.py
```diff
@@ -0,0 +1,131 @@
...
+from aea.common import JSONLike
+from aea.configurations.base import PublicId
+from aea.contracts.base import Contract
+from aea.crypto.base import LedgerApi
+from aea_ledger_ethereum import EthereumApi
+
+PUBLIC_ID = PublicId.from_str("valory/mech_marketplace:0.1.0")
+
+
+class MechMarketplace(Contract):
```
OjusWiZard, line 33: **[Important — verify, do not assume]** None of the three new contract packages ship a `tests/` dir, and `.coveragerc` enforces `fail_under = 100` over `packages/valory`. *However*, the existing vendored `mech_mm` and `mech_marketplace_legacy` also have no tests and no in-repo importers, yet CI passes — so un-imported vendored contracts appear exempt, and these wrappers (imported only by external `#95`) match that precedent.

Action: run `tox -e coverage` to settle it. If it fails, add a `test_contract.py` per package (mock `get_instance`, assert `encode_abi` args + return shape), following `velodrome_cl_gauge/tests/test_contract.py`.

packages/valory/skills/optimus_abci/tests/test_composition.py
```diff
@@ -73,6 +74,9 @@ def test_abci_app_transition_mapping_keys() -> None:
         MechFinalStates.FinishedMechResponseRound,
         MechFinalStates.FinishedMechResponseTimeoutRound,
         MechFinalStates.FinishedMechRequestSkipRound,
+        MechFinalStates.FinishedOffchainMechRequestRound,
+        MechFinalStates.FinishedOffchainMechDepositNeededRound,
+        MechFinalStates.FailedOffchainMechRequestRound,
```
OjusWiZard, line 79: **[Important] New composition edges aren't individually pinned.** This is a flat value-set assertion, and both `MechRequestRound` and `MechResponseRound` were already in the expected set — so swapping the off-chain targets (e.g. routing `FinishedWithOffchainMechDepositSettledRound` to `MechResponseRound` instead of `MechRequestRound`, which would break `_retry_pending`) would still pass both the key-set and value-set tests.

Add explicit edge assertions to `test_abci_app_transition_mapping_specific_transitions`:
```python
assert m[LiquidityTraderAbci.FinishedWithOffchainMechDepositSettledRound] == MechRequestStates.MechRequestRound
assert m[MechFinalStates.FinishedOffchainMechRequestRound] == MechResponseStates.MechResponseRound
assert m[MechFinalStates.FinishedOffchainMechDepositNeededRound] == TxSettlementAbci.RandomnessTransactionSubmissionRound
assert m[MechFinalStates.FailedOffchainMechRequestRound] == LiquidityTraderAbci.CheckStakingKPIMetRound
```

**[Advisory]** Relatedly, nothing pins `transition_function[PostTxSettlementRound][OFFCHAIN_MECH_DEPOSIT_SETTLED] == FinishedWithOffchainMechDepositSettledRound` in `test_rounds.py` — worth adding alongside, following `TestCoreCycleTransitions`.

packages/valory/skills/liquidity_trader_abci/states/post_tx_settlement.py
```diff
@@ -81,6 +84,9 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
                 # MechRequestRound, so it must map back to the staking loop or the
                 # multiplexer returns UNRECOGNIZED and the loop dies.
                 MechRequestRound.auto_round_id(): Event.MECH_REQUEST_TX_EXECUTED,
+                # Settled off-chain auto-deposit: re-enter MechRequestRound
+                # for ``_retry_pending``, not forward to MechResponseRound.
+                OFFCHAIN_DEPOSIT_TX_SUBMITTER: Event.OFFCHAIN_MECH_DEPOSIT_SETTLED,
```
OjusWiZard, line 89: **[Advisory] Silent fall-through.** This new sentinel now routes a money-moving deposit settlement through `submitter_to_event`. If a quorum commits a `tx_submitter` that doesn't match (e.g. a node on a different mech-interact version), line 101 falls to `Event.UNRECOGNIZED` → `FailedMultiplexerRound` → reset, with **no log of the offending submitter** — the deposited funds are gone and the retry never fires, diagnosable only from raw Tendermint state. Pre-existing behavior, but higher stakes now. Cheap fix: log `synced_data.tx_submitter` at WARNING before returning `UNRECOGNIZED`.

packages/packages.json
```diff
@@ -50,8 +50,11 @@
         "contract/valory/staking_activity_checker/0.1.0": "bafybeif7h7ewjdojrqjg6fps2t4navbftvnjn45zcpiza554xhefdhprwa",
         "contract/valory/staking_token/0.1.0": "bafybeiev2of6e273dcmm4vqh5lejurn26wbhw4jf4wdabghrfi6abv6sqa",
         "contract/valory/mech_marketplace_legacy/0.1.0": "bafybeifkolgdeaoiveuppykvxkvja7c7pphn6jyivnk6x3bjl72rqpsogm",
+        "contract/valory/mech_marketplace/0.1.0": "bafybeieadbmhzndxe6jcfwujcpjrljpc5kfwffkd3zzpg75uxksufn4l3i",
         "contract/valory/mech/0.1.0": "bafybeid2v3rotkylgln5w2udm3nnmsrnw4g4nlkbwhqfwbte4kwlxup5l4",
         "contract/valory/mech_mm/0.1.0": "bafybeiaez7w5ssgcmt5fqbue4oywhxmcojrnxqimzksuojsksxwaqvwo5u",
+        "contract/valory/balance_tracker_fixed_price_native/0.1.0": "bafybeifobvptvc7mkrztfij664dqqegl6qi32773zewvyuftes32faemw4",
```
OjusWiZard, line 56: **[Important — verify]** These three new contracts are referenced *only* by `packages.json` and their own files — no agent `aea-config.yaml`, `service.yaml`, or local skill pulls them in. They're wired into the dependency closure solely by the external `mech_interact_abci#95`, so `autonomy packages lock --check` can only have passed against a locally-synced `#95`. Please re-confirm the lock and the full closure once `#95` lands and its IPFS hash is published.

packages/valory/skills/optimus_abci/skill.yaml
```diff
@@ -345,6 +345,13 @@ models:
         priority_mech_service_id: 194
         response_timeout: 300
         use_dynamic_mech_selection: false
+        use_offchain: false
```
OjusWiZard, line 348: **[Advisory] Config parity.** These off-chain knobs are added to the skill.yaml defaults, but `services/optimus/service.yaml`'s `MECH_MARKETPLACE_CONFIG` env-default dict does **not** include them. Since a `dict:` env override replaces the whole value, a default deployment's `mech_marketplace_config` won't carry `use_offchain` et al. — fine only if `MechParams` (in `#95`) reads them with `.get()` defaults; a `KeyError` at startup otherwise. Consider adding at least `"use_offchain": false` to the service.yaml env-default for parity.

OjusWiZard (commented):
Automated multi-perspective review (code-reviewer + pr-test-analyzer + silent-failure-hunter, each finding re-verified against source).

**Verdict: approve-with-changes.** FSM wiring is correct and complete (deposit→settle→retry loop closes; no dangling/unreachable states; all ABI functions referenced by the wrappers exist). One concrete type bug in new code (I1) plus test/verification gaps. Nothing critical.

Severity tags inline: **[Important]** / **[Advisory]**. Items I2/I4/A1 cannot be fully validated locally because `mech_interact_abci#95` isn't on the registry yet — flagged for confirmation, not asserted as broken.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Acknowledged. This is a third_party contract vendored from mech-interact (packages.json lists it under third_party, not dev), so the type fix has to land there. Tracked separately on the mech-interact side; will re-bump optimus to the patched IPFS hash when it ships.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Acknowledged, same as above — bytes.fromhex(data[2:]) lives in the mech-interact-owned contract; will guard alongside the type fix in that repo.

packages/valory/contracts/balance_tracker_fixed_price_token/contract.py
DIvyaNautiyal07: Same as above (third_party contract). Will fix in mech-interact and re-bump.

packages/valory/contracts/balance_tracker_fixed_price_token/contract.py
DIvyaNautiyal07: Same as above (third_party contract); the prefix guard belongs in mech-interact.

packages/valory/contracts/mech_marketplace/contract.py
DIvyaNautiyal07: Tested locally: tox -e coverage passes — the un-imported vendored contracts are exempt the same way mech_mm and mech_marketplace_legacy already were, as you suspected. No action on optimus. The contract tests proper still belong in mech-interact and I'll raise them there.

packages/valory/skills/optimus_abci/tests/test_composition.py
DIvyaNautiyal07: Fixed in f127ab45. Pinned each new off-chain edge individually in test_abci_app_transition_mapping_specific_transitions — the four edges you listed (FinishedWithOffchainMechDepositSettledRound → MechRequestRound, FinishedOffchainMechRequestRound → MechResponseRound, FinishedOffchainMechDepositNeededRound → RandomnessTransactionSubmissionRound, FailedOffchainMechRequestRound → CheckStakingKPIMetRound). Advisory follow-up (PostTxSettlementRound[OFFCHAIN_MECH_DEPOSIT_SETTLED] in test_rounds.py) — agreed, will add in a follow-up PR alongside the rest of TestCoreCycleTransitions to keep this PR scoped to wiring.

packages/valory/skills/liquidity_trader_abci/states/post_tx_settlement.py
DIvyaNautiyal07: Fixed in f127ab45. PostTxSettlementRound.end_block now logs tx_submitter at WARNING when it falls through to UNRECOGNIZED, so a silently-dropped off-chain deposit is visible in agent logs instead of only in raw Tendermint state. Stubbed the test round's context.logger so the existing UNRECOGNIZED test still passes.

packages/packages.json
DIvyaNautiyal07: Re-verified. The closure now resolves against the published v0.32.4-rc1 release (bde7e2db bumps pyproject.toml tag and packages.json hash to bafybeihhwzpa6x5ypth426tzd2zjsnys6jmjaab4zvk4edisetzktft754), autonomy packages lock --check passes against the published IPFS hash (no longer against a local sync).

packages/valory/skills/optimus_abci/skill.yaml
DIvyaNautiyal07: Fixed in f127ab45. Added the seven off-chain knobs (use_offchain, offchain_url, offchain_deposit_target_calls, auto_deposit_cap_per_cycle, offchain_poll_interval_seconds, offchain_poll_timeout_seconds, offchain_failover_max_retries) to both services/basius/service.yaml and services/optimus/service.yaml MECH_MARKETPLACE_CONFIG env-defaults with use_offchain: false. A MECH_MARKETPLACE_CONFIG-via-env override now carries the full dict so the agent does not depend on .get() defaults in MechParams for the off-chain path.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Confirmed for follow-up PR on mech-interact (one small PR covers all five). Not blocking on this PR — type annotations are cosmetic (runtime is unaffected because the framework always passes EthereumApi and Python doesn't enforce hints; current mypy config doesn't flag it either), and the contracts ship with no tests by precedent with mech_mm / mech_marketplace_legacy.

packages/valory/contracts/balance_tracker_fixed_price_native/contract.py
DIvyaNautiyal07: Confirmed for follow-up PR on mech-interact (one small PR covers all five). Not blocking on this PR — type annotations are cosmetic (runtime is unaffected because the framework always passes EthereumApi and Python doesn't enforce hints; current mypy config doesn't flag it either), and the contracts ship with no tests by precedent with mech_mm / mech_marketplace_legacy.

packages/valory/contracts/balance_tracker_fixed_price_token/contract.py
DIvyaNautiyal07: Confirmed for follow-up PR on mech-interact (one small PR covers all five). Not blocking on this PR — type annotations are cosmetic (runtime is unaffected because the framework always passes EthereumApi and Python doesn't enforce hints; current mypy config doesn't flag it either), and the contracts ship with no tests by precedent with mech_mm / mech_marketplace_legacy.

packages/valory/contracts/balance_tracker_fixed_price_token/contract.py
DIvyaNautiyal07: Confirmed for follow-up PR on mech-interact (one small PR covers all five). Not blocking on this PR — type annotations are cosmetic (runtime is unaffected because the framework always passes EthereumApi and Python doesn't enforce hints; current mypy config doesn't flag it either), and the contracts ship with no tests by precedent with mech_mm / mech_marketplace_legacy.

packages/valory/contracts/mech_marketplace/contract.py
DIvyaNautiyal07: Confirmed for follow-up PR on mech-interact (one small PR covers all five). Not blocking on this PR — type annotations are cosmetic (runtime is unaffected because the framework always passes EthereumApi and Python doesn't enforce hints; current mypy config doesn't flag it either), and the contracts ship with no tests by precedent with mech_mm / mech_marketplace_legacy.

OjusWiZard (approved):
Approving. The off-chain mech deposit-retry wiring is correct and complete — FSM edges verified against the regenerated spec (deposit→settle→retry loop closes, no dangling/unreachable states), and all CI green (test across 5 OS × Py 3.10–3.14, linter, lock_check, coverage).

All review findings resolved:
- Fixed: explicit composition edge asserts (I3), off-chain config parity in optimus + basius service.yaml (A1), WARNING log on the UNRECOGNIZED branch with covering test (A2).
- Verified green by author: tox -e coverage passes (I2), dependency closure resolves against published mech-interact v0.32.4-rc1 (I4).
- Deferred to a mech-interact follow-up PR (vendored third-party contracts, non-blocking — CI confirms linter passes): EthereumApi annotation on build_deposit_for_data, 0x-prefix guard on bytes.fromhex.

Nit (optional, non-blocking): the A2 warning-logging behavior isn't asserted (logger stubbed no-op; test checks only the returned event) — pin with a Mock + assert_called_once if desired.

Ready once the base epic and the mech-interact dependency land.

PR #357, opened 2026-07-06
fix: mech fee reserve, safe native funding, velodrome unstake bytes

DIvyaNautiyal07:
## Context

Basius (Optimus on Base) Pearl agents were not earning staking rewards. The chain of causes:

1. Mech requests on Base go through the marketplace mech (mech 112), which charges 100 wei of native ETH per request, paid from the agent safe.
2. `fund_requirements` had the base safe's native entry at `topup: 0, threshold: 0`, so the Pearl middleware never auto-funded the safe with native ETH.
3. The trader treats the entire safe balance as investable. Any native ETH that did land on the safe was either invested into pools or swept to USDC by the non-whitelisted-asset cleanup, so there was never anything left to pay mech fees with. No mech requests means the staking KPI is missed and no rewards accrue.

Separately, users trying to withdraw from Velodrome v2 staked positions were stuck: withdrawals looped forever in the FSM (one user withdrawal pending since 2026-07-01).

## Changes

**1. Funding config (basius + optimus service/agent yamls)**
- Base safe native ETH requirement is now `topup: 100000000000000` (0.0001 ETH), `threshold: 350` wei, so the middleware funds the safe and tops it up when it runs low. At 100 wei per request the topup covers mech fees for far longer than the refill cadence.
- The USDC requirement on the base safe is removed (mech fees on Base are paid in native ETH, not USDC).
- The same `fund_requirements` dict is now also passed to `liquidity_trader_abci` (new skill param), for both basius and optimus, so the trader knows what the funding floor is.

**2. Mech fee reserve in the trader (`liquidity_trader_abci`)**
- New helpers `_get_mech_fee_reserve` / `_apply_mech_fee_reserve` in `behaviours/base.py`: the safe `topup` for a token in `fund_requirements` is treated as a reserve and deducted from the balance the trader considers investable (`max(0, balance - reserve)`).
- Applied at the selection layer (`_get_investable_balance` in `evaluate_strategy.py`) and at every execution-layer amount derivation in `decision_making.py`: velodrome investment amounts, token-balance based amounts, deposits, and swap route amounts.
- Withdrawals bypass the reserve: when `investing_paused` is set (the withdrawal flow), route amounts are not reduced, so users can always withdraw their full funds.

**3. Native ETH excluded from the swap-to-USDC cleanup**
- `check_and_prepare_non_whitelisted_swaps` now skips the zero address. Previously it would swap the safe's native ETH (the mech fee gas) into USDC and invest it.

**4. Velodrome v2 gauge unstake crash fix**
- Since 72cda47e the velodrome contract wrappers return tx data as `bytes` (`velodrome_gauge/contract.py` `withdraw` returns `bytes.fromhex(data[2:])`), but `unstake_lp_tokens` in `pools/velodrome.py` still re-applied `bytes.fromhex` on the result. Every non-CL gauge unstake raised `TypeError`, which was swallowed by the `except` in `get_unstake_lp_tokens_tx_hash`, returning no transaction and leaving the FSM looping on the withdrawal forever.
- Fix is a one-line pass-through of the wrapper's bytes. A sweep of all other `bytes.fromhex` call sites in the skill confirmed they consume MultiSend `get_tx_data` outputs or event-log fields, which are still hex strings, so they are correct and untouched.
- Once deployed, stuck withdrawals resume automatically from the exit-pools step, no manual intervention needed.

## Tests

- Unit tests for the reserve helpers (case-insensitive address match, unknown chain/token, invalid values, floor at zero).
- One test per execution-layer reserve site asserting the exact deducted amounts, plus the on/off pair for the withdrawal bypass.
- Native ETH skip test for the cleanup path.
- Unstake fix: the happy-path unit test now stubs the withdraw call with bytes, matching what the real wrapper returns (this test fails on the old code with the production `TypeError`), plus an integration-style test that runs the real `VelodromePoolBehaviour` through `get_unstake_lp_tokens_tx_hash` with only the contract boundary stubbed, asserting the gauge tx bytes reach the Safe call unchanged (also fails on the old code).

## Test plan

- [x] Full `liquidity_trader_abci` test suite passes locally
- [x] All CI checks run locally: black/isort/flake8/darglint, mypy/pylint, check-hash, check-packages, check-abci-docstrings, check-abciapp-specs, check-handlers, check-dependencies, liccheck, check-third-party-hashes, check-doc-hashes, gitleaks
- [x] `autonomy packages lock` + `lock --check`, `autonomy push-all` done
- [ ] Deploy new basius service hash via Pearl and confirm mech requests are funded and the stuck withdrawal completes

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -23,12 +23,12 @@
         "contract/valory/velodrome_cl_gauge/0.1.0": "bafybeihyci7aqcrefyy4uak245jq2ud2536c7lqjgmxuzwilfwtw7p2a4y",
         "contract/valory/balancer_queries/0.1.0": "bafybeihbexpjj6ho3ij7rxk4tzz3sowpa5ztawlf63y5fdmmom7xifpan4",
         "contract/valory/mech_activity/0.1.0": "bafybeienezl3yozrionpoyh75r54e24tjtnstxv674ru25lnosps3a7cr4",
-        "skill/valory/liquidity_trader_abci/0.1.0": "bafybeic5jfo2kpqsprixq25qi7cg25w6v5za3f3pzzcrektauipo3fjr6y",
-        "skill/valory/optimus_abci/0.1.0": "bafybeieibacgr3aoh5dqbh6srymeiczp7gd2fck7w43mc6p2txqplu3fji",
-        "agent/valory/optimus/0.1.0": "bafybeic4nmsdjizfqn3wih4agxfak6novigjqe6huvo3fdv3xnydy2rydq",
-        "agent/valory/basius/0.1.0": "bafybeifa53dshw255knmjwq6kih3mm6tyta4dvlulc4vogt6fgg6xvyene",
-        "service/valory/optimus/0.1.0": "bafybeiaclpkthh3qbskzseujvr3gdtmhscsixpddzbbikb6s5fvjls3d4i",
-        "service/valory/basius/0.1.0": "bafybeigkl2i7bmoeaxdp3axw6kxqbfjpb3jljfsqk7wfp2g565dafazhh4"
+        "skill/valory/liquidity_trader_abci/0.1.0": "bafybeifnu7x5vso3lyi5wr6itfb5q33igoznm6gozqmd3vvuadwckoevna",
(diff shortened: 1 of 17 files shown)


DIvyaNautiyal07:
Closing this one, it was branched from and targeted at feat/offchain-epic by mistake. Replaced by #358 which carries the same changes on a fresh branch off main.

PR #358, opened 2026-07-06
fix: reserve mech fees from investable balance + velodrome unstake tx bytes pass-through

DIvyaNautiyal07:
## Context

Basius (Optimus on Base) Pearl agents were not earning staking rewards. The chain of causes:

1. Mech requests on Base go through the marketplace mech (mech 112), which charges 100 wei of native ETH per request, paid from the agent safe.
2. `fund_requirements` had the base safe's native entry at `topup: 0, threshold: 0`, so the Pearl middleware never auto-funded the safe with native ETH.
3. The trader treats the entire safe balance as investable. Any native ETH that did land on the safe was either invested into pools or swept to USDC by the non-whitelisted-asset cleanup, so there was never anything left to pay mech fees with. No mech requests means the staking KPI is missed and no rewards accrue.

Separately, users trying to withdraw from Velodrome v2 staked positions were stuck: withdrawals looped forever in the FSM (one user withdrawal pending since 2026-07-01).

## Changes

**1. Funding config (basius + optimus service/agent yamls)**
- Base safe native ETH requirement is now `topup: 100000000000000` (0.0001 ETH), `threshold: 350` wei, so the middleware funds the safe and tops it up when it runs low. At 100 wei per request the topup covers mech fees for far longer than the refill cadence.
- The USDC requirement on the base safe is removed (mech fees on Base are paid in native ETH, not USDC).
- The same `fund_requirements` dict is now also passed to `liquidity_trader_abci` (new skill param), for both basius and optimus, so the trader knows what the funding floor is.

**2. Mech fee reserve in the trader (`liquidity_trader_abci`)**
- New helpers `_get_mech_fee_reserve` / `_apply_mech_fee_reserve` in `behaviours/base.py`: the safe `topup` for a token in `fund_requirements` is treated as a reserve and deducted from the balance the trader considers investable (`max(0, balance - reserve)`).
- Applied at the selection layer (`_get_investable_balance` in `evaluate_strategy.py`) and at every execution-layer amount derivation in `decision_making.py`: velodrome investment amounts, token-balance based amounts, deposits, and swap route amounts.
- Withdrawals bypass the reserve: when `investing_paused` is set (the withdrawal flow), route amounts are not reduced, so users can always withdraw their full funds.

**3. Native ETH excluded from the swap-to-USDC cleanup**
- `check_and_prepare_non_whitelisted_swaps` now skips the zero address. Previously it would swap the safe's native ETH (the mech fee gas) into USDC and invest it.

**4. Velodrome v2 gauge unstake crash fix**
- Since 72cda47e the velodrome contract wrappers return tx data as `bytes` (`velodrome_gauge/contract.py` `withdraw` returns `bytes.fromhex(data[2:])`), but `unstake_lp_tokens` in `pools/velodrome.py` still re-applied `bytes.fromhex` on the result. Every non-CL gauge unstake raised `TypeError`, which was swallowed by the `except` in `get_unstake_lp_tokens_tx_hash`, returning no transaction and leaving the FSM looping on the withdrawal forever.
- Fix is a one-line pass-through of the wrapper's bytes. A sweep of all other `bytes.fromhex` call sites in the skill confirmed they consume MultiSend `get_tx_data` outputs or event-log fields, which are still hex strings, so they are correct and untouched.
- Once deployed, stuck withdrawals resume automatically from the exit-pools step, no manual intervention needed.

## Tests

- Unit tests for the reserve helpers (case-insensitive address match, unknown chain/token, invalid values, floor at zero).
- One test per execution-layer reserve site asserting the exact deducted amounts, plus the on/off pair for the withdrawal bypass.
- Native ETH skip test for the cleanup path.
- Unstake fix: the happy-path unit test now stubs the withdraw call with bytes, matching what the real wrapper returns (this test fails on the old code with the production `TypeError`), plus an integration-style test that runs the real `VelodromePoolBehaviour` through `get_unstake_lp_tokens_tx_hash` with only the contract boundary stubbed, asserting the gauge tx bytes reach the Safe call unchanged (also fails on the old code).

## Test plan

- [x] Full `liquidity_trader_abci` test suite passes locally
- [x] All CI checks run locally: black/isort/flake8/darglint, mypy/pylint, check-hash, check-packages, check-abci-docstrings, check-abciapp-specs, check-handlers, check-dependencies, liccheck, check-third-party-hashes, check-doc-hashes, gitleaks
- [x] `autonomy packages lock` + `lock --check`, `autonomy push-all` done
- [ ] Deploy new basius service hash via Pearl and confirm mech requests are funded and the stuck withdrawal completes

## On-chain payment-type evidence

- Base mech 112 (`0xe535D7AcDEeD905dddcb5443f41980436833cA2B`): `paymentType()` = `0xba699a34be8fe0e7725e93dcbce1701b0211a8ca61330aaeb8a05bf2ec7abed1` (`PaymentType.NATIVE`), `maxDeliveryRate()` = 100 wei. This is why the Base USDC safe entry is removed and a native reserve is added instead.
- Optimism mech 195 (`0x3208Dc0DebA018dDE8a8e2FFA09265Ea12aAf55E`): `paymentType()` = `0x6406bb5f31a732f898e1ce9fdd988a80a808d36ab5d9a4a4805a8be8d197d5e3` (`PaymentType.TOKEN_USDC`), `maxDeliveryRate()` = 10000 (0.01 USDC). This is why the Optimism reserve is on USDC (0.7 USDC covers ~70 requests).

### DIFF
--- a/packages/valory/agents/basius/aea-config.yaml
+++ b/packages/valory/agents/basius/aea-config.yaml
@@ -68,9 +68,9 @@ skills:
 - valory/abstract_abci:0.1.0:bafybeie5et43th2nta76cd3kpm3fmurd7oyzk3qxim76octvt56m6r2poi
 - valory/abstract_round_abci:0.1.0:bafybeicioxhbhnsoqyqvkeevd7df7r2vubbiglyf7ucps2el73objazjsq
 - valory/funds_manager:0.1.0:bafybeibne4vkmbper3ryn7q6kt7lclbyylgorziscyn6msr5qt46aihv4m
-- valory/liquidity_trader_abci:0.1.0:bafybeibxkaukxrdjvkqmen5i4v67ouvyfve6v7gcxabscvozgx7ffxndde
+- valory/liquidity_trader_abci:0.1.0:bafybeiboc2qvgto5t6cr3snwo4qmxxce5l24wzcqbqmplk5foy7gtsyipa
 - valory/mech_interact_abci:0.1.0:bafybeidnmszufszq3ro4ebgrqzmhqgctrsc6sykeanfsbvotghvt2cgfsy
-- valory/optimus_abci:0.1.0:bafybeihzmdil44ixvorq37cpdmijeqr7tiaxsu3etyv4odbv2b7vsnan7e
+- valory/optimus_abci:0.1.0:bafybeibblze7sbrya54zilbpi6dkjvxd5uod76vshwxi4gafrhrq74rhza
 - valory/registration_abci:0.1.0:bafybeihjhbn5seffyu5cwugvm26fkamp3svhggmtzh577daijgl2ksxs3e
 - valory/reset_pause_abci:0.1.0:bafybeihc4ruh25s7wuunfy2b3k2n4eypyhvtry3usgom44wsymagbcr3ki
 - valory/termination_abci:0.1.0:bafybeibnnzry5kknci3z46gcjgutv4jid3ir6xn6a5adq3tx6v2e25augq
--- a/packages/valory/agents/basius/aea-config.yaml
+++ b/packages/valory/agents/basius/aea-config.yaml
@@ -68,9 +68,9 @@ skills:
 - valory/abstract_abci:0.1.0:bafybeie5et43th2nta76cd3kpm3fmurd7oyzk3qxim76octvt56m6r2poi
 - valory/abstract_round_abci:0.1.0:bafybeicioxhbhnsoqyqvkeevd7df7r2vubbiglyf7ucps2el73objazjsq
 - valory/funds_manager:0.1.0:bafybeibne4vkmbper3ryn7q6kt7lclbyylgorziscyn6msr5qt46aihv4m
-- valory/liquidity_trader_abci:0.1.0:bafybeigoe5msyzv2ompaxvhyijqf54666nltucnblqzuezysfesnzsjip4
+- valory/liquidity_trader_abci:0.1.0:bafybeiggg2skakianedd7c6i53n4svap7zxwea7qoo5wtcurio2ycyhyey
 - valory/mech_interact_abci:0.1.0:bafybeidnmszufszq3ro4ebgrqzmhqgctrsc6sykeanfsbvotghvt2cgfsy
-- valory/optimus_abci:0.1.0:bafybeihoneiqnskjd7eskcngaauypulhoqh5apvklrkbz6irpnt2556qnm
+- valory/optimus_abci:0.1.0:bafybeighy5be2v2czqrvqit7kfnj5grmlhqp2wikb2d65hmwvx3jak47zi
 - valory/registration_abci:0.1.0:bafybeihjhbn5seffyu5cwugvm26fkamp3svhggmtzh577daijgl2ksxs3e
 - valory/reset_pause_abci:0.1.0:bafybeihc4ruh25s7wuunfy2b3k2n4eypyhvtry3usgom44wsymagbcr3ki
 - valory/termination_abci:0.1.0:bafybeibnnzry5kknci3z46gcjgutv4jid3ir6xn6a5adq3tx6v2e25augq
@@ -192,7 +192,7 @@ type: skill
 models:
   params:
     args:
-      fund_requirements: ${dict:{"base":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":0,"threshold":0},"0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913":{"topup":700000,"threshold":350000}}}}}
+      fund_requirements: ${dict:{"base":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":100000000000000,"threshold":350}}}}}
       rpc_urls: ${dict:{"base":"https://mainnet.base.org"}}
       safe_contract_addresses: ${SAFE_CONTRACT_ADDRESSES:dict:{"base":"0x0000000000000000000000000000000000000000"}}
 ---
@@ -352,6 +352,7 @@ models:
       balancer_graphql_endpoints: ${str:{"optimism":"https://api.studio.thegraph.com/query/75376/balancer-optimism-v2/version/latest","base":"https://api.studio.thegraph.com/query/24660/balancer-base-v2/version/latest","mode":"https://api.studio.thegraph.com/query/75376/balancer-mode-v2/version/latest"}}
       target_investment_chains: ${TARGET_INVESTMENT_CHAINS:list:["optimism"]}
       staking_chain: ${str:optimism}
+      fund_requirements: ${dict:{"base":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":100000000000000,"threshold":350}}}}}
       file_hash_to_strategies: ${FILE_HASH_TO_STRATEGIES:str:{"bafybeib2szveecjhwlu5aeww5umtso5sm4btogsrjuqg4tk5uormjmsppi":"max_apr_selection","bafybeiaktss4rvsyiqg43yauc3aj5mq3ztysgqopc2dvxkslswj25kdtcm":"balancer_pools_search","bafybeiglbwfrhxbgqyyg6oniwur2x6sswmi5ed55crzvrhr3o7p5x3kmne":"uniswap_pools_search","bafybeifiotxje5ndkdj673oo2uxfpj76fd32qjp7tpxbbdmidbzcqna23a":"asset_lending","bafybeielyh7i7qjm54h3tne4gcgiuqutqavmczoqc6bguupqnnjzwjtvm4":"velodrome_pools_search"}}
       strategies_kwargs: ${str:{"merkl_pools_search":{"balancer_graphql_endpoints":{"optimism":"https://api.studio.thegraph.com/query/75376/balancer-optimism-v2/version/latest","base":"https://api.studio.thegraph.com/query/24660/balancer-base-v2/version/latest","mode":"https://api.studio.thegraph.com/query/75376/balancer-mode-v2/version/latest"},"merkl_fetch_campaign_args":{"url":"https://api.merkl.xyz/v3/campaigns","creator":"","live":"false"}},"uniswap_pools_search":{"graphql_endpoints":{"optimism":"https://gateway.thegraph.com/api/575c6d9779719bc1ced85444c98441be/subgraphs/id/7SVwgBfXoWmiK6x1NF1VEo1szkeWLniqWN1oYsX3UMb5","base":"https://gateway.thegraph.com/api/575c6d9779719bc1ced85444c98441be/subgraphs/id/GqzP4Xaehti8KSfQmv3ZctFSjnSUYZ4En5NRsiTbvZpz"}},"balancer_pools_search":{"graphql_endpoint":"https://api-v3.balancer.fi/"},"velodrome_pools_search":{"graphql_endpoint":"https://gateway.thegraph.com/api/b8e4cf1b314c67d2a0109325046a7464/subgraphs/id/7tA4PY1VmbycJeoVtn2mjQK4NbozgwTuZgrxDTxzEDL1"},"asset_lending":{"endpoint":"https://us-central1-stu-dashboard-a0ba2.cloudfunctions.net/v2Aggregators","lending_asset":"0x4200000000000000000000000000000000000006"}}}
       available_protocols: ${list:["VELODROME"]}
--- a/packages/valory/agents/basius/aea-config.yaml
+++ b/packages/valory/agents/basius/aea-config.yaml
@@ -68,9 +68,9 @@ skills:
 - valory/abstract_abci:0.1.0:bafybeie5et43th2nta76cd3kpm3fmurd7oyzk3qxim76octvt56m6r2poi
 - valory/abstract_round_abci:0.1.0:bafybeicioxhbhnsoqyqvkeevd7df7r2vubbiglyf7ucps2el73objazjsq
 - valory/funds_manager:0.1.0:bafybeibne4vkmbper3ryn7q6kt7lclbyylgorziscyn6msr5qt46aihv4m
-- valory/liquidity_trader_abci:0.1.0:bafybeiboc2qvgto5t6cr3snwo4qmxxce5l24wzcqbqmplk5foy7gtsyipa
+- valory/liquidity_trader_abci:0.1.0:bafybeidriubtvnzbstj2ydczozj3xehc6e5dknp6dlma5urri43vrnqpnq
 - valory/mech_interact_abci:0.1.0:bafybeidnmszufszq3ro4ebgrqzmhqgctrsc6sykeanfsbvotghvt2cgfsy
-- valory/optimus_abci:0.1.0:bafybeibblze7sbrya54zilbpi6dkjvxd5uod76vshwxi4gafrhrq74rhza
+- valory/optimus_abci:0.1.0:bafybeid43lif2iegegesbnglc65pondbvadmxaftk6xvgqoer4djpobyba
 - valory/registration_abci:0.1.0:bafybeihjhbn5seffyu5cwugvm26fkamp3svhggmtzh577daijgl2ksxs3e
 - valory/reset_pause_abci:0.1.0:bafybeihc4ruh25s7wuunfy2b3k2n4eypyhvtry3usgom44wsymagbcr3ki
 - valory/termination_abci:0.1.0:bafybeibnnzry5kknci3z46gcjgutv4jid3ir6xn6a5adq3tx6v2e25augq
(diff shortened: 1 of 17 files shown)


packages/valory/agents/basius/aea-config.yaml
```diff
@@ -192,7 +192,7 @@ type: skill
 models:
   params:
     args:
-      fund_requirements: ${dict:{"base":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":0,"threshold":0},"0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913":{"topup":700000,"threshold":350000}}}}}
+      fund_requirements: ${dict:{"base":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":100000000000000,"threshold":350}}}}}
```
OjusWiZard, line 195: **Needs confirmation:** this deletes the Base USDC safe entry that 0286d1f7 added one PR earlier with the opposite rationale ("mech payments settle in USDC — FixedPriceTokenUSDC mech flavour on Base mech 112", after an incident where the agent silently ran out of USDC). The payment type is fetched from the mech at runtime (`mech_interact_abci` `PaymentType.NATIVE` vs `TOKEN_USDC`), so the deciding fact is mech 112's on-chain payment type. If it is indeed fixed-price-native (100 wei) as this PR states, this removal is correct — but please cite the on-chain evidence in the PR body, since two same-week commits now assert opposite facts, and if mech 112 is actually `TOKEN_USDC` this exactly reintroduces the bug 0286d1f7 fixed: no USDC deficit tracking *and* zero USDC reserved from investment.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1217,6 +1217,44 @@ def _get_balance(
...
+        The reserve is the safe topup declared in ``fund_requirements`` (the
+        same config the funds_manager skill reports to the operator for
+        refills). Reserving exactly the topup keeps the refilled amount out
+        of investments so mech payments can't be invested away.
+
+        :param chain: chain identifier (e.g. ``"base"``).
+        :param token: token address; ``ZERO_ADDRESS`` for native.
+        :return: reserved amount in wei/smallest unit, 0 if none configured.
+        """
+        safe_requirements = self.params.fund_requirements.get(chain, {}).get("safe", {})
```
OjusWiZard: **Important:** the try/except only guards the leaf `int(requirement.get("topup", 0))`. If `fund_requirements[chain]` (or its `"safe"` value) isn't a dict — bad env override, YAML mistake — `.get("safe", {})` / `.items()` raises an uncaught `AttributeError` here, outside the guard. Because this helper runs at every reserve site, that takes down the whole DecisionMaking/EvaluateStrategy round instead of degrading gracefully like the leaf-malformed case, contradicting the docstring's "0 if none configured" promise. Suggest extending the guard to the container lookups (or validating once at load — see the `models.py` comment). No test covers this shape either.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
```diff
@@ -1217,6 +1217,44 @@ def _get_balance(
...
+        for address, requirement in safe_requirements.items():
+            if address.lower() == token.lower():
+                try:
+                    return int(requirement.get("topup", 0))
```
OjusWiZard: **Important:** a *valid-but-wrong* topup slips through silently: `int(-500)`, a bool, or `"-500"` all convert without exception, so no warning fires, and `_apply_mech_fee_reserve` then treats `reserve <= 0` as "no reserve" and returns the full balance — with no log at any level. A sign-flip config typo would silently reproduce the invested-away-mech-fees bug this PR fixes. Suggest treating `topup < 0` the same as the malformed branch (warn + return 0). The existing `test_reserve_malformed_topup_warns_and_disables` only covers `"not-a-number"`/`None`.

packages/valory/skills/liquidity_trader_abci/models.py
```diff
@@ -522,6 +522,12 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.staking_chain: Optional[str] = self._ensure(
             "staking_chain", kwargs, Optional[str]
         )
+        # Same shape as the funds_manager skill's fund_requirements. The safe
+        # topups double as mech-fee reserves: that portion of the safe balance
+        # is excluded from investment so refills aren't invested away.
+        self.fund_requirements: Dict[str, Any] = self._ensure(
```
OjusWiZard, line 528: **Suggestion (root-cause fix for the two `base.py` findings):** `funds_manager/models.py` already has `FundRequirements.from_dict` with typed `TokenRequirement(topup, threshold)`, and this skill already depends on `funds_manager`. Validating through that here would turn every bad-config shape (non-dict containers, non-int topup, negative values) into a load-time error and remove the defensive try/except in `_get_mech_fee_reserve` entirely.

Also, the comment above says "same shape as the funds_manager skill's fund_requirements" — worth noting explicitly that it's a *manually duplicated copy* in the agent/service yamls with nothing enforcing sync; an operator editing one without the other silently desyncs the mech-fee reserve from the actual refill target.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
```diff
@@ -2878,6 +2903,28 @@ def test_zero_available(self):
         )
         assert result is None
 
+    def test_mech_fee_reserve_deducted(self):
```
OjusWiZard: **Test gap:** this only reserves on `assets[0]`, so the token1 `_apply_mech_fee_reserve` call's *effect* is never asserted — `desired1` is never capped by a reduced `token1_balance`. A copy-paste bug keying both calls on `assets[0]` (or deleting the second call) would keep 100% line coverage and all tests green while silently un-reserving token1's share of a Velodrome deposit. Suggest parametrizing over both tokens like the sibling `TestGetTokenBalancesAndCalculateAmounts.test_mech_fee_reserve_deducted` already does.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
```diff
@@ -1879,6 +1888,18 @@ def get_deposit_tx_hash(
...
+
+        raw_balance = token_balance
+        token_balance = self._apply_mech_fee_reserve(chain, asset, token_balance)
+        if token_balance == 0 and raw_balance > 0:
+            self.context.logger.warning(
+                f"Balance of {asset} on {chain} is fully reserved for mech fees; "
+                "skipping deposit."
+            )
+            return None, None, None
         amount = int(min(token_balance * relative_funds_percentage, token_balance))
```
OjusWiZard: **Advisory:** the new guard above catches only the fully-reserved case. If the reserve leaves a small positive remainder and `relative_funds_percentage` is small, this line can still round `amount` to 0, which then falls into the pre-existing generic `"Missing information in action"` error — mislabeling a mech-fee-reserve-induced zero as a data problem. Consider extending the warning to cover `amount == 0` after the reserve was applied.

packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
```diff
@@ -1380,11 +1380,14 @@ def check_and_prepare_non_whitelisted_swaps(
                     asset_symbol = asset.get("asset_symbol")
                     balance = asset.get("balance", 0)
 
-                    # Skip if no balance or asset is whitelisted
+                    # Skip if no balance or asset is whitelisted. The native
+                    # token is never swept: it is reserved for gas and, where
+                    # the mech charges in native, for mech request fees.
```
OjusWiZard: **Nit (comment accuracy):** "where the mech charges in native" attaches a conditional justification to code that unconditionally skips `ZERO_ADDRESS` — there's no branch on mech fee currency. Suggest dropping the clause and stating plainly that native balance is always kept out of the USDC sweep (gas + mech fees), with `_apply_mech_fee_reserve` as the separate fee-reserve mechanism.

OjusWiZard (commented):
Multi-perspective review (correctness, silent failures, type design, comments, tests, simplification). Confirmed clean: no double-application of the reserve across selection/execution layers; the `investing_paused` bypass is structurally safe via the final `min()`; the velodrome bytes pass-through matches its consumer (`data=tx_hash` into `get_raw_safe_transaction_hash`); modified test suites pass (1798 passed, 0 failed). Inline comments below for the actionable findings.

Non-anchorable advisories (pre-existing code adjacent to this PR):
- `get_unstake_lp_tokens_tx_hash`'s broad `except Exception` logs only `str(e)` with no exception type or traceback — the original velodrome `TypeError` hid here. Cheap to improve while in the area.
- `check_and_prepare_non_whitelisted_swaps` returns `[]` on any exception, making "sweep crashed" indistinguishable from "nothing to sweep" for the new native-ETH skip too.
- Optional simplification: the duplicated token0/token1 reserve application in `_get_token_balances_and_calculate_amounts` and `_calculate_velodrome_investment_amounts` could share a small pair-helper.
- The 350-wei threshold vs 1e14 topup asymmetry is understood as intentional per the PR description (refill when ~3 requests remain); just noting the funds-status deficit alert will essentially never fire until the buffer is nearly exhausted.

packages/valory/services/optimus/service.yaml
```diff
@@ -114,6 +114,7 @@ models:
       allowed_chains: ${ALLOWED_CHAINS:list:["base","optimism","mode"]}
       target_investment_chains: ${TARGET_INVESTMENT_CHAINS:list:["base","optimism","mode"]}
       staking_chain: ${STAKING_CHAIN:str:""}
+      fund_requirements: ${FUND_REQUIREMENTS:dict:{"optimism":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":0,"threshold":0},"0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85":{"topup":700000,"threshold":350000}}}}}
```
bennyjo, line 117: The reserve helper treats every `safe` topup here as a mech-fee reserve, so this `0.7 USDC` entry now withholds 0.7 USDC from investment on every optimism position — a behavior change beyond the Base fix, and not called out in the description. Non-blocking for the urgent merge: just confirm the optimism mech actually charges in USDC (so this is the right token to reserve); if this topup is only a middleware funding floor, the reserve should be scoped to the mech-fee token instead.

bennyjo (dismissed):
Approving — the basius reward fix and the Velodrome unstake fix are both correct and verified against source. One non-blocking follow-up inline on the optimism reserve.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 9e278423. The whole lookup (chain container, safe container, and the token entry) is now inside the try, so any non-dict level degrades to reserve-disabled with a warning instead of raising. Added parametrized tests for all three malformed container shapes.

packages/valory/skills/liquidity_trader_abci/behaviours/base.py
DIvyaNautiyal07: Fixed in 9e278423. A negative topup is now treated the same as the malformed branch: warning logged and reserve disabled. Added -500 and "-500" to the malformed-topup test cases.

packages/valory/skills/liquidity_trader_abci/tests/behaviours/test_decision_making.py
DIvyaNautiyal07: Fixed in 9e278423, parametrized over both tokens ([600, 800] and [800, 600]) so deleting or mis-keying either reserve call now fails the test.

packages/valory/skills/liquidity_trader_abci/behaviours/decision_making.py
DIvyaNautiyal07: Fixed in 9e278423. The guard now checks amount == 0 after the reserve reduced the balance, so the rounding case (small remainder times small percentage) gets the specific warning too instead of falling into the generic error. Added a test for that case.

packages/valory/skills/liquidity_trader_abci/behaviours/evaluate_strategy.py
DIvyaNautiyal07: Reworded in 9e278423. The comment now just says native is always kept out of the USDC sweep (gas and mech fees) and points to _apply_mech_fee_reserve as the separate fee-reserve mechanism.

packages/valory/agents/basius/aea-config.yaml
DIvyaNautiyal07: Confirmed on-chain: paymentType() on mech 112 (0xe535D7AcDEeD905dddcb5443f41980436833cA2B, Base) returns 0xba699a34be8fe0e7725e93dcbce1701b0211a8ca61330aaeb8a05bf2ec7abed1, which is PaymentType.NATIVE in mech_interact, and maxDeliveryRate() returns 100 wei. So the mech is fixed-price-native and the USDC rationale from 0286d1f7 doesn't apply to this mech. The native reserve added in this PR is the matching protection. I'll add this evidence to the PR body.

packages/valory/skills/liquidity_trader_abci/models.py
DIvyaNautiyal07: Agree it's the cleaner long-term shape, and funds_manager is indeed already a dependency. For this fix I kept the validation local instead: 9e278423 wraps the whole lookup in the guard and rejects negative topups, which covers the same bad-config space (non-dict containers, non-int, negative) with tests, without coupling the hot path to funds_manager's pydantic models mid-hotfix. Happy to move to FundRequirements.from_dict at load time in a follow-up. And yes, fair point on the manual duplication, nothing enforces sync between the two yaml copies today.

packages/valory/services/optimus/service.yaml
DIvyaNautiyal07: Confirmed: paymentType() on the optimism priority mech 195 (0x3208Dc0DebA018dDE8a8e2FFA09265Ea12aAf55E) returns the TOKEN_USDC type hash and maxDeliveryRate() is 10000 (0.01 USDC per request), so USDC is the right token to reserve there and the 0.7 USDC buffer covers about 70 requests. The withholding on optimism is intentional.

bennyjo, line 117: Checking intent: this now holds a live 0.7 USDC reserve out of investment on optimism (the safe USDC `topup`), but the PR is framed as a base/native-ETH mech-fee fix. Is reserving USDC on optimism intended here, or just carried over from the existing funds_manager config?

packages/valory/services/basius/service.yaml
```diff
@@ -114,6 +114,7 @@ models:
       allowed_chains: ${ALLOWED_CHAINS:list:["base","optimism","mode"]}
       target_investment_chains: ${TARGET_INVESTMENT_CHAINS:list:["base"]}
       staking_chain: ${STAKING_CHAIN:str:base}
+      fund_requirements: ${FUND_REQUIREMENTS:dict:{"base":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":200000000000000,"threshold":100000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":100000000000000,"threshold":350}}}}}
```
bennyjo, line 117: Two intent checks on this line: (1) `threshold: 350` wei breaks the old `≈ topup/2` convention (the removed USDC entry was 700000/350000) and reads as arbitrary — deliberate? (2) Dropping the base-safe USDC requirement removes the middleware's USDC top-up floor, not just a mech-fee entry — is it intended that USDC is no longer auto-funded on the safe?

bennyjo (approved):
Approving. The mech-fee reserve and the Velodrome unstake bytes pass-through are correct and well-tested, and the latest commit hardens the config parsing nicely. Two non-blocking intent questions on the funding config left inline.

OjusWiZard (approved):
Approving after two review rounds. Round one (6-agent review) findings were all addressed in 9e278423 and verified in code: hardened fund_requirements parsing (container guards + negative-topup rejection), parametrized token0/token1 reserve tests, reserve-aware deposit-skip warning, and comment fixes. The basius USDC removal is justified by the on-chain paymentType() evidence now cited in the PR body (mech 112 is fixed-price-native, 100 wei). Round two on the hardening delta found nothing blocking: withdrawal paths verified to sweep full balances (only fetch_routes needs and has the investing_paused bypass), the reserve is correctly decoupled from the mech payment tx path, CI green on all platforms, 2471 tests passing locally.

Remaining non-blocking follow-ups tracked in the open models.py thread: validate fund_requirements via funds_manager's FundRequirements.from_dict at load time, and a test for a malformed entry shadowing a well-formed one in the same safe dict.

packages/valory/services/optimus/service.yaml
DIvyaNautiyal07: yes it is intended, optimus uses usdc mechs wherease base uses native mechs

packages/valory/services/basius/service.yaml
DIvyaNautiyal07: yes it’s intentional, since the mech fee is too low, we kept threshold small, but top-up reasonably sufficient to last for some time.

PR #359, opened 2026-07-06
chore: bump mech-interact to v0.32.4

DIvyaNautiyal07:
## Summary
- Repins `mech_interact_abci` to the [v0.32.4](https://github.com/valory-xyz/mech-interact/releases/tag/v0.32.4) release hash (`bafybeici5tsg...`), propagated to both agents, both services, and the dependent skills via `autonomy packages lock`.
- v0.32.4 removes the dedicated `RESPONSE_ROUND_TIMEOUT` event (reverting to the shared `ROUND_TIMEOUT`), so `optimus_abci/fsm_specification.yaml` is regenerated: `(MechResponseRound, ROUND_TIMEOUT) -> CheckStakingKPIMetRound`.
- Moves the `upstream_pins` entry in `pyproject.toml` from `v0.32.4-rc2` to the final `v0.32.4` tag.

No optimus-side code changes needed: `use_offchain` stays `false` in all configs, so the on-chain path is unchanged and the new off-chain poll-budget guard never engages.

## Test plan
- [x] `check-hash`, `check-packages`, `check-dependencies`, `check-third-party-hashes`, `uv lock --check`
- [x] `check-abciapp-specs` (regenerated FSM), `check-abci-docstrings`, `check-handlers`
- [x] black / isort / flake8 / mypy / pylint / darglint / bandit
- [x] `check-doc-hashes`, `liccheck`, gitleaks
- [x] `py3.10-darwin`: 4076 tests pass

### DIFF
--- a/packages/valory/skills/optimus_abci/fsm_specification.yaml
+++ b/packages/valory/skills/optimus_abci/fsm_specification.yaml
@@ -24,7 +24,6 @@ alphabet_in:
 - OFFCHAIN_MECH_DEPOSIT_SETTLED
 - RESET_AND_PAUSE_TIMEOUT
 - RESET_TIMEOUT
-- RESPONSE_ROUND_TIMEOUT
 - ROUND_TIMEOUT
 - SERVICE_EVICTED
 - SERVICE_NOT_STAKED
@@ -160,7 +159,7 @@ transition_func:
     (MechRequestRound, SKIP_REQUEST): GetPositionsRound
     (MechResponseRound, DONE): CheckStakingKPIMetRound
     (MechResponseRound, NO_MAJORITY): MechResponseRound
-    (MechResponseRound, RESPONSE_ROUND_TIMEOUT): CheckStakingKPIMetRound
+    (MechResponseRound, ROUND_TIMEOUT): CheckStakingKPIMetRound
     (MechVersionDetectionRound, NO_MAJORITY): MechVersionDetectionRound
     (MechVersionDetectionRound, NO_MARKETPLACE): MechRequestRound
     (MechVersionDetectionRound, ROUND_TIMEOUT): MechVersionDetectionRound
--- a/packages/valory/skills/optimus_abci/tests/test_models.py
+++ b/packages/valory/skills/optimus_abci/tests/test_models.py
@@ -21,7 +21,8 @@
 
 # pylint: skip-file
 
-from unittest.mock import MagicMock, patch
+from typing import Any
+from unittest.mock import MagicMock, call, patch
 
 from packages.valory.skills.optimus_abci.composition import OptimusAbciApp
 from packages.valory.skills.optimus_abci.models import (
@@ -40,13 +41,21 @@ def test_import() -> None:
 class TestParams:
     """Test Params class."""
 
-    def test_init_extracts_service_endpoint_base(self) -> None:
-        """Test Params __init__ extracts service_endpoint_base."""
+    def test_init_extracts_service_endpoint_base_and_mech_timeout(self) -> None:
+        """Test Params __init__ extracts service_endpoint_base and mech_interact_round_timeout_seconds."""
         mock_context = MagicMock()
         mock_context.skill_id = "test_skill/test:0.1.0"
+
+        def _ensure_side_effect(name: str, _kwargs: Any, type_: Any) -> Any:
+            if name == "service_endpoint_base":
+                return "http://localhost:8000"
+            if name == "mech_interact_round_timeout_seconds":
+                return 900
+            raise AssertionError(f"Unexpected _ensure call for {name}")
+
         with (
             patch.object(
-                Params, "_ensure", return_value="http://localhost:8000"
+                Params, "_ensure", side_effect=_ensure_side_effect
             ) as mock_ensure,
             patch.object(Params.__bases__[0], "__init__", return_value=None),
         ):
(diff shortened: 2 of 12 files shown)


packages/valory/skills/optimus_abci/fsm_specification.yaml
```diff
@@ -160,7 +159,7 @@ transition_func:
     (MechRequestRound, SKIP_REQUEST): GetPositionsRound
     (MechResponseRound, DONE): CheckStakingKPIMetRound
     (MechResponseRound, NO_MAJORITY): MechResponseRound
-    (MechResponseRound, RESPONSE_ROUND_TIMEOUT): CheckStakingKPIMetRound
+    (MechResponseRound, ROUND_TIMEOUT): CheckStakingKPIMetRound
```
OjusWiZard, line 162: With the dedicated `RESPONSE_ROUND_TIMEOUT` (330s class default) gone, this edge now fires on `MechInteractEvent.ROUND_TIMEOUT` — whose class default in `mech_interact_abci/rounds.py` is **30.0s**. `optimus_abci/models.py` `SharedState.setup()` only builds timeout overrides for `LiquidityTraderEvent`/`TransactionSettlementEvent`/`ResetPauseEvent`/`RegistrationEvent`, so unlike every sibling repo (market-resolver 1200s, trader 5400s, meme-ooorr 600s), optimus keeps the 30s default for the mech rounds.

To be fair: `main` (pinned to v0.32.2) behaves identically, so this is a pre-existing gap rather than a regression, and the poll-then-discard design routes both `DONE` and `ROUND_TIMEOUT` to `CheckStakingKPIMetRound`, so nothing is lost financially. But the on-chain marketplace gives mechs `response_timeout: 300`s to answer, so with a 30s round window the response leg will time out on essentially every cycle — the `DONE`-path diagnostics (mech error logging, `penalized_mechs` tracking) become dead paths, and if `use_offchain` is ever flipped on, v0.32.4's `_check_round_timeout_fits_poll_budget` guard will raise `ValueError` (30s < 330s) on the first off-chain cycle.

Suggestion: add `MechInteractEvent` to the override tuple in `SharedState.setup()` (wired to a param covering on-chain latency, e.g. ≥330s), either in this PR or as an immediate follow-up.

OjusWiZard (commented):
Hash propagation and the FSM regen both verified clean against upstream v0.32.4 (consistent hash across all 9 files, no stale `RESPONSE_ROUND_TIMEOUT`/rc references at PR head, `use_offchain: false` everywhere as described). One change requested before this lands — see the inline comment on the regenerated FSM spec: optimus is the only consumer repo that never overrides `MechInteractEvent.ROUND_TIMEOUT`, so this edge now fires at the 30s class default.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

bennyjo (approved):
Verified the pin against the mech-interact v0.32.4 tag and the regenerated FSM spec — clean.

packages/valory/skills/optimus_abci/fsm_specification.yaml
DIvyaNautiyal07: Fixed in `0c9921bc`. `optimus_abci.SharedState.setup()` now overrides `MechInteractEvent.ROUND_TIMEOUT` via a new `mech_interact_round_timeout_seconds` param (default `900s` in `skill.yaml`, env-overridable as `MECH_INTERACT_ROUND_TIMEOUT_SECONDS` in both service.yamls and both aea-configs). 900s clears v0.32.4's `_check_round_timeout_fits_poll_budget` guard (>= 330s) with headroom, and gives on-chain responses ample buffer over their 300s window so the DONE-path diagnostics stop being dead. Mirrors the pattern trader ships. Test updated to cover both `_ensure` calls (4076 passed).

packages/valory/skills/optimus_abci/tests/test_models.py
```diff
@@ -40,30 +41,45 @@ def test_import() -> None:
...
+                    "mech_interact_round_timeout_seconds",
+                    expected_kwargs,
+                    type_=int,
+                ),
+            ]
             assert params.service_endpoint_base == "http://localhost:8000"
+            assert params.mech_interact_round_timeout_seconds == 900
 
 
 class TestSharedState:
```
OjusWiZard, line 85: Non-blocking: the behavioral fix in `0c9921bc` is the `event_to_timeout[MechInteractEvent.ROUND_TIMEOUT]` wiring in `SharedState.setup()` (`models.py:159-161`), but `test_setup` in this class still only asserts the four pre-existing overrides (`VALIDATE_TIMEOUT`, `LiquidityTraderEvent.ROUND_TIMEOUT`, `RESET_AND_PAUSE_TIMEOUT`, `RegistrationEvent.ROUND_TIMEOUT`) — the new `TestParams` test covers extraction only. If a future refactor drops the override line, the suite stays green, which is the exact regression this PR exists to prevent. Worth one more assertion: set `mech_interact_round_timeout_seconds` to a concrete value on the mock params and assert `OptimusAbciApp.event_to_timeout[MechInteractEvent.ROUND_TIMEOUT]` equals it.

OjusWiZard (approved):
Approving. Verified the delta since my review comment: `mech_interact_round_timeout_seconds` is declared in `skill.yaml` (900) and overridden consistently as `${MECH_INTERACT_ROUND_TIMEOUT_SECONDS:int:900}` in both agents and both services; the override in `SharedState.setup()` lands after the generic `round_timeout_overrides` build so nothing clobbers it; 900s clears v0.32.4's 330s off-chain poll-budget guard with headroom; and the fingerprint/hash chain plus the cffi pin are consistent at head. One non-blocking test note inline.

bennyjo (approved):
Re-review at head: the new `mech_interact_round_timeout_seconds` override (models.py setup) is applied after the generic round-timeout build so nothing clobbers it, `MechInteractEvent.ROUND_TIMEOUT` is a distinct key, and 900s clears the 330s off-chain guard with buffer over the 300s on-chain window — fixes the 30s-default gap from my earlier pass. Test is now discriminating; CI green including re-locked fingerprints. Approving.

PR #361, opened 2026-07-07
fix(optimus_abci): add fund_requirements default to composed skill.yaml

DIvyaNautiyal07:
## Summary

Unblocks the standalone agent-runner build that failed on the [v0.12.4 Release Flow](https://github.com/valory-xyz/optimus/actions/runs/28879174065/job/85662585883):

\`\`\`
aea.exceptions.AEAEnforceError: 'fund_requirements' of type 'typing.Dict[str, typing.Any]' required,
but it is not set in models.params.args of skill.yaml of valory/optimus_abci:0.1.0
\`\`\`

## Root cause

\`liquidity_trader_abci/models.py:528\` declares \`fund_requirements\` as a required param via \`self._ensure("fund_requirements", kwargs, Dict[str, Any])\`. \`optimus_abci\` composes that base skill and its \`skill.yaml\` sets the sibling required params under \`models.params.args\` (\`target_investment_chains\`, \`staking_chain\`, \`file_hash_to_strategies\`, \`strategies_kwargs\`, \`max_slippage_percentage\`, etc.) but was **missing \`fund_requirements\`**.

\`aea-config.yaml\` and \`service.yaml\` for both \`basius\` and \`optimus\` agents already override \`fund_requirements\` at deploy time (chain-scoped topups + thresholds), so live deployments work. The \`check-agent-runner\` smoke test in the Release Flow runs the built PyInstaller binary without those agent-level overrides applied — that's the path where the skill-level default matters, and where \`AEAEnforceError\` fires.

## Fix

- \`packages/valory/skills/optimus_abci/skill.yaml\`: adds \`fund_requirements: {}\` between \`staking_chain\` and \`file_hash_to_strategies\`, matching the order of the corresponding \`_ensure\` calls in \`liquidity_trader_abci/models.py\`.
- \`autonomy packages lock\` cascades the skill hash into both agent (\`basius\`, \`optimus\`) and service (\`basius\`, \`optimus\`) hashes.

The empty dict is a safe default: aea-config overrides on both agents still land the real chain-scoped fund_requirements at boot; the check-agent-runner smoke test now boots the composed skill without complaint.

## Release status recap

v0.12.4 shipped docker images, package hashes, and multi-arch manifests successfully. Only the standalone agent-runner binaries (and therefore \`upload-assets\`) failed. Once this lands on \`main\`, cutting v0.12.5 should produce a green release across all matrix entries.

## Test plan

- [ ] CI green (lock_check, copyright_and_dependencies_check, linter_checks, tests)
- [ ] \`build-agent-runner\` job on the release cut from this branch (v0.12.5) green across all matrix entries
- [ ] Basius / Optimus agent boots on staging deployment with the aea-config \`fund_requirements\` override honoured (no regression)

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -24,11 +24,11 @@
         "contract/valory/balancer_queries/0.1.0": "bafybeihbexpjj6ho3ij7rxk4tzz3sowpa5ztawlf63y5fdmmom7xifpan4",
         "contract/valory/mech_activity/0.1.0": "bafybeienezl3yozrionpoyh75r54e24tjtnstxv674ru25lnosps3a7cr4",
         "skill/valory/liquidity_trader_abci/0.1.0": "bafybeianso2jgr74rfyxfvwugkzlmmdqmwqyaxgpoquyjozjbwrvlhdz64",
-        "skill/valory/optimus_abci/0.1.0": "bafybeihclddpjmfu6waw4ph4xz2pwxjtj4uerbbdd2442ugtotwgggfyh4",
-        "agent/valory/optimus/0.1.0": "bafybeihidpdpzchbxqnndnjgk4uqmhgzcjwgqsaafceux2a2hyraogjshq",
-        "agent/valory/basius/0.1.0": "bafybeifubm255az2d37huxdg5pt5xmjv3oq5h2ms3zqwxu2eze2htjv6dy",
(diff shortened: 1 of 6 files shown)


bennyjo (approved):
Verified root cause (models.py:528 `_ensure`), cross-checked all required `Params` keys against the composed skill.yaml (fund_requirements was the only gap), and confirmed both agents' aea-config `${dict:...}` overrides still carry the real chain-scoped values. Lock cascade validated by CI. LGTM.

iasonrovis (approved):

PR #367, opened 2026-08-10
chore(optimus): bump mech_interact_abci to v0.32.7; hoist use_offchain and offchain_deposit_target_calls to top-level params

DIvyaNautiyal07:
## Summary

Companion to [valory-xyz/mech-interact#113](https://github.com/valory-xyz/mech-interact/pull/113) ([release v0.32.7](https://github.com/valory-xyz/mech-interact/releases/tag/v0.32.7)).

Two knobs — `use_offchain` and `offchain_deposit_target_calls` — moved out of the nested `MechMarketplaceConfig` dataclass into top-level `MechParams` attributes upstream. This PR wires the change through every optimus-side overlay so the composed `MechParams.__init__` does not hit the v0.32.7 fail-loud migration guard.

## Changes

- **`packages/packages.json`** — `mech_interact_abci` third_party hash bumped to `bafybeic5xfxdod5l56o7aojdamojnvi562nql2mlah73bf47d4guz4xnwi`. Cascade re-lock through `liquidity_trader_abci`, `optimus_abci`, both agents (`basius`, `optimus`), both services (`basius`, `optimus`).
- **`pyproject.toml`** — `[tool.tomte].upstream_pins` entry `valory-xyz/mech-interact@v0.32.6` -> `v0.32.7`. Required for `check-third-party-hashes` CI.
- **Yaml overlays** (5 files) — hoist to top-level Params args:
  - `packages/valory/skills/optimus_abci/skill.yaml` — move `use_offchain: false` and `offchain_deposit_target_calls: 10` out of the nested `mech_marketplace_config` block into top-level params args (plain values in the composed skill.yaml).
  - `packages/valory/agents/basius/aea-config.yaml` and `packages/valory/agents/optimus/aea-config.yaml` — declare `use_offchain: ${USE_OFFCHAIN:bool:false}` and `offchain_deposit_target_calls: ${OFFCHAIN_DEPOSIT_TARGET_CALLS:int:10}` next to the `mech_marketplace_config` line. The default `MECH_MARKETPLACE_CONFIG` dict did not carry either key already, so no removal from the JSON default was needed.
  - `packages/valory/services/basius/service.yaml` and `packages/valory/services/optimus/service.yaml` — strip `use_offchain` and `offchain_deposit_target_calls` from the `${MECH_MARKETPLACE_CONFIG:dict:{...}}` JSON default and add top-level `${USE_OFFCHAIN:bool:false}` / `${OFFCHAIN_DEPOSIT_TARGET_CALLS:int:10}` env-var overlays.
- **Python consumer swap** — none required. `grep -rn "marketplace_config.use_offchain\|_config\.offchain_deposit_target_calls" packages/valory/` returns zero hits in this repo, so no Python callers had to move from the old nested read shape.
- **Tests** — no local test fixtures reference either key. `grep -rn "use_offchain\|offchain_deposit_target_calls" packages/valory/skills/optimus_abci/tests/ packages/valory/skills/liquidity_trader_abci/tests/` returns nothing.

## Operator migration note

**Every deploy-time `MECH_MARKETPLACE_CONFIG` value** (env vars, Propel envs, `.env` files, Docker Compose files) **MUST drop the two nested keys and set the replacements at top level:**

```diff
-MECH_MARKETPLACE_CONFIG='{"...","use_offchain":false,"offchain_deposit_target_calls":10,...}'
+MECH_MARKETPLACE_CONFIG='{"..."}'  # drop both keys from the JSON
+USE_OFFCHAIN=false
+OFFCHAIN_DEPOSIT_TARGET_CALLS=10
```

Otherwise `MechParams.__init__` raises a `ValueError` at agent boot with a message naming the replacement env vars (fail-loud migration guard added in v0.32.7).

## Test plan

- [x] `uv run autonomy packages lock --check` green (53 third-party + 20 dev packages verified)
- [x] `uv run tomte tox -e check-third-party-hashes` green (`valory-xyz/mech-interact@v0.32.7` reachable, all 53 hashes consistent)
- [x] `PYTHONPATH=$PWD uv run pytest packages/valory/skills/optimus_abci/tests/ packages/valory/skills/liquidity_trader_abci/tests/ -q` -> **3184 passed**
- [x] `PYTHONPATH=$PWD uv run pytest packages/valory/skills/mech_interact_abci/tests/ -q` -> **561 passed** (upstream skill loads clean under the new hash)
- [x] `uv run tomte tox -e mypy` green
- [x] `uv run tomte tox -e black-check` green
- [x] `uv run tomte tox -e isort-check` green
- [x] `uv run tomte tox -e flake8` green
- [x] `uv run tomte tox -e darglint` green
- [ ] CI green on this PR
- [ ] Deploy-time env vars migrated on Propel / Pearl before rollout


OjusWiZard (approved):

packages/valory/services/optimus/service.yaml
```diff
@@ -92,7 +92,9 @@ models:
...
       mech_interaction_sleep_time: ${MECH_INTERACTION_SLEEP_TIME:int:10}
       use_mech_marketplace: ${USE_MECH_MARKETPLACE:bool:true}
-      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x46C0D07F55d4F9B5Eed2Fc9680B5953e5fd7b461","priority_mech_address":"0x3208Dc0DebA018dDE8a8e2FFA09265Ea12aAf55E","priority_mech_service_id":195,"response_timeout":300,"use_dynamic_mech_selection":false,"use_offchain":false,"offchain_url":null,"offchain_deposit_target_calls":10,"auto_deposit_cap_per_cycle":5000000,"offchain_poll_interval_seconds":3.0,"offchain_poll_timeout_seconds":300.0,"offchain_failover_max_retries":2}}
+      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x46C0D07F55d4F9B5Eed2Fc9680B5953e5fd7b461","priority_mech_address":"0x3208Dc0DebA018dDE8a8e2FFA09265Ea12aAf55E","priority_mech_service_id":195,"response_timeout":300,"use_dynamic_mech_selection":false,"offchain_url":null,"auto_deposit_cap_per_cycle":5000000,"offchain_poll_interval_seconds":3.0,"offchain_poll_timeout_seconds":300.0,"offchain_failover_max_retries":2}}
+      use_offchain: ${USE_OFFCHAIN:bool:false}
```
jmoreira-valory, line 96: 🔵 **F1 · LOW — USE_OFFCHAIN=true alone bricks boot at every layer — the new knob is not independently settable**

The new `USE_OFFCHAIN` env var reads as a self-contained on/off switch, but there is no layer in this repo where flipping it to `true` on its own produces a bootable agent. `validate_offchain_params` in v0.32.7 cross-checks it against three fields that still live inside `mech_marketplace_config`:

- **Service layer** (this file, and `services/basius/service.yaml`): the `MECH_MARKETPLACE_CONFIG` default has `offchain_url: null` **and** `use_dynamic_mech_selection: false`, so `USE_OFFCHAIN=true` raises `ValueError: use_offchain requires either offchain_url or use_dynamic_mech_selection (to discover the mech's URL)`.
- **Agent layer** (`agents/{optimus,basius}/aea-config.yaml:339`): the `MECH_MARKETPLACE_CONFIG` default dict carries no off-chain fields at all — no `offchain_url`, no `auto_deposit_cap_per_cycle`. `USE_OFFCHAIN=true` there raises a *different* error: `use_offchain requires auto_deposit_cap_per_cycle to be set`.

Both are fail-loud at boot with a readable message, so this is not a shipping blocker — the shipped default is `false` and the PR's stated goal (clear the migration guard) is met. But an operator handed `USE_OFFCHAIN` as a documented knob will hit a boot failure whose message names a key they were just told to stop putting in `MECH_MARKETPLACE_CONFIG`, and the failure differs by layer because the two `MECH_MARKETPLACE_CONFIG` defaults have drifted.

Cheapest fix that removes the surprise: bring the agent-layer default dict up to parity with the service-layer one (add `"offchain_url":null,"auto_deposit_cap_per_cycle":5000000,"offchain_poll_interval_seconds":3.0,"offchain_poll_timeout_seconds":300.0,"offchain_failover_max_retries":2` to `agents/{optimus,basius}/aea-config.yaml`), so at least both layers fail the same way. Alternatively, add a one-line comment next to `use_offchain` noting that enabling it also requires `offchain_url` or `use_dynamic_mech_selection: true` in `MECH_MARKETPLACE_CONFIG`. Happy to see this deferred to whichever PR actually turns off-chain on.

packages/valory/skills/optimus_abci/skill.yaml
```diff
@@ -347,13 +347,13 @@ models:
...
         response_timeout: 300
         use_dynamic_mech_selection: false
-        use_offchain: false
         offchain_url: null
-        offchain_deposit_target_calls: 10
         auto_deposit_cap_per_cycle: 5000000
         offchain_poll_interval_seconds: 3.0
         offchain_poll_timeout_seconds: 300.0
         offchain_failover_max_retries: 2
+      use_offchain: false
```
jmoreira-valory, line 355: ⚪ **F2 · NIT — No CI guard that the composed skill.yaml declares every MechParams _ensure key**

This PR exists because a required sub-skill param can go missing from the composed overlay and only surface as an `AEAEnforceError` at agent boot. Nothing in CI catches that class of break today: `autonomy packages lock --check` only verifies hashes, and the 3184 passing unit tests never instantiate `optimus_abci.models.Params` from the yaml args, so a fourth hoisted param in `v0.32.8` would go green here and fail on Propel.

A cheap guard is an import-only test that loads `packages/valory/skills/optimus_abci/skill.yaml`, collects `models.params.args` keys, and asserts every `self._ensure("<key>", kwargs, ...)` literal reachable through the `Params` MRO is present. Pure-yaml, no skill context needed. Not blocking this PR — worth a follow-up issue, since the same trap will recur on every mech-interact bump.

jmoreira-valory (approved):
### Review round 1 — ✅ Approve

**What this does.** Upstream mech-interact v0.32.7 moved two knobs, `use_offchain` and `offchain_deposit_target_calls`, out of the nested `MechMarketplaceConfig` dataclass and made them required top-level `MechParams` args, with a fail-loud guard that rejects the old nested keys. This PR bumps the `mech_interact_abci` third-party hash in optimus and re-declares both knobs at the top level in all five overlay files (composed skill.yaml, both agents' aea-config.yaml, both services' service.yaml), plus the `upstream_pins` entry. Without it, every optimus/basius agent would raise `AEAEnforceError` at boot on the new hash.

Verified independently against source at HEAD rather than taking the PR body at its word:

- **Hash matches upstream exactly.** `bafybeic5xfxdod5l56o7aojdamojnvi562nql2mlah73bf47d4guz4xnwi` is byte-identical to the `mech_interact_abci` entry in `packages.json` on the `v0.32.7` tag.
- **The upstream delta really is just these two params.** Diffing `MechParams.__init__` between `v0.32.6` and `v0.32.7` shows only the legacy-key guard plus the two new `_ensure` reads; the `MechMarketplaceConfig` dataclass loses exactly those two fields and nothing else. No other newly-required param is silently missing from the overlays. The other UPGRADING.md breaking changes (`valid_tools` / `ignored_mechs` / `irrelevant_tools` removal) predate `v0.32.6` and this repo is already migrated (`valid_mechs` is present, the removed keys are absent).
- **Only `optimus_abci` needs the keys.** `optimus_abci.models.Params` mixes in `MechParams`; `liquidity_trader_abci.Params` derives from `BaseParams` only, so the hash-only bump in `liquidity_trader_abci/skill.yaml` is correct and complete.
- **Placement and typing are right.** Both keys land in the `valory/optimus_abci` skill's `models.params.args` block in each aea-config/service.yaml (not a stray block), at sibling indentation, with `bool` / `int` env types matching the `_ensure` types.
- **Cascade is complete.** The six dev packages in `packages.json` are exactly the transitive consumers; no dev package was missed.
- **Migration blast radius is smaller than the PR body implies.** Pearl's `olas-operate-middleware` builds `MECH_MARKETPLACE_CONFIG` in `operate/services/manage.py` from a fixed f-string that carries neither legacy key, and neither `quickstart` nor `olas-operate-app` mentions `use_offchain` anywhere. So the unchecked "migrate deploy-time env vars" box only bites hand-written Propel envs / `.env` files, not the managed deployment paths.

CI is green (24/24 checks) at this HEAD and the PR is mergeable. Both findings below are non-blocking.

**Findings** — 2 · 🔵 1 Low · ⚪ 1 Nit

| # | Sev | Location | Issue |
|----|-----|----------|-------|
| F1 | 🔵 LOW | [packages/valory/services/optimus/service.yaml:96](https://github.com/valory-xyz/optimus/blob/c53cdf73e67b1591fb25d4c8ad18bd3fe3db8763/packages/valory/services/optimus/service.yaml#L96) | **USE_OFFCHAIN=true alone bricks boot at every layer — the new knob is not independently settable** |
| F2 | ⚪ NIT | [packages/valory/skills/optimus_abci/skill.yaml:355](https://github.com/valory-xyz/optimus/blob/c53cdf73e67b1591fb25d4c8ad18bd3fe3db8763/packages/valory/skills/optimus_abci/skill.yaml#L355) | **No CI guard that the composed skill.yaml declares every MechParams _ensure key** |

PR #370, opened 2026-08-13
fix(services): enable dynamic mech selection so offchain-enabled Optimus + Basius boot

DIvyaNautiyal07:
## Summary
- Flips `use_dynamic_mech_selection` from `false` → `true` inside the `MECH_MARKETPLACE_CONFIG` default of `optimus` and `basius` service.yaml.
- Lets `mech_interact_abci` look up the pinned priority mech's registered offchain endpoint from the marketplace contract at boot when `USE_OFFCHAIN=true`.
- Priority mech pinning is unchanged (`priority_mech_address` still points at `0x3208…F55E` on Optimism / `0xe535…cA2B` on Base). Only the URL discovery flips from static (null) to dynamic (resolve via marketplace).

## Context
When `USE_OFFCHAIN=true` is exported by Pearl for Optimus/Basius, `mech_interact_abci.models.validate_offchain_params` raises:

```
ValueError: use_offchain requires either offchain_url or
use_dynamic_mech_selection (to discover the mech's URL)
```

The validator requires at least one of `offchain_url` or `use_dynamic_mech_selection` to be truthy so the skill knows where to POST offchain requests. Trader satisfies the validator because its yaml omits the key and the model default is `True`. Optimus and Basius previously set the key explicitly to `false` in the JSON default, which was fine for on-chain-only dispatch but blocks offchain.

## Test plan
- `autonomy packages lock` updates only the two touched service package hashes.
- `autonomy packages lock --check` verifies clean.
- After merge + release, a Pearl build that pins the new version and exports `USE_OFFCHAIN=true` should boot Optimus/Basius without the `ValueError` on skill load.


Tanya-atatakai (approved):

OjusWiZard (approved):

kupermind:
Reviewed at `8cb3075` against the pinned `mech_interact_abci` (`bafybeif3cnuns2i7xh2wyrpj45xdmkjkgqpy72ox66tijtbf5okn2l2iiu`, which is `mech-interact@ee9f4d6`). Not approving — I think this trades a boot-time failure for a runtime one, and the flag does more than the description says. Flagging post-merge so it can be caught before the release rather than in a Pearl build.

## `use_dynamic_mech_selection=true` unpins the priority mech; it is not a URL-discovery switch

The description says *"Priority mech pinning is unchanged … Only the URL discovery flips from static (null) to dynamic"*. The skill says otherwise at boot — `models.py:409-416`:

```python
if (
    self.mech_marketplace_config.use_dynamic_mech_selection
    and self.mech_marketplace_config.priority_mech_address
):
    self.context.logger.info(
        "A priority mech has been set while dynamic mech selection is enabled. "
        "The priority mech will be ignored."
    )
```

and at request time `get_priority_mech_address` (`behaviours/request.py:570`) takes an entirely different branch — the ranked list from the mechs subgraph — instead of reading `priority_mech_address`. `priority_mech_service_id` goes unused with it. `UPGRADING.md` calls this out as the Discoverability feature: *"The priority mech is now dynamically picked."*

## On this config that branch raises

`ranked_mechs_addresses` → `ranked_mechs` → `relevant_mechs_info` (`states/base.py:420`), whose comprehension evaluates `self.mech_tool` per candidate → `db.get_strict("mech_tool")` (`states/base.py:395`). Optimus and Basius never write that db key — `mech_tool` is a skill param, read via `_ensure` and consumed as `MechMetadata.tool` in `check_staking_kpi_met._build_mech_request_metadata` (`liquidity_trader_abci/behaviours/check_staking_kpi_met.py:242`). Grepping the repo, the only occurrences are the param.

Ran the real classes at the pinned hash rather than arguing it from the source:

```
A mechs_info=1  mech_tool absent : RAISED ValueError: 'mech_tool' field is not set for this period [0] and no default value was provided.
B mechs_info=1  mech_tool set    : ranked_mechs_addresses -> ['0x3208Dc0DebA018dDE8a8e2FFA09265Ea12aAf55E']
C mechs_info=0  mech_tool absent : ranked_mechs_addresses -> []
```

**B** is trader's shape — it writes `mech_tool` to the db from its tool-selection round, which is why the same flag is fine there. **A** is ours once `MechInformationRound` has populated `mechs_info`, and `MechRequestBehaviour.setup()` (`request.py:621`) calls this unguarded, so it raises before the request is built. **C** is the other branch: if `MechInformationRound` can't resolve tools it returns `None` → `Event.NONE` → `FailedMechInformationRound`, which `composition.py` routes straight back to `MechVersionDetectionRound` — the FSM ping-pongs between the two and never reaches `MechRequestRound`.

Either way the staking liveness request stops going out. Which of the two you land on depends on whether the mechs subgraph + tools manifest resolve at that moment, which is not something the config controls.

The information leg is definitely on the path — both marketplaces resolve as v2, which is what `MechVersionDetectionBehaviour` probes via `get_max_fee_factor`:

```
MAX_FEE_FACTOR() @ 0x46C0D07F55d4F9B5Eed2Fc9680B5953e5fd7b461 (optimism) -> 10000
MAX_FEE_FACTOR() @ 0xf24eE42edA0fc9b33B7D41B06Ee8ccD2Ef7C5020 (base)     -> 10000
```

The offchain path is gated on the same flag and hits the same property — `_pick_next_mech` (`offchain_request.py:1663`) iterates `self._synced.ranked_mechs` when the flag is on, and its own comment is precisely this hazard:

> `ranked_mechs` reads `mech_tool` strictly from db, which is only written by the v2 dynamic-selection round. Consumers using a hardcoded `priority_mech_address` never write it, so traversing `ranked_mechs` here would raise on every offchain cycle. Mirrors the same gate `request.py` uses on the on-chain path.

That gate is what the flip removes for these two services.

## Blast radius is wider than the offchain case this is fixing

`use_offchain` still defaults to `false` in both service.yamls, and `use_dynamic_mech_selection` is read on the on-chain path too (`request.py:570`). So every Optimus/Basius deployment built from this release changes mech-selection behaviour, not just the ones Pearl exports `USE_OFFCHAIN=true` to. The validator being unblocked was an offchain-only problem; the fix is unconditional.

## The other arm of the validator looks like the intended fix

`validate_offchain_params` accepts *either* `offchain_url` **or** `use_dynamic_mech_selection`. Setting `offchain_url` and leaving `use_dynamic_mech_selection: false` satisfies it while keeping the pin: `get_priority_mech_address` stays on the static branch (and the pinned address is already in `valid_mechs`, so the allowlist check passes), and `_pick_next_mech` skips the `ranked_mechs` traversal and returns `(priority_mech_address, offchain_url)` from its static fallback. No db key required.

One correction on where the URL comes from, since it affects that choice: it is not read from the marketplace contract. `MechInformationBehaviour.populate_tools` fetches the mech's IPFS metadata manifest and `_extract_manifest_url` pulls the `url` key out of it — the value operators publish via `make update-metadata` in mech-deployments. Worth confirming both pinned mechs actually publish that key before relying on discovery, because with `offchain_url: null` there is no fallback: `_pick_next_mech` returns `(None, None)` and the request ends in `OFFCHAIN_ALL_FAILED`.

## Minor

Both agents' `aea-config.yaml` and `optimus_abci/skill.yaml` still carry `use_dynamic_mech_selection: false`. The service override wins for deployed services so this doesn't change any of the above, but `autonomy fetch agent` now yields a different default from the service, which will be confusing next time someone debugs this locally.

Happy to be wrong about the `mech_tool` db key if there's a writer I've missed — the probe above is easy to re-run with a populated db.

PR #371, opened 2026-08-13
fix(services): pin priority mech + offchain URL on optimus + basius (addresses #370 review)

DIvyaNautiyal07:
Follow-up to #370 addressing kupermind's [post-merge review](https://github.com/valory-xyz/optimus/pull/370#issuecomment-5283835139) before v0.12.9 ships in a Pearl build.

## Summary
- `offchain_url` is set to the priority mech's published propel endpoint in `optimus` and `basius` service.yaml.
- `use_dynamic_mech_selection` reverts to `false` (undoing #370's flip).
- Together these satisfy `mech_interact_abci.validate_offchain_params` while keeping the static-pin path intact.

## URLs
- Optimus (Optimism, priority mech `0x3208…F55E`): `https://1dcd79932c59eb58.agent.propel.autonolas.tech`
- Basius (Base, priority mech `0xe535…cA2B`): `https://9ea5c24c68527852.agent.propel.autonolas.tech`

## Why the flag flip was the wrong arm
`validate_offchain_params` accepts *either* `offchain_url` or `use_dynamic_mech_selection`. #370 chose the second arm, but on this consumer that arm unpins the priority mech and reroutes selection through `ranked_mechs`:

- `mech_interact_abci/behaviours/request.py:570-596` — with `use_dynamic_mech_selection=true`, `get_priority_mech_address` reads `synchronized_data.ranked_mechs_addresses` instead of `priority_mech_address`.
- `mech_interact_abci/states/base.py:397` — `ranked_mechs` evaluates the `mech_tool` property, which does `self.db.get_strict("mech_tool")`.
- `liquidity_trader_abci` never writes `mech_tool` to synchronized data — it only reads `self.params.mech_tool` as a skill param (`check_staking_kpi_met.py:242`). Grepping the tree confirms no producer round.
- So `db.get_strict("mech_tool")` raises `ValueError: 'mech_tool' field is not set for this period [0]` on every `mech_request_round` cycle for Optimus/Basius, regardless of whether `USE_OFFCHAIN` is exported.
- `offchain_request.py:1682-1685` already documents this hazard and gates the offchain path against it; the on-chain path in `request.py` has no such guard, so #370's flip removes both.

## Why the static-pin + explicit URL arm works
`_pick_next_mech` returns `(priority_mech_address, offchain_url)` from the static fallback, so no db read of `mech_tool` and the pinned mech (already in `valid_mechs`) passes the allowlist check. `get_priority_mech_address` stays on the static branch.

## Test plan
- `autonomy packages lock` updates only the two service package hashes (basius, optimus).
- `autonomy packages lock --check` passes.
- After merge + release, a Pearl build that pins the new version and exports `USE_OFFCHAIN=true` boots Optimus/Basius without either the boot-time `ValueError: use_offchain requires either offchain_url or use_dynamic_mech_selection` or the runtime `ValueError: 'mech_tool' field is not set for this period`.

## Notes
- `aea-config.yaml` and `optimus_abci/skill.yaml` still carry `use_dynamic_mech_selection: false` (they always did — #370 only touched service.yaml). Reverting service.yaml to `false` puts everything back in alignment; no drift for `autonomy fetch agent` locally.
- `use_offchain` still defaults to `false` in the yaml. The offchain path only engages when Pearl exports `USE_OFFCHAIN=true`; when it doesn't, the static-pin behaviour is unchanged from pre-#370.


Tanya-atatakai (approved):

OjusWiZard (approved):

kupermind (approved):
Re-reviewed at `b4e55e3`. This is the right arm of the validator and the reasoning in the description matches what the skill actually does. Approving.

Rather than reading the diff for it, I ran the new config through the pinned `mech_interact_abci` — both service dicts, all four properties that mattered:

```
[optimus] parsed: dynamic=False url=https://1dcd79932c59eb58.agent.propel.autonolas.tech
[optimus] validate_offchain_params(use_offchain=True) -> ok
[optimus] get_priority_mech_address -> 0x3208Dc0DebA018dDE8a8e2FFA09265Ea12aAf55E
[optimus] _pick_next_mech([])        -> ('0x3208…F55E', 'https://1dcd79932c59eb58.agent.propel.autonolas.tech')
[optimus] _pick_next_mech([pinned])  -> (None, None)

[basius]  parsed: dynamic=False url=https://9ea5c24c68527852.agent.propel.autonolas.tech
[basius]  validate_offchain_params(use_offchain=True) -> ok
[basius]  get_priority_mech_address -> 0xe535D7AcDEeD905dddcb5443f41980436833cA2B
[basius]  _pick_next_mech([])        -> ('0xe535…cA2B', 'https://9ea5c24c68527852.agent.propel.autonolas.tech')
[basius]  _pick_next_mech([pinned])  -> (None, None)
```

The stub I passed for `synchronized_data` raises on any `ranked_mechs` / `ranked_mechs_addresses` access, and nothing raised — so the static path genuinely never reaches the `mech_tool` strict read on either the on-chain or the offchain side. `should_use_marketplace_v2()` was stubbed `True` (both marketplaces do expose `MAX_FEE_FACTOR`), so the `and` in `get_priority_mech_address` is short-circuiting on the flag exactly as intended, not on the version check. Both pinned addresses pass the `valid_mechs` allowlist. All checks green, and `lock_check` confirms the two service hashes are the only movement.

Two things to look at before the release goes into a Pearl build. Neither blocks the merge, since `use_offchain` still defaults to `false` and nothing changes until Pearl exports it.

**The Basius endpoint is not currently serving.** Three consecutive requests, ~0.6s each, so it's nginx answering rather than a timeout:

```
GET https://9ea5c24c68527852.agent.propel.autonolas.tech/healthcheck -> 503  (x3)
GET https://1dcd79932c59eb58.agent.propel.autonolas.tech/healthcheck -> 200
     {"is_healthy": true, "liveness": {"ok": true}, "readiness": {"ok": true, "reason": "idle-ok"},
      "current_round": "reset_and_pause_round", "period": 563, ...}
```

The Optimism side is healthy and clearly the right kind of thing on the other end. The Base one may just be a down deployment rather than a wrong subdomain, but worth confirming which before `USE_OFFCHAIN=true` reaches Basius — and note that `offchain_failover_max_retries: 2` buys nothing here: the static fallback yields exactly one candidate, and `_pick_next_mech([pinned])` is `(None, None)` (see the probe above), so a dead endpoint is a straight `OFFCHAIN_ALL_FAILED` on the first cycle, not a retry.

**The "no drift" note in the description isn't quite right.** The agents' `aea-config.yaml` dicts are shorter than the service ones — they stop after `use_dynamic_mech_selection`:

```
{"mech_marketplace_address":"0x46C0…b461","priority_mech_address":"0x3208…F55E",
 "priority_mech_service_id":195,"response_timeout":300,"use_dynamic_mech_selection":false}
```

No `offchain_url`, no `auto_deposit_cap_per_cycle`. So the flag is back in alignment, but the agent config still can't run offchain at all — with `USE_OFFCHAIN=true` it fails both arms of `validate_offchain_params` (missing URL *and* `auto_deposit_cap_per_cycle is None`). That predates both PRs and only affects `autonomy fetch agent` locally, so it's a fine follow-up rather than something to hold this for — just not the "no drift" the description claims.

Thanks for the quick turnaround, and for writing up *why* the first arm was wrong rather than just swapping the values.

